PR #2, opened 2015-01-21
Refinement

guptashvm:


inducer:
The main comments I've got for now are superficial (sorry):

* Take a look at [PEP8](https://www.python.org/dev/peps/pep-0008/). It sets forth some conventions that lots of Python code (including mine) follows. The main thing that applies here is the line length limit. I stick to line lengths of 85 or shorter. I find [flake8](https://gitlab.com/pycqa/flake8) useful for keeping my code sane. There's a config file for it already in the repository.
* You don't need backslash line continuations inside of brackets or parentheses.
* for/if loops to test refinement flags can be simplified and accelerated with `np.where`.
* Sprinkle comments and blank lines. They help readability.
* Instead of repeatedly computing values (such as the min/max expressions in the indices), create short-lived temporaries.
* If you can break up one massive function into multiple smaller ones with a self-documenting name, that's usually a win.

inducer:
Closing this in favor of https://gitlab.tiker.net/inducer/meshmode/merge_requests/6.

PR #3, opened 2016-07-28
Implement make_refinement_connection().

mattwala:


mattwala:
@inducer, I'd appreciate if you took a look at two things in particular:
1. Does the interface make sense? Note that it requires running the refiner first before calling `make_refinement_connection`.
2. Am I testing for the right thing? Is there anything else that should be tested?

inducer:
Looks great! Only spotted two small things:

* `make_refinement_connection` should probably be in its own submodule under `m.d.connection`.
* No documentation for the top-level interface.

To answer your question, the test looks exactly like what I would've tested. Could you open the next version of this PR on gitlab? Then it'll have the tests run automatically. Added you to the project. (And a bunch of other ones while I'm at it.) Closing here.

PR #11, opened 2020-05-19
Fix #9 (Mesh distribution fails when number of ranks >= 29)

### LINKED ISSUE #9: Mesh distribution fails when number of ranks >= 29
When distributing a mesh generated by `generate_regular_rect_mesh`,  I'm encountering the following error when running on 29 or more MPI ranks:

```
(dg) bash-4.2$ srun -n 29 python -m mpi4py example.py
  File "<removed>/.miniconda3/envs/dg/lib/python3.8/runpy.py", line 193, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "<removed>/.miniconda3/envs/dg/lib/python3.8/runpy.py", line 86, in _run_code
    exec(code, run_globals)
  File "<removed>/.miniconda3/envs/dg/lib/python3.8/site-packages/mpi4py/__main__.py", line 7, in <module>
    main()
  File "<removed>/.miniconda3/envs/dg/lib/python3.8/site-packages/mpi4py/run.py", line 196, in main
    run_command_line(args)
  File "<removed>/.miniconda3/envs/dg/lib/python3.8/site-packages/mpi4py/run.py", line 47, in run_command_line
    run_path(sys.argv[0], run_name='__main__')
  File "<removed>/.miniconda3/envs/dg/lib/python3.8/runpy.py", line 263, in run_path
    return _run_module_code(code, init_globals, run_name,
  File "<removed>/.miniconda3/envs/dg/lib/python3.8/runpy.py", line 96, in _run_module_code
    _run_code(code, mod_globals, init_globals,
  File "<removed>/.miniconda3/envs/dg/lib/python3.8/runpy.py", line 86, in _run_code
    exec(code, run_globals)
  File "example.py", line 21, in <module>
    local_mesh = mesh_dist.send_mesh_parts(mesh, part_per_element, num_parts)
  File "<removed>/code/meshmode/meshmode/distributed.py", line 85, in send_mesh_parts
    parts = [partition_mesh(mesh, part_per_element, i)[0]
  File "<removed>/code/meshmode/meshmode/distributed.py", line 85, in <listcomp>
    parts = [partition_mesh(mesh, part_per_element, i)[0]
  File "<removed>/code/meshmode/meshmode/mesh/processing.py", line 163, in partition_mesh
    part_mesh = Mesh(
  File "<removed>/code/meshmode/meshmode/mesh/__init__.py", line 718, in __init__
    raise ValueError("too few bits in element_id_dtype to represent all "
ValueError: too few bits in element_id_dtype to represent all boundary tags
```

Below is a simplified version of the code that can be used to reproduce the error:

```
from mpi4py import MPI


comm = MPI.COMM_WORLD

num_parts = comm.Get_size()

from meshmode.distributed import MPIMeshDistributor, get_partition_by_pymetis
mesh_dist = MPIMeshDistributor(comm)

dim = 2

if mesh_dist.is_mananger_rank():
    from meshmode.mesh.generation import generate_regular_rect_mesh
    mesh = generate_regular_rect_mesh(a=(-0.5,)*dim,
                                      b=(0.5,)*dim,
                                      n=(32,)*dim)

    part_per_element = get_partition_by_pymetis(mesh, num_parts)

    local_mesh = mesh_dist.send_mesh_parts(mesh, part_per_element, num_parts)
else:
    local_mesh = mesh_dist.receive_mesh_part()
```

I also have the serialized `mesh` output from an attempted run with 29 ranks ([make_mesh.py](https://github.com/inducer/meshmode/files/4601329/make_mesh.py.txt)).

majosm:
This change moves creation of the partitioned mesh's facial adjacency groups before creation of the actual mesh object so that the neighboring ranks can be detected beforehand. This allows `boundary_tags` to be assembled using only the neighbors instead of adding a `BTAG_PARTITION` for every rank.


inducer:
Thanks for getting going on this! By default, I won't look at MRs in too much detail until

- they're not marked draft
- the CIs pass

But I'm happy to look earlier if you'd like me to. Just let me know.

majosm:
Fixed the tests... well, kind of. :) I ran into a couple of problems with the EOCRecorder, so I've temporarily disabled one of the assertions ([here](https://github.com/inducer/meshmode/blob/073cd96713c44d770f961bb29d07a6183b715c94/test/test_partition.py#L199-L203)). The problems I ran into are:
1. If one of the error data points is exactly 0, I get an exception raised in `log10` when computing the order estimate
2. After this change, the number of data points collected for a given pair of partitions will vary depending on the decomposition. In some cases there aren't enough to compute an order estimate, leading to an error.

 Looking for advice on how I might deal with these.

test/test_partition.py
```diff
@@ -238,6 +245,19 @@ def test_partition_mesh(num_parts, num_meshes, dim, scramble_partitions):
         [count_tags(new_meshes[i][0], BTAG_ALL) for i in range(num_parts)]), \
         "part_mesh has the wrong number of BTAG_ALL boundaries"
 
+    connected_parts = set()
```
inducer, line 248: Would `meshmode.distributed.get_connected_partitions` be able to do this?

test/test_partition.py
```diff
@@ -102,6 +97,23 @@ def f(x):
                               adjncy=mesh.nodal_adjacency.neighbors.tolist())
             part_per_element = np.array(p)
 
+        connected_parts = set()
```
inducer, line 100: Would `meshmode.distributed.get_connected_partitions` be able to do this?

test/test_partition.py
```diff
@@ -192,8 +196,11 @@ def f(x):
     for (i, j), e in eoc_rec.items():
         if e is not None:
             print("Error of connection from part %i to part %i." % (i, j))
-            print(e)
-            assert(e.order_estimate() >= order - 0.5 or e.max_error() < 1e-11)
+            # TODO: Figure out how to deal with divide-by-zero and
+            # not-enough-data-points errors
+            # print(e)
+            # assert(e.order_estimate() >= order - 0.5 or e.max_error() < 1e-11)
```
inducer, line 202: Not sure I understand why these are an issue. Could you explain?

meshmode/mesh/processing.py
```diff
@@ -156,22 +156,25 @@ def partition_mesh(mesh, part_per_element, part_num):
...
+
+    def boundary_tag_bit(boundary_tag):
+        from meshmode.mesh import _boundary_tag_bit
+        return _boundary_tag_bit(boundary_tags, btag_to_index, boundary_tag)
 
-    adj_data = [[] for _ in range(len(part_mesh.groups))]
+    from meshmode.mesh import _compute_facial_adjacency_from_vertices
+    facial_adjacency_groups = _compute_facial_adjacency_from_vertices(
+                                    new_mesh_groups, boundary_tags,
+                                    mesh.element_id_dtype, mesh.face_id_dtype)
```
inducer, line 171: It is not OK to recompute facial adjacency: First, it's already there. Second, `mesh.facial_adjacency` may contain better information than what can be computed from vertices, e.g. in the case of hanging nodes.

meshmode/mesh/processing.py
```diff
@@ -248,7 +261,7 @@ def partition_mesh(mesh, part_per_element, part_num):
                     n_faces[bdry_idx] = adj_n_faces[adj_idx]
                     adj_idx += 1
 
-            connected_mesh.facial_adjacency_groups[igrp][None] =\
+            part_mesh.facial_adjacency_groups[igrp][None] =\
```
inducer, line 264: Yikes. In-place updates are not OK. I missed those when I reviewed this code initially.

test/test_partition.py
majosm: Yes! Fixed.

meshmode/mesh/processing.py
majosm: This recomputation already existed in the original code (previously it occurred when creating the partitioned mesh with `facial_adjacency_groups=None`). I basically just moved it. I think I can see how it would be possible to use the global mesh's facial adjacency to generate that of the partitioned mesh instead of recomputing from scratch; I could take a stab at it. Should that be part of this PR, or a new one after this?

meshmode/mesh/processing.py
majosm: By in-place update you mean the reassigning of `facial_adjacency_groups` to `InterPartitionAdjacencyGroup`? I wonder if I could fix that when removing the facial adjacency recomputation (other comment).

test/test_partition.py
majosm: I took another look at this, and I'm tempted to say the `order_estimate` stuff should just be removed. In the unmodified code, the max error at each mesh size is either identically 0 or quite small (order 1e-12 was the largest I saw), with no obvious trend as the mesh resolution is increased. In most cases the error estimate ends up being NaN or negative and `e.order_estimate() >= order - 0.5` evaluates to `False`. It does evaulate to `True` in a couple of cases, but it appears to be incidental, e.g.:

```
Error of connection from part 1 to part 2.
h                  | Error                 | Running EOC       
-------------------+-----------------------+--------------------
0.3333333333333333 | 5.284661597215745e-13 |                   
0.25               | 4.334310688136611e-13 | 0.6891148899301487

Overall EOC: 0.6891148899301487
```

The original comments in the code I added reference a couple of warnings/errors I was seeing when it computes the order estimate. In the unmodified code, for some cases the error would be identically 0 and it would emit:

```
<removed>/pytools/convergence.py:21: RuntimeWarning: divide by zero encountered in log10
  coefficients = np.polyfit(np.log10(abscissae), np.log10(errors), 1)
```

After my modifications, it would in some of the cases also give 

```
E           RuntimeError: Need more than one value to guess order of convergence.
```

because it no longer creates a data point for every pair of partitions at each mesh scale, only those that are adjacent, and thus sometimes does not have enough data points to compute an order estimate.

majosm:
Could split this into several PRs pretty easily if the diff is too cluttered. Let me know.

meshmode/mesh/processing.py
```diff
@@ -93,173 +92,314 @@ def partition_mesh(mesh, part_per_element, part_num):
...
-        if group_num not in skip_groups:
-            new_mesh_groups.append(
-                type(mesh_group)(
-                    mesh_group.order, new_indices[group_num],
-                    new_nodes[group_num],
-                    unit_nodes=mesh_group.unit_nodes))
-
-    from meshmode.mesh import BTAG_ALL, BTAG_PARTITION
-    boundary_tags = [BTAG_PARTITION(n) for n in np.unique(part_per_element)]
+    for indices in new_indices:
```
inducer, line 144: This function is crazy long, which makes it hard to read. Do you think you could break it up a bit?

majosm:
Alright, I think this is ready for another look.

meshmode/mesh/processing.py
```diff
@@ -62,6 +62,367 @@ def find_group_indices(groups, meshwide_elems):
...
+        new_indices.append(grp.vertex_indices[group_elements - grp.element_nr_base])
+        ambient_dim = grp.nodes.shape[0]
+        new_nodes.append(
+            np.zeros(
+                (ambient_dim, end_idx - start_idx, grp.nunit_nodes)))
+        for i in range(ambient_dim):
+            for j in range(start_idx, end_idx):
+                new_idx = j - start_idx
+                elem = group_elements[new_idx] - grp.element_nr_base
+                new_nodes[-1][i, new_idx, :] = grp.nodes[i, elem, :]
```
inducer, line 137: With a 'new-to-old' element index map in an array, these four could be one line (using fancy indexing)

meshmode/mesh/processing.py
```diff
@@ -62,6 +62,367 @@ def find_group_indices(groups, meshwide_elems):
...
+        index_sets = np.append(index_sets, set(new_indices[-1].ravel()))
+
+    # A sorted np.array of vertex indices we need (without duplicates).
+    #required_vertex_indices = np.unique(np.sort(index_set))
+    required_vertex_indices = np.array(list(set.union(*index_sets)))
+
+    # Our indices need to be in range [0, len(mesh.nelements)].
```
inducer, line 144: This comment seems misleading (as applied to `new_indices`, which IIUC are vertex indices).

meshmode/mesh/processing.py
```diff
@@ -62,6 +62,367 @@ def find_group_indices(groups, meshwide_elems):
...
+    for indices in new_indices:
+        for i in range(len(indices)):
+            for j in range(len(indices[0])):
+                original_index = indices[i, j]
+                indices[i, j] = np.where(required_vertex_indices
+                            == original_index)[0]
```
inducer, line 150: `np.unique` has a handy `return_inverse` feature, which is I think what you're looking for here. FWIW, per-vertex loops aren't ideal, especially when they're avoidable (like these). You can use `reshape` to avoid the `j` loop.

https://numpy.org/doc/stable/reference/generated/numpy.unique.html

test/test_partition.py
```diff
@@ -190,13 +183,11 @@ def f(x):
...
 
-    for (i, j), e in eoc_rec.items():
-        if e is not None:
-            print("Error of connection from part %i to part %i." % (i, j))
-            print(e)
-            assert(e.order_estimate() >= order - 0.5 or e.max_error() < 1e-11)
+            # Can't currently expect exact results due to limitations of
+            # interpolation 'snapping' in DirectDiscretizationConnection's
+            # _resample_point_pick_indices
+            assert(err < 1e-11)
```
inducer, line 190:
```suggestion
            assert err < 1e-11
```

test/test_partition.py
```diff
@@ -309,11 +307,12 @@ def test_partition_mesh(num_parts, num_meshes, dim, scramble_partitions):
...
+    for i_remote_part in range(num_parts):
         tag_sum = 0
-        for mesh, _ in new_meshes:
-            tag_sum += count_tags(mesh, BTAG_PARTITION(i_tag))
-        assert num_tags[i_tag] == tag_sum,\
+        for i_local_part, (mesh, _) in enumerate(new_meshes):
+            if (i_local_part, i_remote_part) in connected_parts:
+                tag_sum += count_tags(mesh, BTAG_PARTITION(i_remote_part))
+        assert num_tags[i_remote_part] == tag_sum,\
                 "part_mesh has the wrong number of BTAG_PARTITION boundaries"
```
inducer, line 316: Do we have a test somewhere that exercises the large number of parts that triggered the failure?

meshmode/mesh/processing.py
```diff
@@ -62,6 +62,367 @@ def find_group_indices(groups, meshwide_elems):
...
+                    unit_nodes=grp.unit_nodes))
+
+    return new_groups, group_to_new_group, required_vertex_indices
+
+
+def _create_local_to_local_adjacency_groups(mesh, global_elem_to_part_elem,
+            part_mesh_groups, global_group_to_part_group,
+            part_mesh_group_elem_base):
+    """Create local-to-local facial adjacency groups for partitioned mesh."""
+    local_to_local_adjacency_groups = [dict() for _ in part_mesh_groups]
```
inducer, line 168:
```suggestion
    local_to_local_adjacency_groups = [{} for _ in part_mesh_groups]
```

meshmode/mesh/processing.py
```diff
@@ -62,6 +62,367 @@ def find_group_indices(groups, meshwide_elems):
...
+
+            elem_base_i = mesh.groups[igrp].element_nr_base
+            elem_base_j = mesh.groups[jgrp].element_nr_base
+
+            elements_are_local = global_elem_to_part_elem[facial_adj.elements
+                        + elem_base_i] >= 0
+            neighbors_are_local = global_elem_to_part_elem[facial_adj.neighbors
+                        + elem_base_j] >= 0
+
+            adj_indices = np.where(np.logical_and(elements_are_local,
```
inducer, line 191: `&` is the same as `np.logical_and` (in a slightly weird overloading of the 'binary-and' operator).

meshmode/mesh/processing.py
```diff
@@ -62,6 +62,367 @@ def find_group_indices(groups, meshwide_elems):
...
+
+
+def _collect_nonlocal_adjacency_data(mesh, part_per_elem, global_elem_to_part_elem,
+            part_mesh_groups, global_group_to_part_group,
+            part_mesh_group_elem_base):
+    """
+    Collect nonlocal adjacency data for partitioned mesh, and store it in an
+    intermediate data structure to use when constructing inter-partition adjacency
+    groups.
+    """
```
inducer, line 224: Please briefly describe the structure of the intermediate data structure.

meshmode/mesh/processing.py
```diff
@@ -62,6 +62,367 @@ def find_group_indices(groups, meshwide_elems):
...
+
+    return nonlocal_adj_data
+
+
+def _collect_bdry_data(mesh, global_elem_to_part_elem, part_mesh_groups,
+            global_group_to_part_group, part_mesh_group_elem_base):
+    """
+    Collect boundary data for partitioned mesh, and store it in an intermediate
+    data structure to use when constructing inter-partition adjacency groups.
+    """
```
inducer, line 283: Please briefly describe the structure of the intermediate data structure.

meshmode/mesh/processing.py
```diff
@@ -62,6 +62,367 @@ def find_group_indices(groups, meshwide_elems):
...
+            elem_base_j = mesh.groups[jgrp].element_nr_base
+
+            elements_are_local = global_elem_to_part_elem[facial_adj.elements
+                        + elem_base_i] >= 0
+            neighbors_are_nonlocal = global_elem_to_part_elem[facial_adj.neighbors
+                        + elem_base_j] < 0
+            adj_indices = np.where(np.logical_and(elements_are_local,
+                        neighbors_are_nonlocal))[0]
+
+            if len(adj_indices) > 0:
```
inducer, line 246:
```suggestion
            if adj_indices:
```

meshmode/mesh/processing.py
```diff
@@ -62,6 +62,367 @@ def find_group_indices(groups, meshwide_elems):
...
+                continue
+
+            elem_base_i = mesh.groups[igrp].element_nr_base
+            elem_base_j = mesh.groups[jgrp].element_nr_base
+
+            elements_are_local = global_elem_to_part_elem[facial_adj.elements
+                        + elem_base_i] >= 0
+            neighbors_are_nonlocal = global_elem_to_part_elem[facial_adj.neighbors
+                        + elem_base_j] < 0
+            adj_indices = np.where(np.logical_and(elements_are_local,
```
inducer, line 243: `&`

meshmode/mesh/processing.py
```diff
@@ -62,6 +62,367 @@ def find_group_indices(groups, meshwide_elems):
...
+                                element_faces))
+                    adj_neighbor_parts = np.concatenate((adj_neighbor_parts,
+                                neighbor_parts))
+                    adj_global_neighbors = np.concatenate((adj_global_neighbors,
+                                global_neighbors))
+                    adj_neighbor_faces = np.concatenate((adj_neighbor_faces,
+                                neighbor_faces))
+                    nonlocal_adj_data[i_part_grp] = (adj_elements,
+                                adj_element_faces, adj_neighbor_parts,
+                                adj_global_neighbors, adj_neighbor_faces)
```
inducer, line 273: It might be worth making a `namedtuple` or `dataclass` (Py 3.7+) for this.

meshmode/mesh/processing.py
```diff
@@ -62,6 +62,367 @@ def find_group_indices(groups, meshwide_elems):
...
+        adj_indices = np.where(global_elem_to_part_elem[facial_adj.elements
+                    + elem_base] >= 0)[0]
+
+        if len(adj_indices) > 0:
+            part_elem_base = part_mesh_group_elem_base[i_part_grp]
+            elements = global_elem_to_part_elem[facial_adj.elements[adj_indices]
+                        + elem_base] - part_elem_base
+            element_faces = facial_adj.element_faces[adj_indices]
+            neighbors = facial_adj.neighbors[adj_indices]
+            bdry_data[i_part_grp] = (elements, element_faces, neighbors)
```
inducer, line 304: It might be worth making a `namedtuple` or `dataclass` (Py 3.7+) for this.

meshmode/mesh/processing.py
```diff
@@ -62,6 +62,367 @@ def find_group_indices(groups, meshwide_elems):
 
 # {{{ partition_mesh
 
+def _create_element_part_map(part_per_element, parts, element_id_dtype):
```
inducer, line 65:
```suggestion
def _compute_global_elem_to_part_elem(part_per_element, parts, element_id_dtype):
```
(for recognizability below)

meshmode/mesh/processing.py
```diff
@@ -62,6 +62,367 @@ def find_group_indices(groups, meshwide_elems):
...
+                else:
+                    # Next entry is a boundary
+                    ibdry = icombined - nnonlocal
+                    elements[imerged] = bdry_elements[ibdry]
+                    element_faces[imerged] = bdry_element_faces[ibdry]
+                    neighbor_parts[imerged] = -1
+                    neighbors[imerged] = bdry_neighbors[ibdry]
+                    neighbor_elements[imerged] = -1
+                    neighbor_faces[imerged] = -1
+                imerged = imerged + 1
```
inducer, line 413: Can this be done without looping? `np.cumsum` can build a running index.

meshmode/mesh/processing.py
```diff
@@ -83,183 +444,75 @@ def partition_mesh(mesh, part_per_element, part_num):
...
+    bdry_data = _collect_bdry_data(mesh, global_elem_to_part_elem, part_mesh_groups,
+                global_group_to_part_group, part_mesh_group_elem_base)
+
+    # Assemble full set of neighbors
+    all_neighbor_parts = set()
+    for adj in nonlocal_adj_data:
+        if adj is not None:
+            _, _, grp_neighbor_parts, _, _ = adj
+            for i_neighbor_part in grp_neighbor_parts:
+                all_neighbor_parts.add(i_neighbor_part)
```
inducer, line 483: Rewrite as set comprehension?

inducer:
- [ ] Grudge test against this branch: https://github.com/inducer/grudge/pull/6

meshmode/mesh/processing.py
majosm: This one didn't work for me:

```python
>               if adj_indices:
E               ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
```

meshmode/mesh/processing.py
```diff
@@ -62,6 +62,367 @@ def find_group_indices(groups, meshwide_elems):
 
 # {{{ partition_mesh
 
+def _create_element_part_map(part_per_element, parts, element_id_dtype):
+    """Create a map from global elem to partition elem for a set of partitions."""
+    n_elems_in_part = {ipart: 0 for ipart in parts}
```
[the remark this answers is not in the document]
majosm: At least for the way I'm currently using this function, `parts` is meant to be small (there are two existing calls; in one, `parts` contains only the current partition, and in the other it contains only the neighboring partitions). Would it still be more efficient in this case?

Maybe the variable naming is just confusing here? (`parts` -> `selected_parts`, or something like that?)

majosm:
I think that's everything from the previous round... how are we looking now?

inducer:
This breaks grudge: https://github.com/inducer/grudge/pull/6/checks?check_run_id=921128480

Here's how to understand the failure: Simply copying the relevant array array (using `array.copy()`) before passing it may address this. Suppose you have a 3x3 array `ary`, then the slice `a[:2,:2]` would be non-contiguous in memory. PyOpenCL can currently only transfer contiguous arrays to/from the device. Copying restores contiguity.

majosm:
@inducer CI is acting strangely... Any ideas?

> Reading package lists...
Building dependency tree...
Reading state information...
gfortran-7 is already the newest version (7.5.0-3ubuntu1~18.04).
gfortran-7 set to manually installed.
0 upgraded, 0 newly installed, 0 to remove and 33 not upgraded.
ln: failed to create symbolic link '/usr/bin/gfortran': File exists
##[error]Process completed with exit code 1.

inducer:
That looks like they (Github) may have messed with the CI container image. Let me take a look.

inducer:
Same problem on master. (https://github.com/inducer/meshmode/pull/11/checks?check_run_id=923601326) See #21.

inducer:
Could you merge in master to get the CI to pass?

inducer:
Thanks for working on this! LGTM, in it goes.

PR #15, opened 2020-07-13
Add test and implementation for three-argument-where

MTCam:
The purpose of this PR is to add some if_positive capability to the array-context numpy workalike.

### DIFF
--- a/meshmode/array_context.py
+++ b/meshmode/array_context.py
@@ -227,7 +227,7 @@ class _PyOpenCLFakeNumpyNamespace(_BaseFakeNumpyNamespace):
     @obj_array_vectorized_n_args
     def where(self, criterion, then, else_):
         import pyopencl.array as cl_array
-        return cl_array.if_positive(criterion.astype(np.bool), then, else_)
+        return cl_array.if_positive(criterion != 0, then, else_)
 
 
 class PyOpenCLArrayContext(ArrayContext):
--- a/meshmode/array_context.py
+++ b/meshmode/array_context.py
@@ -224,6 +224,11 @@ class _PyOpenCLFakeNumpyNamespace(_BaseFakeNumpyNamespace):
 
         return super().__getattr__(name)
 
+    @obj_array_vectorized_n_args
+    def where(self, criterion, then, else_):
+        import pyopencl.array as cl_array
+        return cl_array.if_positive(criterion.astype(np.bool), then, else_)
+
 
 class PyOpenCLArrayContext(ArrayContext):
     """
--- a/test/test_meshmode.py
+++ b/test/test_meshmode.py
@@ -1467,6 +1467,7 @@ def test_array_context_np_workalike(ctx_factory):
             ("arctan2", 2),
             ("minimum", 2),
             ("maximum", 2),
+            ("where", 3),
             ]:
         args = [np.random.randn(discr.ndofs) for i in range(n_args)]
         ref_result = getattr(np, sym_name)(*args)
--- a/test/test_meshmode.py
+++ b/test/test_meshmode.py
@@ -1478,6 +1478,17 @@ def test_array_context_np_workalike(ctx_factory):
 
         assert np.allclose(actx_result, ref_result)
 
+        n_args = 2
+        args = [np.random.randn(discr.ndofs) for i in range(n_args)]
+        ref_result = np.where(((args[0] - args[1]) > 0), args[0], args[1])
+
+        actx_args = [unflatten(actx, discr, actx.from_numpy(arg)) for arg in args]
+        actx_result = actx.to_numpy(
+            flatten(actx.np.where(((args[0] - args[1]) > 0), args[0], args[1]))
+        )
+
+        assert np.allclose(actx_result, ref_result)
+
 
 if __name__ == "__main__":
     import sys
--- a/test/test_meshmode.py
+++ b/test/test_meshmode.py
@@ -1483,9 +1483,10 @@ def test_array_context_np_workalike(ctx_factory):
         ref_result = np.where(((args[0] - args[1]) > 0), args[0], args[1])
 
         actx_args = [unflatten(actx, discr, actx.from_numpy(arg)) for arg in args]
-        actx_result = actx.to_numpy(
-            flatten(actx.np.where(((args[0] - args[1]) > 0), args[0], args[1]))
-        )
+        actx_cond = actx.from_numpy((args[0] - args[1]) > 0)
+        actx_result = actx.to_numpy(flatten(actx.np.where(actx_cond,
+                                                          actx_args[0],
+                                                          actx_args[1])))
 
         assert np.allclose(actx_result, ref_result)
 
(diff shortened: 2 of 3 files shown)


test/test_meshmode.py
```diff
@@ -1478,6 +1478,18 @@ def test_array_context_np_workalike(ctx_factory):
...
+        ref_result = np.where(((args[0] - args[1]) > 0), args[0], args[1])
+
+        actx_args = [unflatten(actx, discr, actx.from_numpy(arg)) for arg in args]
+        actx_cond = actx.from_numpy((args[0] - args[1]) > 0)
+        actx_result = actx.to_numpy(flatten(actx.np.where(actx_cond,
+                                                          actx_args[0],
+                                                          actx_args[1])))
+
+        assert np.allclose(actx_result, ref_result)
+
```
inducer, line 1492:
```suggestion
```

Simply adding `("where", 3),` to the list above should do the trick.

MTCam:
I would have thought that'd work too, however it did not - which is why I added a test dedicated to `where`.   Here's the error:
```
        for sym_name, n_args in [
                ("sin", 1),
                ("exp", 1),
                ("arctan2", 2),
                ("minimum", 2),
                ("maximum", 2),
                ("where", 3),
                ]:
            args = [np.random.randn(discr.ndofs) for i in range(n_args)]
            ref_result = getattr(np, sym_name)(*args)
    
            actx_args = [unflatten(actx, discr, actx.from_numpy(arg)) for arg in args]
    
            actx_result = actx.to_numpy(
>                   flatten(getattr(actx.np, sym_name)(*actx_args)))

test_meshmode.py:1478: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
<decorator-gen-17>:2: in where
    ???
../../emirge/pytools/pytools/obj_array.py:208: in obj_array_vectorize_n_args
    result[i] = f(*new_args)
../meshmode/array_context.py:230: in where
    return cl_array.if_positive(criterion.astype(np.bool), then, else_)
../../../../../miniforge3/envs/dgfem3/lib/python3.8/site-packages/pyopencl/array.py:1323: in astype
    result.add_event(self._copy(result, self, queue=queue))
../../../../../miniforge3/envs/dgfem3/lib/python3.8/site-packages/pyopencl/array.py:139: in kernel_runner
    knl = kernel_getter(*args, **kwargs)
../../../../../miniforge3/envs/dgfem3/lib/python3.8/site-packages/pyopencl/array.py:905: in _copy
    return elementwise.get_copy_kernel(
<decorator-gen-23>:2: in get_copy_kernel
    ???
../../../../../miniforge3/envs/dgfem3/lib/python3.8/site-packages/pyopencl/tools.py:101: in first_arg_dependent_memoize
    result = func(cl_object, *args)
../../../../../miniforge3/envs/dgfem3/lib/python3.8/site-packages/pyopencl/elementwise.py:485: in get_copy_kernel
    "tp_dest": dtype_to_ctype(dtype_dest),
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <pyopencl.compyte.dtypes.DTypeRegistry object at 0x104b8ca00>, dtype = dtype('bool')

    def dtype_to_ctype(self, dtype):
        if dtype is None:
            raise ValueError("dtype may not be None")
    
        dtype = np.dtype(dtype)
    
        try:
            return self.dtype_to_name[dtype]
        except KeyError:
>           raise ValueError("unable to map dtype '%s'" % dtype)
E           ValueError: unable to map dtype 'bool'

../../../../../miniforge3/envs/dgfem3/lib/python3.8/site-packages/pyopencl/compyte/dtypes.py:107: ValueError
======================================== short test summary info =========================================
```

meshmode/array_context.py
```diff
@@ -224,6 +224,11 @@ def __getattr__(self, name):
 
         return super().__getattr__(name)
 
+    @obj_array_vectorized_n_args
+    def where(self, criterion, then, else_):
+        import pyopencl.array as cl_array
+        return cl_array.if_positive(criterion.astype(np.bool), then, else_)
```
inducer, line 230:
```suggestion
        return cl_array.if_positive(criterion != 0, then, else_)
```

inducer, line 230:
```suggestion
        return cl_array.if_positive(criterion != 0, then, else_)
```

inducer: My mistake. And the error message is spot on, actually.
```
 ValueError: unable to map dtype 'bool'
```
Confusingly, `bool` has no defined  in-memory data format (not even a size) in OpenCL. That's why casting to `bool` doesn't do the trick.

MTCam:
Ugh, yes, I misread the error.  Fwiw, I did finally figure this out.   Your soln is better.  I did this in the test:
```
        if sym_name == "where":
            args[0] = args[1] - args[2]
            ref_result = np.where(args[0] > 0, args[1], args[2])
        else:
            ref_result = getattr(np, sym_name)(*args)
```
and changed the call inside the `where` dupe to this:

```
    @obj_array_vectorized_n_args
    def where(self, criterion, then, else_):
        return cl_array.if_positive(criterion, then, else_)
```

MTCam:
This fails CI because it somehow fails at pre-req install of pyopencl.

inducer:
Yeah, I substantially redid my [CI scripts](https://gitlab.tiker.net/inducer/ci-support/) yesterday, this is fallout from that. Pushed a fix.

inducer:
Thanks for working on this!

MTCam:
Thanks for your help and patience!  I don't mind admitting that the np work-alike facility mechanism remains sort of mysterious to me.

inducer:
In what way? Feel free to use this PR to point at code and ask questions.

inducer:
Also, likewise! If we're patient with each other, we'll get this done. :)

MTCam:
> In what way? Feel free to use this PR to point at code and ask questions.

I am just now seeing this comment. Thanks! and great idea.  

It is hard to know where to start. If I follow the chain of calls and instantiations for a given `actx.np.<stuff>` call - it is very hard for me to parse out how/when the actual call is made that "does the work".  The `where` function we've just added is an exception because there is an explicit call there.  For any others however, I trace it down to where it starts interacting with Loopy and I quickly get completely befuddled. 

Also, I know this will come with experience, but even though I can follow a good deal of the code and have some (albeit often cursory) understanding of what the individual bits and pieces are doing; it is far from obvious *why* those things are done.   That is, I have little "active knowledge" of how to do things in this framework.  Tweaking or bending things is one thing, but creating from scratch is quite another.

inducer:
> it is far from obvious why those things are done.

IMO: Thosre "why" questions are really important! And I think I can supply (at least some) reasons. Here's an idea: In the mirgecom docs, should we start a section for "FAYs" (Frequently Asked Whys :). You (or anyone) starts a PR for a question, I (or anyone) pushes an answer, you mergethe PR once you're satisfied that the question has been answered. And the Q/A pair sticks around for others to read/complain about.

MTCam:
Cool idea.  :)

PR #17, opened 2020-07-17
Allow getting mode IDs from the element group

inducer:
@MTCam Could you provide a review?


MTCam:
I apologize - just now combing through my missed mail and "filter" items and found this review request.  I guess it works :)

PR #19, opened 2020-07-23
 Add Visualization Support for VTK_LAGRANGE Elements

alexfikl:
Counterpart to [!85](https://gitlab.tiker.net/inducer/meshmode/-/merge_requests/85) to make sure CI passes.

PR #20, opened 2020-07-29
Add pvtu viz file writing util

MTCam:
Added a quick snippet for using existing code to generate the pvtu.   Working on testing it.


inducer:
It looks like this reverts #19.

MTCam:
I'm unsure how it did that.   I forked this less than an hour ago.

MTCam:
Turns out, when you have open changes that hang for a long time, they get out of sync with what you are trying to change.  Who knew?

MTCam:
Good news everyone; this works in mirgecom with a parallel vortex case. 

Less good, haven't been able to write an effective test inside meshmode for this capability.

meshmode/discretization/visualization.py
```diff
@@ -441,18 +441,25 @@ def write_high_order_vtk_file(self, file_name, names_and_fields,
                 overwrite=overwrite)
 
     def write_vtk_file(self, file_name, names_and_fields,
```
inducer, line 443: I think I'd prefer if the parallel version were a wrapper around this function (`write_distributed_vtk_file` alongside here maybe?) rather than more flag-based gunk stuffed into an already complex function body.

inducer, line 443: As is, there's the potential for inconsistency between the `file_name` argument and `par_namelist`. This would be avoided if both were generated from a supplied `.format()` pattern.

meshmode/discretization/visualization.py
```diff
@@ -529,6 +536,23 @@ def _write_vtk_file(self, file_name, names_and_fields, connectivity,
 
             generator(grid).write(outf)
 
+        if mpicomm is not None:
+            rank = mpicomm.Get_rank()
+            nproc = mpicomm.Get_size()
+            if nproc > 1:  # don't bother for serial runs
+                if rank == 0:
+                    if par_filename is None:
+                        par_filename = file_name+'.pvtu'
```
inducer, line 545: This'll generate file names ending in `.vtu.pvtu`. Maybe strip the `.vtu` before adding the `.pvtu`?

meshmode/discretization/visualization.py
```diff
@@ -441,18 +441,25 @@ def write_high_order_vtk_file(self, file_name, names_and_fields,
                 overwrite=overwrite)
 
     def write_vtk_file(self, file_name, names_and_fields,
-            compressor=None, real_only=False, overwrite=False):
+                       par_filename=None, par_namelist=None,
+                       compressor=None, mpicomm=None,
```
inducer, line 445: I'm not sure about passing in the MPI communicator. I'm guessing it's OK, since it's a soft dependency, and most uses of this will be through MPI.

Still, please document this argument as
```rst
needs to support ``mpi_comm.Get_rank`` and ``mpi_comm.Get_size`` (such as an instance of ``mpi4py.Comm``)
```
(I.e. make it duck typed.)

And please rename it `mpi_comm`.

meshmode/discretization/visualization.py
MTCam: Yeah, I'd like that.  The trouble is that I seem to need the grid with the data arrays set up into it as is done in the `write_vtk_file`.  Let me see if I can make it more palatable.

meshmode/discretization/visualization.py
```diff
@@ -520,7 +527,7 @@ def _write_vtk_file(self, file_name, names_and_fields, connectivity,
             if overwrite:
                 os.remove(file_name)
             else:
-                raise FileExistsError("output file '%s' already exists" % file_name)
+                raise FileExistsError(f'output file {file_name} already exists')
```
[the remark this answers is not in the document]
MTCam: Ah, yes, I see what I've done here.   I usually try to delineate these in my own codes.   Dunno what happened here - just spaced out.

MTCam:
I hope this helps clean up a great deal of the issues with the previous iteration.   It is still embedded in the existing API because it requires some expensive looking infrastructure that is already tackled by `write_vtk_file`.

inducer:
(See #21 for the CI failures.)

meshmode/discretization/visualization.py
```diff
@@ -441,18 +441,33 @@ def write_high_order_vtk_file(self, file_name, names_and_fields,
...
+        r"""
+        Parameters
+        ----------
+        par_namelist
+
+            If par_namelist is given by the caller (and non-empty),
+            it is expected to be a list of filenames wherein the
+            PVTU-format filename and the part source filenames are given.
+            When used by a parallel application, only one rank should
+            write the PVTU file.
```
inducer, line 455:
```suggestion
        :arg par_namelist: If par_namelist is given by the caller (and non-empty),
            it is expected to be a list of filenames wherein the
            PVTU-format filename and the part source filenames are given.
            When used by a parallel application, only one rank should
            write the PVTU file.
```
(My packages don't use the numpy docstring convention, at least not yet.)

meshmode/discretization/visualization.py
```diff
@@ -520,7 +535,8 @@ def _write_vtk_file(self, file_name, names_and_fields, connectivity,
             if overwrite:
                 os.remove(file_name)
             else:
-                raise FileExistsError("output file '%s' already exists" % file_name)
+                raise FileExistsError("output file '{%s}' already exists"
+                                      % file_name)
```
inducer, line 539:
```suggestion
                raise FileExistsError("output file '%s' already exists"
                                      % file_name)
```

meshmode/discretization/visualization.py
inducer: Fair.

meshmode/discretization/visualization.py
```diff
@@ -441,18 +441,33 @@ def write_high_order_vtk_file(self, file_name, names_and_fields,
                 overwrite=overwrite)
 
     def write_vtk_file(self, file_name, names_and_fields,
```
inducer, line 443: I like what you've done with the `par_namelist`, but I think the concern regarding inconsistencies between `file_name` and `par_namelist` (maybe `par_file_names`?) is real, and writing (redundant) code to generate both will get old. Want to write a wrapper that takes a (duck-typed) MPI communicator and a file name pattern (like `my-vis-{rank:05d}.vtu`)?

inducer:
Could you merge in master to get the CI to pass?

meshmode/discretization/visualization.py
MTCam: It would be my preference to leave the formation of the filenames to the caller.  I have a filename generator already in mirgecom.  Additionally, one doesn't  actually *need* an MPI communicator to write a PVTU file.  I would also dislike it if I cannot directly specify the filenames and direct a routine to write the filenames I specify.   Asking this (vtk file writing) routine to take some responsibility for forming filenames I think is beyond the scope of what it should do.

MTCam: Regarding our discussion this morning.  Your suggested solution is definitely more succinct, but it will require more arguments, and is still redundant.  Please see the fixed-up version to see if this is a decent compromise.

inducer:
Some proposed changes in https://github.com/MTCam/meshmode/pull/1.

MTCam:
Some friendly people helped me with this.  I think it is in-shape now.  Could you have a look @inducer?

inducer:
Thanks for working on this! I'll merge it as soon as CI passes.

test/test_meshmode.py
```diff
@@ -86,6 +88,67 @@ def test_circle_mesh(visualize=False):
...
+    pvtu_filename = file_name_pattern.format(rank=0).replace("vtu", "pvtu")
+
+    vis.write_parallel_vtk_file(
+            FakeComm(),
+            file_name_pattern,
+            [
+                ("scalar", discr.zeros(actx)),
+                ("vector", make_obj_array([discr.zeros(actx) for i in range(dim)]))
+                ],
+            overwrite=True)
```
MTCam, line 143: Nice

PR #30, opened 2020-08-10
Treat boundary_tags as a list consistently

majosm:
Documentation specifies `boundary_tags` as a tuple, but it's used in most places as a list ([example](https://github.com/inducer/meshmode/blob/6cc0de2fec2e39e25dea38035171a74d6021916a/meshmode/mesh/__init__.py#L726-L737)). This makes it officially a list.

cc @MTCam


inducer:
Thanks!

PR #34, opened 2020-08-13
warn when creating PyOpenCLArrayContext without allocator

matthiasdiener:


meshmode/array_context.py
```diff
@@ -250,7 +250,13 @@ def __init__(self, queue, allocator=None):
         super().__init__()
         self.context = queue.context
         self.queue = queue
-        self.allocator = allocator
+        self.allocator = allocator if allocator else None
```
inducer, line 253: I'm not sure `allocator=False` actually disables the warning.

meshmode/array_context.py
```diff
@@ -250,7 +250,13 @@ def __init__(self, queue, allocator=None):
...
+        if allocator is None:
+            from warnings import warn
+            warn("PyOpenCLArrayContext created without an allocator. "
+                 "This can lead to high numbers of memory allocations. "
+                 "Please consider using a pyopencl.tools.MemoryPool. "
+                 "Run with allocator=False to disable this warning.")
```
inducer, line 259: Also needs a doc update to describe the `allocator=False` argument.

meshmode/array_context.py
matthiasdiener: I just tested it again and it did disable the warning. Why would it not?

meshmode/array_context.py
matthiasdiener: I put a draft doc there.

meshmode/array_context.py
```diff
@@ -244,13 +244,24 @@ class PyOpenCLArrayContext(ArrayContext):
         A :class:`pyopencl.CommandQueue`.
 
     .. attribute:: allocator
+        A PyOpenCL memory allocator. Can also be `None` (default) or `False` to
+        use the default allocator. Please note that running with the default
+        allocator will cause large numbers of memory allocations, and will
+        print a warning at runtime. Specify `False` to avoid the warning.
+        Please consider using a pyopencl.tools.MemoryPool.
```
inducer, line 251:
```suggestion
        A PyOpenCL memory allocator. Can also be `None` (default) or `False` to
        use the default allocator. Please note that running with the default
        allocator allocates and deallocates OpenCL buffers directly. If lots
        of arrays are created (e.g. as results of computation), the associated cost
        may become significant. Using e.g. :class:`pyopencl.tools.MemoryPool`
        as the allocator can help avoid this cost.
```

inducer:
Thanks!

PR #38, opened 2020-08-15
Add conj to the ArrayContext numpy workalike

alexfikl:
Added to handle an `DOFArray` in [pytential](https://github.com/inducer/pytential/blob/master/pytential/symbolic/execution.py#L287).

As far as I can tell, this should just work™, but the numpy ufunc stuff looks for `conjugate` and the `cl.array.Array` only has a `conj` (i.e. this works fine for object arrays of `ndarray`). This kind of sounds like a bug, since at the toplevel `np.conj` is an alias of `np.conjugate`.

### DIFF
--- a/meshmode/array_context.py
+++ b/meshmode/array_context.py
@@ -72,6 +72,15 @@ class _BaseFakeNumpyNamespace:
 
         return obj_array_vectorized_n_args(f)
 
+    @obj_array_vectorized_n_args
+    def conjugate(self, x):
+        # NOTE: conjugate distribute over object arrays, but it looks for a
+        # `conjugate` ufunc, while some implementations only have the shorter
+        # `conj` (e.g. cl.array.Array), so this should work for everybody
+        return x.conj()
+
+    conj = conjugate
+
 
 class ArrayContext:
     """An interface that allows a :class:`Discretization` to create and interact with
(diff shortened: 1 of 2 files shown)


meshmode/array_context.py
```diff
@@ -221,6 +221,10 @@ def __getattr__(self, name):
         if name in ["minimum", "maximum"]:
             import pyopencl.array as cl_array
             return obj_array_vectorized_n_args(getattr(cl_array, name))
+        elif name in ["conj", "conjugate"]:
+            # NOTE: conj/conjugate distribute over object arrays.. but only
+            # for object array of ndarrays (at least up to of numpy-1.19)
+            return obj_array_vectorized_n_args(lambda x: x.conj())
```
inducer, line 227: This should go in `_BaseFakeNumpyNamespace`, not here, as there's nothing CL-specific. Also, avoid `if`/`else` name lookup, that's slow! (just `def` it as a method) Maybe `minimum`/`maximum` should be implemented that way too, to avoid setting a bad precedent. The code duplication is negligible, after all.

alexfikl: Updated in a6668b7.

Took the liberty to also change `maximum` and `minimum` as I was in there.

inducer:
For my understanding: Would https://github.com/inducer/pyopencl/pull/361 also address this?

alexfikl:
> For my understanding: Would [inducer/pyopencl#361](https://github.com/inducer/pyopencl/pull/361) also address this?

I think so, yes. I tried just creating an `Array` and then monkey patching it with `x.conjugate = x.conj` and that worked.

inducer:
Sounds good. Let's do this anyway, if only for the minimum/maximum changes.

PR #39, opened 2020-08-20
Fix partition connection batches

### LINKED ISSUE #36: Distributed setup calls _make_cross_face_batches once per face
[Here](https://github.com/inducer/meshmode/blob/2f84174f03ec9164c3c314b9eb9cdb2c4bd41db7/meshmode/discretization/connection/opposite_face.py#L443-L522). It definitely shouldn't. As the name possibly suggests, `_make_cross_face_batches` is supposed to be a batched routine. (called once per rank pair) @MTCam spotted this when doing some parallel runs.

This is actually pretty bad, since it creates lots of small interpolation batches rather than one big one.

inducer:
Closes #36 

cc @majosm

### DIFF
--- a/meshmode/discretization/connection/direct.py
+++ b/meshmode/discretization/connection/direct.py
@@ -72,7 +72,7 @@ class InterpolationBatch(object):
 
     .. attribute:: to_element_face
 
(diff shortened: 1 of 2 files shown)


inducer:
Superseded by #40.

PR #40, opened 2020-08-26
Fix partition connection batches

### LINKED ISSUE #36: Distributed setup calls _make_cross_face_batches once per face

majosm:
Closes #36

I tried the approach outlined in #39 but ran into some trouble. Namely, I wasn't sure how to map the remote volume element indices stored in the local partition's adjacency data to the remote boundary discretization elements. I'm able to do it the other way around (local volume elements in remote partition's adjacency -> local boundary discretization elements) because I have access to `local_bdry_conn` and `_make_bdry_el_lookup_table`.

Note: Both the original code and this change assume that the entries in `remote_adj_groups[i_remote_grp]` (filtered by face) and the entries in `remote_from_elem_indices[i_remote_grp][iface]` have the same ordering. Is this guaranteed?


inducer:
I pushed some (purely stylistic) (what I view as) improvements. I think it's possible to maintain the original loop style in which the outermost loop is over local groups. I think the reason why this was difficult is because the remote partition's `InterPartitionAdjacencyGroup` is being passed in, while the local one is sitting right there, already part of the local mesh. So I propose that
- we don't even transfer the remote `InterPartitionAdjacencyGroup` (in `meshmode.distributed`) at all, and
- we change the logic in `make_partition_connection` to make use of the local `InterPartitionAdjacencyGroup`.

inducer:
So I ended up staring at this for a long time. I still have no idea why the original code worked, at all. In trying to understand why this worked, I got increasingly annoyed, and I ended up rewriting it in a way that's (hopefully) more straightforwardly correct. I'm guessing the lesson here is that, probably, I should have stopped as soon as the code didn't make any sense and asked you to rewrite it then.

Oh well, at least I understand this code better now. :) Thanks for your work on it.

PR #45, opened 2020-08-27
Deprecate DirectDiscretizationConnection.full_resample_matrix

alexfikl:
Fixes https://gitlab.tiker.net/inducer/meshmode/-/issues/13

### DIFF
--- a/meshmode/discretization/connection/direct.py
+++ b/meshmode/discretization/connection/direct.py
@@ -171,8 +171,6 @@ class DirectDiscretizationConnection(DiscretizationConnection):
 
     .. automethod:: __call__
 
-    .. automethod:: full_resample_matrix
-
     """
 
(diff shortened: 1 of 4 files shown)


meshmode/discretization/connection/direct.py
```diff
@@ -412,4 +348,84 @@ def pick_knl():
 # }}}
 
 
+# {{{ dense resampling matrix
+
+def make_direct_full_resample_matrix(actx, conn):
```
inducer, line 353: Did you mean to leave this out of the module docstring?

alexfikl: Kind of, given all the warnings for its usage. I can add it back though, if you want.

inducer: No, that's OK. Just wanted to make sure it was intentional.

inducer:
Thanks!

PR #49, opened 2020-09-05
Add missing space in array context docs

alexfikl:
Another silly one-liner that fixes the rendering [here](https://documen.tician.de/meshmode/array.html#meshmode.array_context.PyOpenCLArrayContext.allocator).


inducer:
Thanks!

PR #50, opened 2020-09-08
Fix num_reference_derivative for multiple groups

alexfikl:
Managed to hit this in [pytential](https://github.com/inducer/pytential/blob/master/pytential/symbolic/execution.py#L219) when using multiple groups. That passed in a generator that got consumed by the first group.

### DIFF
--- a/test/test_meshmode.py
+++ b/test/test_meshmode.py
@@ -1541,6 +1541,13 @@ def test_mesh_multiple_groups(actx_factory, ambient_dim, visualize=False):
             error = flat_norm(bdry_f - em_bdry_f)
             assert error < 1.0e-11, error
 
+    # check some derivatives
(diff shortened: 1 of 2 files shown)


test/test_meshmode.py
```diff
@@ -1541,6 +1541,13 @@ def test_mesh_multiple_groups(actx_factory, ambient_dim, visualize=False):
             error = flat_norm(bdry_f - em_bdry_f)
             assert error < 1.0e-11, error
 
+    # check some derivatives
```
inducer, line 1544:
```suggestion
    # check some derivatives (nb: flatten is a generator)
```

inducer (commented):
Thanks, LGTM. I'll apply my one minor suggestion (which I'll apply).

PR #51, opened 2020-09-11
Update write_vtk so it can handle empty data fields

MTCam:
I would like to tweak `write_vtk_file` slightly so that it can handle the situation where the name_and_fields list contains an empty/None field array.

Example:
```python
# ConservedVars dataclass has 
# mass : np.ndarray = something
# energy : np.ndarray = something
# momentum : np.ndarray = something
# mass_fractions : np.ndarray = None
names_and_fields = [("cv", ConservedVars), ("foo", Bar)]
```

This PR fixes the write-time error experienced by _mirgecom_ when writing singe-component fluid solutions.

### DIFF
--- a/meshmode/discretization/visualization.py
+++ b/meshmode/discretization/visualization.py
@@ -550,7 +550,7 @@ class Visualizer(object):
                             getattr(fld, dclass_field.name))
                         for dclass_field in dataclasses.fields(fld)
                         if getattr(fld, dclass_field.name) is not None)
-            else if fld is not None:
+            elif fld is not None:
                 new_names_and_fields.append((name, fld))
 
         names_and_fields = new_names_and_fields
--- a/meshmode/discretization/visualization.py
+++ b/meshmode/discretization/visualization.py
@@ -540,12 +540,11 @@ class Visualizer(object):
         for name, fld in names_and_fields:
             if hasattr(type(fld), "__dataclass_fields__"):
                 import dataclasses
-                for dclass_field in dataclasses.fields(fld):
-                    field = getattr(fld, dclass_field.name)
-                    if field is not None:
-                        new_names_and_fields.append(
-                            (f"{name}_{dclass_field.name}", field)
-                        )
+                new_names_and_fields.extend(
+                        (f"{name}_{dclass_field.name}",
+                            getattr(fld, dclass_field.name))
+                        for dclass_field in dataclasses.fields(fld)
+                        if getattr(fld, dclass_field.name) is not None)
             else:
                 new_names_and_fields.append((name, fld))
 
--- a/meshmode/discretization/visualization.py
+++ b/meshmode/discretization/visualization.py
@@ -550,7 +550,7 @@ class Visualizer(object):
                             getattr(fld, dclass_field.name))
                         for dclass_field in dataclasses.fields(fld)
                         if getattr(fld, dclass_field.name) is not None)
-            else:
+            else if fld is not None:
                 new_names_and_fields.append((name, fld))
 
         names_and_fields = new_names_and_fields
--- a/meshmode/discretization/visualization.py
+++ b/meshmode/discretization/visualization.py
@@ -496,6 +496,11 @@ class Visualizer(object):
             *value* may also be a data class (see :mod:`dataclasses`),
             whose attributes will be inserted into the visualization
             with their names prefixed by *name*.
+            If *value* is *None*, then there is no data to write and the
+            corresponding *name* will not appear in the data file.
+            If *value* is *None*, it should be *None* collectively across all
+            ranks for parallel writes; otherwise the behavior of this routine
+            is undefined.
         :arg overwrite: If *True*, silently overwrite existing
             files.
         :arg use_high_order: Writes arbitrary order Lagrange VTK elements.
--- a/meshmode/discretization/visualization.py
+++ b/meshmode/discretization/visualization.py
@@ -540,10 +540,12 @@ class Visualizer(object):
         for name, fld in names_and_fields:
             if hasattr(type(fld), "__dataclass_fields__"):
                 import dataclasses
-                new_names_and_fields.extend(
-                        (f"{name}_{dclass_field.name}",
-                            getattr(fld, dclass_field.name))
-                        for dclass_field in dataclasses.fields(fld))
+                for dclass_field in dataclasses.fields(fld):
+                    field = getattr(fld, dclass_field.name)
+                    if field is not None:
+                        new_names_and_fields.append(
+                            (f"{name}_{dclass_field.name}", field)
+                        )
             else:
                 new_names_and_fields.append((name, fld))
 


meshmode/discretization/visualization.py
```diff
@@ -543,7 +543,8 @@ def write_vtk_file(self, file_name, names_and_fields,
                 new_names_and_fields.extend(
                         (f"{name}_{dclass_field.name}",
                             getattr(fld, dclass_field.name))
-                        for dclass_field in dataclasses.fields(fld))
+                        for dclass_field in dataclasses.fields(fld)
+                        if getattr(fld, dclass_field.name) is not None)
```
inducer, line 547: Add this behavior to the documentation.

inducer, line 547: One particular complication: what happens if the fields are present on *some* ranks but not others in the distributed-memory case?

inducer, line 547: There's an inconsistency here between the treatment of the dataclasses and `names_and_fields`: one can tolerate `None`, the other can't. Is there a principled reason why?

MTCam: This should not happen on any of the simulations we plan.  We'll either use a mixture or not.  If the gas is a mixture, then *all* ranks will have non-None for the mass fractions.

inducer: Well, then this expectation should be documented. (A good way to phrase this "If (condition), behavior is undefined."

MTCam: Actually, what happens is that the rank-specific file will contain the data, ranks that do not have the data won't have it in the file.   Paraview will barf on this - but maybe one could write post-processors for that data.

MTCam: I'm not sure what to say about that.   As you know, if I run a single component fluid, then I won't have any mass fractions and the CV dataclass needs to support that (i.e. *if* we want mass fractions in CV (where it belongs since it is formally a conserved variable).  But I don't think that has anything to do with your question, if I am following properly.

What I *think* you are asking is - why shouldn't it just be handled fluidly instead of this arguably hokey post-processing where the names_and_fields needs to be edited on the inside of the I/O routine.   Ideally, mass_fractions wouldn't be added to the names_and_fields io list except in cases where it was actually present.   

As we have it, we are trying to strike a balance between what our developers need to pay attention to, and what can be done behind-the-scenes.  This one-line edit seems to be a reasonable approach because otherwise the onus falls on the _mirgecom_ application side where a fairly substantial re-think of the CV dataclass and mixture support needs to be reckoned with.

inducer: I understand all that. What I'm not happy with is that dataclasses allow `None`, but the top-level "fields" do not. If the `None` is OK inside a dataclass, why is this not also OK?
```
[
("velocity", None),
]
```
Consistency is important.

MTCam: I realize my previous rambling doesn't address your question.  But again, I'm not sure what to say - so I'll ramble some more.   

There is not a good reason this inconsistency has to be so.  The fact that it is easy to handle at the level where we edit the `names_and_fields` list (which we are already doing here), is beside the point.   I think if *value* is allowed to be a dataclass, then this routine needs to deal with it (i.e. fixing up `names_and_fields` depending what's in the dataclass), or the handling of it will need to be propagated down through subsequent constructs.

MTCam: I was trying to take an approach that prevents any of the subsequent machinery from needing to deal with *None* fields, and ensure it does not even appear in the viz file.

FYI, here's how it fails:

```
  File "lump-mpi.py", line 155, in <module>
    main()
  File "lump-mpi.py", line 133, in main
    advance_state(rhs=my_rhs, timestepper=timestepper,
  File "/usr/WS1/xpacc/Users/mtcampbe/CEESD/Quartz/Source/emirge.master/mirgecom/mirgecom/steppers.py", line 77, in advance_state
    checkpoint(state=state, step=istep, t=t, dt=dt)
  File "lump-mpi.py", line 126, in my_checkpoint
    return sim_checkpoint(discr, visualizer, eos, q=state,
  File "/usr/WS1/xpacc/Users/mtcampbe/CEESD/Quartz/Source/emirge.master/mirgecom/mirgecom/simutil.py", line 134, in sim_checkpoint
    visualizer.write_parallel_vtk_file(
  File "/usr/WS1/xpacc/Users/mtcampbe/CEESD/Quartz/Source/emirge.master/meshmode/meshmode/discretization/visualization.py", line 468, in write_parallel_vtk_file
    self.write_vtk_file(
  File "/usr/WS1/xpacc/Users/mtcampbe/CEESD/Quartz/Source/emirge.master/meshmode/meshmode/discretization/visualization.py", line 556, in write_vtk_file
    names_and_fields = [
  File "/usr/WS1/xpacc/Users/mtcampbe/CEESD/Quartz/Source/emirge.master/meshmode/meshmode/discretization/visualization.py", line 557, in <listcomp>
    (name, resample_to_numpy(self.connection, fld))
  File "/usr/WS1/xpacc/Users/mtcampbe/CEESD/Quartz/Source/emirge.master/meshmode/meshmode/discretization/visualization.py", line 87, in resample_to_numpy
    resampled = conn(vec)
  File "/usr/WS1/xpacc/Users/mtcampbe/CEESD/Quartz/Source/emirge.master/meshmode/meshmode/discretization/connection/direct.py", line 258, in __call__
    raise TypeError("non-array passed to discretization connection")
TypeError: non-array passed to discretization connection
```

MTCam: > I understand all that. What I'm not happy with is that dataclasses allow `None`, but the top-level "fields" do not. If the `None` is OK inside a dataclass, why is this not also OK?
> 
> ```
> [
> ("velocity", None),
> ]
> ```
> 
> Consistency is important.

I think I just spiraled out of control trying to answer this.  If the mass_fractions are *None* - this is fine inside our solvers.  Not all fluids will have multiple species.  This is consistent throughout the _mirgecom_ side. 

The trouble occurs when the I/O substrate automatically generates a names_and_fields list for I/O assuming that everything in the dataclass has associated data assigned.

MTCam: > I understand all that. What I'm not happy with is that dataclasses allow `None`, but the top-level "fields" do not. If the `None` is OK inside a dataclass, why is this not also OK?
> 
> ```
> [
> ("velocity", None),
> ]
> ```
> 
> Consistency is important.

I think I have fixed up the inconsistency about which you are concerned.  Sigh :)   Can you please have another look?

inducer: Yes, functionally, that's what I meant. Stylistically, I kind of still dislike that there are two copies of the `None` handling when one would suffice (at the cost of another copy of the list), but I'll let that go.

inducer:
LGTM. Thanks for your patience!

MTCam:
> LGTM. Thanks for your patience!

Of course! Thank you for yours!

PR #53, opened 2020-09-29
Remove imports of FileExistsError

alexfikl:
That was just added for python 2/3 compatibility.

### DIFF
--- a/meshmode/__init__.py
+++ b/meshmode/__init__.py
@@ -24,6 +24,7 @@ THE SOFTWARE.
 __doc__ = """
 .. exception:: Error
 .. exception:: DataUnavailable
+.. exception:: FileExistsError
 """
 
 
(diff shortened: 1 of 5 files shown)


meshmode/__init__.py
```diff
@@ -34,6 +33,3 @@ class Error(RuntimeError):
 
 class DataUnavailable(Error):
     pass
-
-
-from builtins import FileExistsError  # noqa: F401
```
inducer, line 39: Not sure I'm wild about this. It might break downstream code. Could you create grudge/pytential test balloons to check whether they get broken?

alexfikl: Oops! `grudge` seems to use this, so it most definitely will break. I'll just add it back.

alexfikl:
Looks like something broke with the `pyvisfile` update too.

inducer (approved):

PR #54, opened 2020-09-29
Allow generation of X-structured meshes

inducer:
cc @MTCam @lukeolson


inducer:
See https://github.com/inducer/pyvisfile/pull/7 for the (fix for) the failures.

meshmode/mesh/generation.py
```diff
@@ -690,6 +691,30 @@ def generate_box_mesh(axis_coords, order=1, coord_dtype=np.float64,
...
+            _______
+            |\    |
+            | \   |
+            |  \  |
+            |   \ |
+            |    \|
+            ^^^^^^^
+
+        Specifying a value other than *None* for all other mesh
+        types is an error.
```
inducer, line 717:
```suggestion
        dimensionalities and element types is an error.
```

PR #55, opened 2020-10-08
Nodal-DG interop

inducer:
cc @lukeolson @MTCam


inducer:
This is essentially done. It just needs a bit of CI engineering to make sure it keeps getting tested.

cc @majosm (sorry, forgot you at first)

PR #58, opened 2020-10-13
Remove stateful usages of arrays managed by ArrayContext

kaushikcfd:


meshmode/discretization/connection/direct.py
```diff
@@ -306,41 +366,73 @@ def pick_knl():
...
+                kwargs[f"to_element_indices_{i_batch}"] = (
+                    batch.to_element_indices)
+
+                kernels.append(knl)
+
+            fused_knl = lp.fuse_kernels(kernels)
+            # order of operations doesn't matter
+            fused_knl = lp.add_nosync(fused_knl, "global", "writes:result",
+                                      "writes:result", bidirectional=True,
+                                      force=True)
```
kaushikcfd, line 425: Downside: the concurrency in the kernel is no longer realized by PyOpenCLArrayContext.transform_loopy_program.

inducer: True. I think  a good approach to this would be to revive https://gitlab.tiker.net/inducer/loopy/-/merge_requests/261 and appropriately tag the inames so that we can make the transform logic generic.

thomasgibson:
Closing since https://github.com/inducer/meshmode/pull/192 took care of this

PR #59, opened 2020-10-13
Rename group_factory to group_cls in mesh generation

### LINKED ISSUE #57: meshmode.mesh.generation talks about group_factory, but means something different than what's defined in poly_element
https://github.com/inducer/meshmode/pull/56#discussion_r503467596

alexfikl:
Fixes #57

### DIFF
--- a/meshmode/mesh/generation.py
+++ b/meshmode/mesh/generation.py
@@ -25,6 +25,7 @@ import numpy as np
 import numpy.linalg as la
 import modepy as mp
 
+from functools import wraps
 import logging
 logger = logging.getLogger(__name__)
 
@@ -70,6 +71,28 @@ Tools for Iterative Refinement
 
 .. autofunction:: warp_and_refine_until_resolved
 """
+
+
+def _deprecate_kwargs(oldkey, newkey):
+    def super_wrapper(func):
+        @wraps(func)
+        def wrapper(*args, **kwargs):
+            if oldkey in kwargs:
+                from warnings import warn
+                warn(f"The '{oldkey}' keyword is deprecated and will be removed. "
+                        f"Use '{newkey}' instead.",
+                        DeprecationWarning, stacklevel=2)
+
+                if newkey in kwargs:
+                    raise ValueError(f"Cannot use both '{oldkey}' and '{newkey}'.")
+
+                kwargs[newkey] = kwargs[oldkey]
+                del kwargs[oldkey]
+
+            return func(*args, **kwargs)
+        return wrapper
+
+    return super_wrapper
 
 
 # {{{ test curve parametrizations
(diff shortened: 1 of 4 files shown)


meshmode/mesh/generation.py
```diff
@@ -71,6 +72,26 @@
 .. autofunction:: warp_and_refine_until_resolved
 """
 
+def _deprecate_kwargs(oldkey, newkey):
```
alexfikl, line 75: I copy pasted this from [pytential](https://github.com/inducer/pytential/blob/master/pytential/symbolic/primitives.py#L227). Should I add it to `pytools`? It seems like it could be useful.

setup.py
```diff
@@ -48,6 +48,7 @@ def main():
               "pytest>=2.3",
               "loo.py>=2014.1",
               "recursivenodes",
+              "dataclasses;python_version<=3.6",
```
alexfikl, line 51: Unrelated, but dataclasses are used in a few places, so should be here, right?

meshmode/mesh/generation.py
inducer: Yeah, I'm in favor of that. Thanks for suggesting it.

setup.py
inducer: Huh, TIL you could `;python_version<=3.6`. Also, yes, absolutely.

inducer (approved):
Thanks! LGTM, but I imagine it makes more sense to yoink `deprecate_kwargs` before merging.

meshmode/mesh/generation.py
alexfikl: Done in inducer/pytools#43.

setup.py
alexfikl: Yeah, you can also tell it [OS versions and other things](https://www.python.org/dev/peps/pep-0508/#environment-markers)!

nchristensen: Cool, I just tried this in pytools but I found it broke the package install unless I put the version number in quotes like `"dataclasses>=0.7;python_version<='3.6'"`.

alexfikl: I found that out the hard way too! (the snippet on top is outdated)

alexfikl:
@inducer This is currently failing because it doesn't pick up the new `pytools`, so I'm guessing that needs to be released on PyPI/conda first?

inducer:
Did that this morning, rerunning: https://github.com/conda-forge/pytools-feedstock/pull/29

alexfikl:
> Did that this morning, rerunning: [conda-forge/pytools-feedstock#29](https://github.com/conda-forge/pytools-feedstock/pull/29)

That did the trick! Thank you.

inducer:
Thx!

PR #60, opened 2020-10-14
Add wait_event_queue_legnth feature to PyOpenCLArrayContext

inducer:
This should make it possible to merge https://github.com/inducer/grudge/pull/23 unchanged.

cc @nchristensen fischer1 matthiasdiener

### DIFF
--- a/meshmode/array_context.py
+++ b/meshmode/array_context.py
@@ -264,11 +264,38 @@ class PyOpenCLArrayContext(ArrayContext):
         as the allocator can help avoid this cost.
     """
 
-    def __init__(self, queue, allocator=None):
+    def __init__(self, queue, allocator=None, wait_event_queue_legnth=None):
+        r"""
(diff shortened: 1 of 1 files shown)


meshmode/array_context.py
```diff
@@ -264,12 +264,39 @@ class PyOpenCLArrayContext(ArrayContext):
         as the allocator can help avoid this cost.
     """
 
-    def __init__(self, queue, allocator=None):
+    def __init__(self, queue, allocator=None, wait_event_queue_legnth=None):
```
alexfikl, line 267: `legnth -> length` :scream:

inducer: Argh! Good catch. Thanks. I'll fix this in #61

inducer: https://github.com/inducer/meshmode/pull/61/commits/913e2c6dce81e9a5a9ca0ddef76328ad26560ec7

PR #61, opened 2020-10-14
Drastically limit wait_for_event_queue_length for pytest runs

inducer:
cc @nchristensen

### DIFF
--- a/meshmode/array_context.py
+++ b/meshmode/array_context.py
@@ -291,7 +291,7 @@ class PyOpenCLArrayContext(ArrayContext):
         self.queue = queue
         self.allocator = allocator if allocator else None
 
-        if wait_event_queue_length is not None:
+        if wait_event_queue_length is None:
             wait_event_queue_length = 10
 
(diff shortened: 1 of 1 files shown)


meshmode/array_context.py
```diff
@@ -264,37 +264,37 @@ class PyOpenCLArrayContext(ArrayContext):
...
         """
         super().__init__()
         self.context = queue.context
         self.queue = queue
         self.allocator = allocator if allocator else None
 
-        if wait_event_queue_legnth is not None:
-            wait_event_queue_legnth = 10
+        if wait_event_queue_length is not None:
+            wait_event_queue_length = 10
```
alexfikl, line 295: Is this intended? It seems like the length is either `None` or `10`?

inducer: D'oh! Lack of sleep?

inducer: 08c2e09

inducer: Also, thanks for catching that!

PR #67, opened 2020-10-23
Skip nodal-dg test when oct2py is not installed

alexfikl:


inducer:
```
FileNotFoundError: [Errno 2] No such file or directory: '/home/runner/work/meshmode/meshmode/test/nodal-dg-master/Codes1.1/CFD2D/CouetteIC2D.m'
```
The issue might be that tests get executed in parallel.

alexfikl:
> The issue might be that tests get executed in parallel.

Yeah, that sounds about right. Thanks for the hint! I added back the `git clone` in the CI, so it should work now.

inducer:
Nah, I'm not a huge fan of having redundant mechanisms to create these checkouts. Simply make it atomic by unzipping in a temporary folder and then moving it into place, taking care to tolerate the folder already being there.

alexfikl:
> Nah, I'm not a huge fan of having redundant mechanisms to create these checkouts. Simply make it atomic by unzipping in a temporary folder and then moving it into place, taking care to tolerate the folder already being there.

Fair enough. I'll get on it!

alexfikl:
@inducer Tests managed to pass and it at least looks less racy now. What do you think?

inducer:
I agree with "less racy". :) LGTM, thanks!

PR #71, opened 2020-10-31
Fix torus png file

alexfikl:
Looks like 5aff7bf corrupted the png file.


inducer:
Whoops. Sorry, thanks for spotting that.

PR #73, opened 2020-11-04
Put download_nodal_dg_if_not_present inside nodal_dg module so it can be used elsewhere

majosm:


inducer (approved):
Thx! I'll just wait for CI to pass.

PR #76, opened 2020-11-10
WIP: DOF tagging

nchristensen:


nchristensen:
Odd, it cannot find `mpi4py` or `firedrake`

```
SKIPPED [1] test_firedrake_interop.py:52: could not import 'firedrake': No module named 'firedrake'
SKIPPED [2] test_chained.py:114: implementation detail
SKIPPED [2] test_chained.py:143: implementation detail
SKIPPED [4] test_partition.py:514: could not import 'mpi4py': No module named 'mpi4py'
```

inducer:
Those are not the failures it's complaining about, though.

.ci/install-for-firedrake.sh
```diff
@@ -19,5 +19,5 @@ pip install dataclasses
 pip install pytest
 
 pip install -r /tmp/myreq.txt
-pip install --force-reinstall git+https://github.com/inducer/loopy.git@firedrake-usable_for_potentials
+pip install --force-reinstall git+https://github.com/nchristensen/loopy.git@test-firedrake-usable
```
nchristensen, line 22: The firedrake tests keep failing because they are looking for pytools 2020.4.4 which does not exist. I figured it this branch might need its requirements.txt updated but changing it on this branch seems not to help.

setup.py
```diff
@@ -45,11 +45,13 @@ def main():
               "modepy>=2020.2",
               "gmsh_interop",
               "pytools>=2020.4.1",
+              #"pytools>=2020.4.4",
               "pytest>=2.3",
 
               # 2019.1 is required for the Firedrake CIs, which use an very specific
               # version of Loopy.
               "loopy>=2019.1",
```
nchristensen, line 53: This may be problematic. Meshmode now requires Loopy version 2020.2.2 to function but the Firedrake tests require Loopy 2019.1. I think unless we backport changes to 2019.1 the Firedrake tests will fail.

meshmode/dof_array.py
```diff
@@ -50,6 +53,18 @@
 """
 
 
+# {{{ DOFTag
+
+class DOFTag(Tag):
```
inducer, line 58:
```suggestion
class IsDOFArray(Tag):
```

meshmode/dof_array.py
```diff
@@ -50,6 +53,18 @@
...
+    """A tag to mark arrays of DOFs in Loopy kernels. Applications
```
inducer, line 59:
```suggestion
    """A tag to mark arrays of DOFs in :mod:`loopy` kernels. Applications
```

meshmode/dof_array.py
```diff
@@ -50,6 +53,18 @@
...
+    could use this to decide how to change the memory layout of
+    these arrays.
+    """
+    pass
+
+    # }}}
```
inducer, line 65: Also use this in the transform logic.

nchristensen: Added some checks in transform_loopy_program.

nchristensen:
@inducer, currently the context factory only works with a PyOpenCLArrayContext which is a problem in Grudge because Grudge needs a GrudgeArrayContext. Should the context factory be modified to accept a context type as an argument or do we expect every array context subclass to have its own version of this? https://github.com/nchristensen/meshmode/blob/1849c35a2d2fadf04da05ed71937e839930e9d87/meshmode/array_context.py#L604

inducer:
>  currently the context factory only works with a PyOpenCLArrayContext 

What's a context factory? Sorry if I'm being dense.

nchristensen:
> What's a context factory? Sorry if I'm being dense.

Sorry, I should have been more clear. It is the ArrayContextFactory defined in this chunk of code. The difficulty is `PyOpenCLArrayContext` is currently hardcoded into the class. When Grudge calls this function in the tests it this means a `PyOpenCLArrayContext` is created instead of a `GrudgeArrayContext`.

```
def pytest_generate_tests_for_pyopencl_array_context(metafunc):
    """Parametrize tests for pytest to use a :mod:`pyopencl` array context.
    Performs device enumeration analogously to
    :func:`pyopencl.tools.pytest_generate_tests_for_pyopencl`.
    Using the line:
    .. code-block:: python
       from meshmode.array_context import pytest_generate_tests_for_pyopencl \
            as pytest_generate_tests
    in your pytest test scripts allows you to use the arguments ctx_factory,
    device, or platform in your test functions, and they will automatically be
    run for each OpenCL device/platform in the system, as appropriate.
    It also allows you to specify the ``PYOPENCL_TEST`` environment variable
    for device selection.
    """

    import pyopencl as cl
    from pyopencl.tools import _ContextFactory

    class ArrayContextFactory(_ContextFactory):
        def __call__(self):
            ctx = super().__call__()
            return PyOpenCLArrayContext(cl.CommandQueue(ctx))

        def __str__(self):
            return ("<array context factory for <pyopencl.Device '%s' on '%s'>" %
                    (self.device.name.strip(),
                     self.device.platform.name.strip()))

    import pyopencl.tools as cl_tools
    arg_names = cl_tools.get_pyopencl_fixture_arg_names(
            metafunc, extra_arg_names=["actx_factory"])

    if not arg_names:
        return

    arg_values, ids = cl_tools.get_pyopencl_fixture_arg_values()
    if "actx_factory" in arg_names:
        if "ctx_factory" in arg_names or "ctx_getter" in arg_names:
            raise RuntimeError("Cannot use both an 'actx_factory' and a "
                    "'ctx_factory' / 'ctx_getter' as arguments.")

        for arg_dict in arg_values:
            arg_dict["actx_factory"] = ArrayContextFactory(arg_dict["device"])

    arg_values = [
            tuple(arg_dict[name] for name in arg_names)
            for arg_dict in arg_values
            ]

    metafunc.parametrize(arg_names, arg_values, ids=ids)
```

inducer:
We could potentially parametrize that, like so:
```py
from functools import partial

def pytest_generate_tests_for_cl_array_context(context_from_cl_device, metafunc):

pytest_generate_tests_for_pyopencl_array_context = partial(pytest_generate_tests_for_cl_array_context, ArrayContextFactory)
```
Also, there's no need to copy-paste whole chunks of code. Github lets you link to code chunks at specific versions pretty easily. (I'll show you how next meeting.)

nchristensen:
> ```python
> pytest_generate_tests_for_pyopencl_array_context = partial(pytest_generate_tests_for_cl_array_context, ArrayContextFactory)
> ```

`partial` seems to interfere with pytest.

```
 from meshmode.array_context import (  # noqa
          pytest_generate_tests_for_pyopencl_array_context
          as pytest_generate_tests)
  from meshmode.array_context import PyOpenCLArrayContext
  from functools import partial
  pytest_generate_tests = partial(pytest_generate_tests, PyOpenCLArrayContext)
```

fails with

```
E   TypeError: pytest_generate_tests_for_pyopencl_array_context() missing 1 required positional argument: 'metafunc'
```

I see Metafunc.config gives access to the configuration values. Could the context be passed through there? https://docs.pytest.org/en/stable/reference.html?highlight=config#pytest.python.Metafunc.config

nchristensen:
@inducer Could you approve the workflows here?

meshmode/discretization/__init__.py
```diff
@@ -601,10 +606,30 @@ def resample_mesh_nodes(grp, iaxis):
...
+                    offset=lp.auto),  # In default data layout apparently
+                lp.GlobalArg("arg0", fp_format, shape=(ni, nj), offset=lp.auto,
+                    tags=[IsOpArray()]),
+                lp.GlobalArg("out",  fp_format, shape=(ne, ni), offset=lp.auto,
+                    tags=[IsDOFArray()], is_output=True),
+                lp.ValueArg("Ni", tags=[ParameterValue(ni)]),
+                lp.ValueArg("Nj", tags=[ParameterValue(nj)]),
+                lp.ValueArg("Ne", tags=[ParameterValue(ne)]),
+                ...
+            ]
```
inducer, line 625: Much of this information is prescriptive rather than descriptive, i.e. it prescribes a certain implementation (I.e. the metadata here is not interesting if the implementation isn't using loopy and only interesting if the einsum is indeed a single kernel.) What aspects of this are actually important to your transformation? Could you live with only the descriptive bits? (`IsOpArray`, `IsDOFArray`?)

nchristensen: Currently the eager autotuner expects to know the data type and `Ne` to determine how much of the input DOF arrays can be prefetched into local memory, `Ni` to decide how to split the  `i` loop and to set the local work dimensions, and `Nj` to decide how to unroll the `j` axis. The timing code also needs the dimensions and data types to create input arrays for the timing runs.

nchristensen: Since the arrays are arguments to  actx.einsum, I could handle most of the logic there. So yes, possibly only the IsOpArray and IsDOFArrays tags need to be applied here.

nchristensen: Restructured this so now only the tags are passed.

PR #77, opened 2020-11-11
Stop subclassing DOFArray from numpy.ndarray

inducer:
Having `DOFArray` be a subclass of `numpy.ndarray` had the potential to be confusing. For example, @MTCam ran into a situation where `DOFArray + np.ndarry([0.1, 0.2])` produced a `DOFArray` that (technically) represented DOFs for a discretization with two groups, because of unintended broadcasting. This change catches that problem (and, likely, other confusing behaviors) at the source and raises an error right away.

To merge:

- [x] Fix remaining `obj_array_vectorizes` in `discretization/visualization.py`
- [x] Test that `DOFArray` arithmetic behaves correctly
- [x] Make a grudge PR for compatibility with this
    - https://github.com/inducer/grudge/pull/29
- [x] Make a mirgecom PR for compatibility with this (shouldn't be big, @MTCam said he would take a look)
    - https://github.com/illinois-ceesd/mirgecom/pull/149
- [x] Make a pytential PR for compatibility with this
    - https://github.com/inducer/pytential/pull/36

cc @MTCam @kaushikcfd @alexfikl @majosm @w-hagen @lukeolson

### DIFF
--- a/meshmode/dof_array.py
+++ b/meshmode/dof_array.py
@@ -242,7 +242,7 @@ def obj_or_dof_array_vectorize_n_args(f, *args):
             if isinstance(arg, DOFArray)]
 
     if not dofarray_arg_indices:
-        from pytools import obj_array_vectorize_n_args
+        from pytools.obj_array import obj_array_vectorize_n_args
         return obj_array_vectorize_n_args(f, *args)
 
     leading_da_index = dofarray_arg_indices[0]
--- a/meshmode/dof_array.py
+++ b/meshmode/dof_array.py
@@ -199,6 +199,9 @@ class DOFArray:
     def __and__(self, arg): return self._bop(operator.and_, self, arg)  # noqa: E704
     def __xor__(self, arg): return self._bop(operator.xor, self, arg)  # noqa: E704
     def __or__(self, arg): return self._bop(operator.or_, self, arg)  # noqa: E704
+    def __rand__(self, arg): return self._bop(operator.and_, arg, self)  # noqa: E704
+    def __rxor__(self, arg): return self._bop(operator.xor, arg, self)  # noqa: E704
+    def __ror__(self, arg): return self._bop(operator.or_, arg, self)  # noqa: E704
 
     # bit shifts unimplemented for now
 
--- a/meshmode/dof_array.py
+++ b/meshmode/dof_array.py
@@ -22,17 +22,26 @@ THE SOFTWARE.
 
 import operator
 import numpy as np
-from typing import Optional, Iterable, Any
+from typing import Optional, Iterable, Any, Tuple, Union
 from functools import partial
+from numbers import Number
+import operator as op
+import decorator
 
 from pytools import single_valued, memoize_in
-from pytools.obj_array import obj_array_vectorize, obj_array_vectorize_n_args
+from pytools.obj_array import obj_array_vectorize
 
 from meshmode.array_context import ArrayContext, make_loopy_program
 
 
 __doc__ = """
 .. autoclass:: DOFArray
+
+.. autofunction:: obj_or_dof_array_vectorize
+.. autofunction:: obj_or_dof_array_vectorized
+.. autofunction:: obj_or_dof_array_vectorize_n_args
+.. autofunction:: obj_or_dof_array_vectorized_n_args
+
 .. autofunction:: thaw
 .. autofunction:: freeze
 .. autofunction:: flatten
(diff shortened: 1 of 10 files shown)


meshmode/dof_array.py
```diff
@@ -114,96 +118,180 @@ def from_list(cls, actx: Optional[ArrayContext], res_list) -> "DOFArray":
...
+
+    def _bop(self, f, op1, op2):
+        """Broadcasting logic for a generic binary operator."""
+        if isinstance(op1, DOFArray) and isinstance(op2, DOFArray):
+            if len(op1._data) != len(op2._data):
+                raise ValueError("DOFArray objects in binary operator must have "
+                        f"same length, got {len(op1._data)} and {len(op2._data)}")
+            return self._like_me([
+                f(op1_i, op2_i)
+                for op1_i, op2_i in zip(op1._data, op2._data)])
```
alexfikl, line 159:
```suggestion
            return self._like_me(f(op1_i, op2_i) for op1_i, op2_i in zip(op1._data, op2._data))
```
Does this work? It gets cast to a tuple anyway.

meshmode/dof_array.py
```diff
@@ -114,96 +118,180 @@ def from_list(cls, actx: Optional[ArrayContext], res_list) -> "DOFArray":
...
+    # bit shifts unimplemented for now
 
-    def __ge__(self, other):
-        return self._comparison(operator.ge, other)
+# }}}
 
-    # }}}
 
-# }}}
+def obj_or_dof_array_vectorize(f, ary):
```
alexfikl, line 211: These are great! The `obj_or_dof` part seems a bit clumsy though.

Would it be too mean to keep the `obj_array_vectorize` name? With the understanding that they're imported from the `dof_array` module, so they work on those too.

meshmode/dof_array.py
```diff
@@ -114,96 +118,180 @@ def from_list(cls, actx: Optional[ArrayContext], res_list) -> "DOFArray":
...
-    return DOFArray.from_list(None, [
-        ary.array_context.freeze(subary)
-        for subary in ary
-        ])
+    return DOFArray(None, tuple(
+        ary.array_context.freeze(subary) for subary in ary))
 
 
-def flatten(ary: np.ndarray) -> Any:
+def flatten(ary: Union[np.ndarray]) -> Any:
```
alexfikl, line 297:
```suggestion
def flatten(ary: Union[DOFArray, np.ndarray]) -> Any:
```
?

meshmode/dof_array.py
```diff
@@ -276,7 +360,7 @@ def prg():
 
     group_starts = np.cumsum([0] + group_sizes)
 
-    return DOFArray.from_list(actx, [
+    return DOFArray(actx, tuple(
```
alexfikl, line 366: Can these just be cast to a tuple in the constructor? Which can then take any iterable.

alexfikl (commented):
Just gave it a look and left a couple of nitpicks.

What was the intended effect of doing `DOFArray + np.ndarry([0.1, 0.2])` when there was only one group in the `DOFArray`?

inducer:
> What was the intended effect of doing DOFArray + np.ndarry([0.1, 0.2]) when there was only one group in the DOFArray?

It was a mishap, but arguably one that could/should have been caught. (Backstory: In the friendliest possible manner, I got pummeled on Tuesday along the lines of "object array of DOF array of CL array yougottobekiddingme" in a CEESD meeting)

meshmode/dof_array.py
inducer: My thought was that list comprehensions are probably minimally faster than generator expression. But I have exactly zero data to back that up.

meshmode/dof_array.py
inducer: > Would it be too mean to keep the `obj_array_vectorize` name? With the understanding that they're imported from the `dof_array` module, so they work on those too.

Not a fan. I like my explicit names.

meshmode/dof_array.py
alexfikl: I was thinking the other way around! Something to `timeit` for peace of mind, I guess.

alexfikl:
> It was a mishap, but arguably one that could/should have been caught.

Fair enough. Doesn't seem like it's too disruptive to the outside world anyway, so looks good to me!

meshmode/dof_array.py
alexfikl: You're right! At least for [this contrived example](https://gist.github.com/alexfikl/1c9fef4bade9a22ed445b25cfeffadc3).

TIL

inducer: Thanks for checking!

meshmode/dof_array.py
inducer: Since constructors can't be circumvented for object creation, I've gravitated towards making their arguments as close to the object representation as possible. If more convenience is needed, it's always possible to write wrappers.

PR #78, opened 2020-11-12
Vectorize connections

alexfikl:


meshmode/discretization/connection/chained.py
```diff
@@ -61,11 +62,12 @@ def __init__(self, connections, from_discr=None):
 
         self.connections = connections
 
-    def __call__(self, vec):
+    @obj_array_vectorized_n_args
```
inducer, line 65:
```suggestion
    @obj_array_vectorized

File: meshmode/discretization/connection/direct.py [hunk 2]
```diff
@@ -249,11 +250,15 @@ def full_resample_matrix(self, actx):
 
         return make_direct_full_resample_matrix(actx, self)
 
+    @obj_array_vectorized_n_args
```
inducer, line 253:
```suggestion
    @obj_array_vectorized
```

meshmode/discretization/connection/direct.py
```diff
@@ -249,11 +250,15 @@ def full_resample_matrix(self, actx):
...
     def __call__(self, ary):
         from meshmode.dof_array import DOFArray
         if not isinstance(ary, DOFArray):
             raise TypeError("non-array passed to discretization connection")
 
+        if ary.shape != (len(self.from_discr.groups),):
+            raise ValueError("array must have same size as discretization groups")
```
inducer, line 260: Rephrase

meshmode/discretization/connection/projection.py
```diff
@@ -115,11 +116,15 @@ def det(v):
...
-    def __call__(self, vec):
-        if not isinstance(vec, DOFArray):
+    @obj_array_vectorized_n_args
+    def __call__(self, ary):
+        if not isinstance(ary, DOFArray):
             raise TypeError("non-array passed to discretization connection")
 
-        actx = vec.array_context
+        if ary.shape != (len(self.from_discr.groups),):
+            raise ValueError("array must have same size as discretization groups")
```
inducer, line 125: Rephrase

meshmode/discretization/connection/projection.py
```diff
@@ -115,11 +116,15 @@ def det(v):
 
         return weights
 
-    def __call__(self, vec):
-        if not isinstance(vec, DOFArray):
+    @obj_array_vectorized_n_args
```
inducer, line 119:
```suggestion
    @obj_array_vectorized
```

inducer (commented):
Thanks!

meshmode/discretization/connection/chained.py
alexfikl: That didn't seem to work. I guess because it passed `self` in too?

inducer: Ah, fair. Nvm al these then.

meshmode/discretization/connection/direct.py
alexfikl: Hm.. "array size does not match incomming discretization"? "expected array of size %d, but got %d"?

inducer:
Either is fine.

meshmode/discretization/connection/direct.py
alexfikl: Should be fixed in f4a0385.

Actually realized that the `DirectDiscretizationConnection` already had this check lower down, so just moved it up before the allocations.

inducer:
Thx!

PR #79, opened 2020-11-12
Specify matplotlib backend in Github CI

inducer:


alexfikl:
How come this just managed to fail now?

inducer:
Good question. I imagine something changed on the conda end.

Issue #33, opened 2020-08-12
Check consistency during FacialAdjacencyGroup construction


inducer:
- boundary groups may only contain boundary tags (all negative `neighbors`)
- non-boundary groups may not contain boundary tags (all positive `neighbors`)

PR #83, opened 2020-11-18
Keep boundary tags when mapping meshes

alexfikl:


meshmode/mesh/processing.py
```diff
@@ -1040,6 +1040,7 @@ def map_mesh(mesh, f):  # noqa
 
     from meshmode.mesh import Mesh
     return Mesh(vertices, new_groups, skip_tests=True,
+            boundary_tags=mesh.boundary_tags,
             nodal_adjacency=mesh.nodal_adjacency_init_arg(),
             facial_adjacency_groups=mesh._facial_adjacency_groups,
             is_conforming=mesh.is_conforming)
```
alexfikl, line 1046:
```suggestion
    return mesh.copy(vertices=vertices, groups=new_groups)
```
?

inducer (approved):
Thx!

inducer:
(Just waiting for CI to pass.)

meshmode/mesh/processing.py
alexfikl: Actually, [get_copy_kwargs](https://github.com/inducer/meshmode/blob/2c33c8dd6c4e9b9a3f4882b9455751704302f9be/meshmode/mesh/__init__.py#L862) does not seem to propagate the `is_conforming` flag, if I'm reading it right?

alexfikl: Should be fixed in b1bb6cb

inducer:
@alexfikl Do my changes look OK to you?

alexfikl:
> @alexfikl Do my changes look OK to you?

Yeah, looks good! 

One unrelated worry is that passing both `vertices` and keeping a previous `facial_adjacency_groups` can make it inconsistent, but that should be fixed by #33?

inducer:
Thx!

inducer:
> One unrelated worry is that passing both vertices and keeping a previous facial_adjacency_groups can make it inconsistent, but that should be fixed by #33?

Yeah, there are a few opportunities for inconsistency, but IMO that's the case no matter what. I agree that #33 will help catch some of them.

PR #85, opened 2020-11-18
Fake numpy linalg norm

inducer:


meshmode/array_context.py
```diff
@@ -67,8 +127,11 @@ def loopy_implemented_elwise_func(*args):
                     **{"inp%d" % i: arg for i, arg in enumerate(args)})
             return result
 
-        from meshmode.dof_array import obj_or_dof_array_vectorized_n_args
-        return obj_or_dof_array_vectorized_n_args(loopy_implemented_elwise_func)
+        if name in self._numpy_math_functions:
+            from meshmode.dof_array import obj_or_dof_array_vectorized_n_args
+            return obj_or_dof_array_vectorized_n_args(loopy_implemented_elwise_func)
```
kaushikcfd, line 134: Do we need `loopy_implemented_elwise_func`, I think the functions supported by loopy are a subset of what's already in `_numpy_math_functions`. (Looks like a code path that's never going to be taken)

inducer: Not sure I understand your comment. `_numpy_math_functions` doesn't represent any implementation. Its purpose is just to limit which functions this will try and shove into loopy.

meshmode/array_context.py
```diff
@@ -67,8 +129,11 @@ def loopy_implemented_elwise_func(*args):
                     **{"inp%d" % i: arg for i, arg in enumerate(args)})
             return result
 
-        from meshmode.dof_array import obj_or_dof_array_vectorized_n_args
-        return obj_or_dof_array_vectorized_n_args(loopy_implemented_elwise_func)
+        if name in self._numpy_math_functions:
```
inducer, line 132:
```suggestion
        # limit which functions we try to hand off to loopy
        if name in self._numpy_math_functions:
```

meshmode/array_context.py
kaushikcfd: Oops, made a mistake. Though these were enumerated so `pyopencl`'s implementations could be called. Resolving.

meshmode/array_context.py
```diff
@@ -225,21 +296,84 @@ def thaw(self, array):
...
 
 class _PyOpenCLFakeNumpyNamespace(_BaseFakeNumpyNamespace):
+    def _get_fake_numpy_linalg_namespace(self):
+        return _PyOpenCLFakeNumpyLinalgNamespace(self._array_context)
+
     def maximum(self, x, y):
         import pyopencl.array as cl_array
         from meshmode.dof_array import obj_or_dof_array_vectorize_n_args
-        return obj_or_dof_array_vectorize_n_args(cl_array.maximum, x, y)
+        return obj_or_dof_array_vectorize_n_args(
```
alexfikl, line 305: Are these ever passed an object array?

alexfikl: At least given my usage, I would expect this to be more like
```
@obj_array_vectorize_n_args
def maximum(...):
     return dof_array_vectorize_n_args(...)
```

inducer: Could happen. Object array of `DOFArray` is not super uncommon. E.g. in the Lax-Friedrichs Flux in mirgecom.

alexfikl: (Obviously unrelated to the current PR, so feel free to ignore)

inducer: I'll go ahead and merge, but I'm happy to continue this discussion.

alexfikl: Unless I'm reading this wrong, it currently works for object arrays of `cl.Array`s and `DOFArray`s. No?

That's why I was thinking that it made more sense to make it work for object arrays of `DOFArray`s and `DOFArray`s.

alexfikl: `maximum` might not be the best example for this though, since it would broadcast weirdly, but maybe the Loopy `elwise_func` stuff.

inducer: You're right. Let's continue the discussion in #86.

PR #86, opened 2020-11-19
Recurse on obj arrays

inducer:
cc @alexfikl

### DIFF
--- a/meshmode/dof_array.py
+++ b/meshmode/dof_array.py
@@ -44,6 +44,7 @@ __doc__ = """
 
 .. autofunction:: thaw
 .. autofunction:: freeze
+
 .. autofunction:: flatten
 .. autofunction:: unflatten
 """
@@ -124,8 +125,6 @@ class DOFArray:
             raise TypeError("actx must be of type ArrayContext")
 
         return cls(actx, tuple(res_list))
-
-    # }}}
 
     def __str__(self):
         return str(self._data)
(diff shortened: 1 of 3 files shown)


meshmode/dof_array.py
```diff
@@ -213,16 +212,18 @@ def __ror__(self, arg): return self._bop(operator.or_, arg, self)  # noqa: E704
...
-    for :class:`DOFArray`\ s.
+    Works like :func:`~pytools.obj_array.obj_array_vectorize`, but recurses
+    on object arrays and also tolerates one final 'layer' of :class:`DOFArray`\ s.
     """
 
     if isinstance(ary, DOFArray):
         return ary._like_me([f(ary_i) for ary_i in ary._data])
     else:
-        return obj_array_vectorize(f, ary)
+        return obj_array_vectorize(partial(obj_or_dof_array_vectorize, f), ary)
```
alexfikl, line 226: I think the worry I had here before was that the caller needed to know if `ary` is a `DOFArray`, object array of `DOFArrays` or object array of something else, since `f` would need to handle those inputs.

This looks better from that point of view, since `f` always gets the group data from the `DOFArray`.

alexfikl (approved):
This is great! 

I have a lot of `make_obj_array([actx.np.log(...)])` and others that I can just unwrap now!

meshmode/dof_array.py
inducer: Thanks for following up! I'm less worried now. ;)

inducer:
Turns out: not entirely trouble-free: https://gitlab.tiker.net/inducer/meshmode/-/issues/14

cc @xywei

inducer:
#87

PR #90, opened 2020-11-20
Broadcast DOFArray across numpy arrays in binary operations

majosm:
Adds broadcasting operations between `DOFArray` and numpy arrays. Removes the need to use `make_obj_array([...])` for scalars, e.g.:

```diff
- np.outer(mom, mom)/make_obj_array([cv.mass]) + np.eye(dim)*make_obj_array([p])
+ np.outer(mom, mom)/cv.mass + np.eye(dim)*p
```

### DIFF
--- a/meshmode/dof_array.py
+++ b/meshmode/dof_array.py
@@ -154,6 +154,12 @@ class DOFArray:
 
     def _bop(self, f, op1, op2):
         """Broadcasting logic for a generic binary operator."""
+        if isinstance(op1, np.ndarray):
+            return obj_array_vectorize(lambda op: self._bop(f, op, op2),
+                op1.astype(np.object, copy=False))
+        if isinstance(op2, np.ndarray):
+            return obj_array_vectorize(lambda op: self._bop(f, op1, op),
+                op2.astype(np.object, copy=False))
         if isinstance(op1, DOFArray) and isinstance(op2, DOFArray):
             if len(op1._data) != len(op2._data):
                 raise ValueError("DOFArray objects in binary operator must have "


majosm:
Seems to work with grudge (inducer/grudge#34) and mirgecom (illinois-ceesd/mirgecom#154) (though I did have to fix a couple of tests in mirgecom that were using PyOpenCL arrays directly instead of `DOFArray`s).

@inducer What do you think? Any complaints about the implementation? Anything else we should do to evaluate this?

meshmode/dof_array.py
```diff
@@ -154,6 +154,18 @@ def _like_me(self, data):
 
     def _bop(self, f, op1, op2):
         """Broadcasting logic for a generic binary operator."""
+        if isinstance(op1, np.ndarray):
+            if op1.dtype == np.object:
+                return obj_array_vectorize(lambda op: self._bop(f, op, op2), op1)
+            else:
+                return obj_array_vectorize(lambda op: self._bop(f, op, op2),
+                    np.array(op1, dtype=np.object))
```
inducer, line 162:
```suggestion
                    # make obj_array_vectorize do the right thing
                    op1.astype(np.object))                    
```

meshmode/dof_array.py
```diff
@@ -154,6 +154,18 @@ def _like_me(self, data):
...
+        if isinstance(op2, np.ndarray):
+            if op2.dtype == np.object:
+                return obj_array_vectorize(lambda op: self._bop(f, op1, op), op2)
+            else:
+                return obj_array_vectorize(lambda op: self._bop(f, op1, op),
+                    np.array(op2, dtype=np.object))
```
inducer, line 168:
```suggestion
                    # make obj_array_vectorize do the right thing
                    op2.astype(np.object))
```

meshmode/dof_array.py
inducer, line 162: Perhaps change the logic to avoid repeating code between the two branches.

meshmode/dof_array.py
inducer, line 168: Perhaps change the logic to avoid repeating code between the two branches.

inducer (commented):
Thanks for thinking of this (and implementing it)! Aside from a few minor suggestions for code beautification, this looks great to me. Also, could you also do a test PR with https://github.com/inducer/pytential/? Good to merge after that.

majosm:
Everything looks good with pytential: inducer/pytential#42

inducer:
LGTM. Thanks for seeing this through!

PR #89, opened 2020-11-20
Array context: deprecate acos-style, nudge towards numpy arccos-style, fix loopy compilation of arc* functions

inducer:

PR #92, opened 2020-11-20
Fix, test array context reductions

### LINKED ISSUE #91: Reductions in `ArrayContext.np` behave differently from `numpy`
When the reduction result is a scalar,
- reductions in `numpy`, e.g., `numpy.sum`, `numpy.max`, return a scalar value, while
- reductions in `ArrayContext.np` return an array of size one.

This difference affects, for example, [gmres residual reporting in pytential](https://github.com/inducer/pytential/blob/10c2ea345e69367cfbba67ba5aa2954ebac467cf/pytential/solve.py#L252), where the formatting `.8e` will cause an error if `norm` is an array.

inducer:
Closes #91.

### DIFF
--- a/meshmode/array_context.py
+++ b/meshmode/array_context.py
@@ -325,15 +325,16 @@ class _PyOpenCLFakeNumpyNamespace(_BaseFakeNumpyNamespace):
 
     def sum(self, a, dtype=None):
         import pyopencl.array as cl_array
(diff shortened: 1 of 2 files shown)


inducer:
This will fail just like #89, for the same reason (pytential, https://github.com/inducer/pytential/pull/41). I'll merge it anyway, unless other failures crop up.

PR #94, opened 2020-11-26
add pytest_generate_tests_for_pyopencl_array_context to doc

matthiasdiener:
cc https://github.com/illinois-ceesd/mirgecom/pull/159#discussion_r530575775

### DIFF
--- a/meshmode/array_context.py
+++ b/meshmode/array_context.py
@@ -547,7 +547,7 @@ class PyOpenCLArrayContext(ArrayContext):
 # {{{ pytest integration
 
 def pytest_generate_tests_for_pyopencl_array_context(metafunc):
-    """Parametrize tests for :mod:`pytest` to use a :mod:`pyopencl` array context."""
+    """Parametrize tests for pytest to use a :mod:`pyopencl` array context."""
 
     import pyopencl as cl
     from pyopencl.tools import _ContextFactory
--- a/meshmode/array_context.py
+++ b/meshmode/array_context.py
@@ -550,8 +550,20 @@ def pytest_generate_tests_for_pyopencl_array_context(metafunc):
     """Parametrize tests for pytest to use a :mod:`pyopencl` array context.
 
     Performs device enumeration analogously to
-    :func:`pyopencl.tools.pytest_generate_tests_for_pyopencl`,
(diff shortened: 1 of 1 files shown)


inducer:
Problems with this:

* no docstring to explain what it's for
* why would this appear *before* the array context in the docs?

matthiasdiener:
> Problems with this:
> 
> * no docstring to explain what it's for
> * why would this appear _before_ the array context in the docs?

Should be fixed now.

meshmode/array_context.py
```diff
@@ -546,6 +547,8 @@ def transform_loopy_program(self, program):
 # {{{ pytest integration
 
 def pytest_generate_tests_for_pyopencl_array_context(metafunc):
+    """Parametrize tests for pytest to use a :mod:`pyopencl` array context."""
```
inducer, line 550:
```suggestion
    """Parametrize tests for pytest to use a :mod:`pyopencl` array context.
    
    Performs device enumeration analogously to :func:`pyopencl.tools.pytest_generate_tests_for_pyopencl`.
```

inducer (approved):

meshmode/array_context.py
inducer: Maybe also copy some usage instructions from [there](https://documen.tician.de/pyopencl/tools.html#pyopencl.tools.pytest_generate_tests_for_pyopencl).

inducer (approved):

meshmode/array_context.py
matthiasdiener: Added a bit more information there, is that what you had in mind?

inducer: That helps. The other aspect I had in mind is how to use this in test files. But I'm also OK with this as it is. Up to you whether to add it.

matthiasdiener: Ok, I expanded the description. (as a side note, I can't merge this PR since I don't have write permissions here).

inducer: > (as a side note, I can't merge this PR since I don't have write permissions here).

:facepalm: You guessed correctly. I had spaced on that. :) Thanks for your patience. Good to go once CI passes.

PR #98, opened 2020-12-31
Xdmf Visualizer

alexfikl:
Uses inducer/pyvisfile#12 to write out some XDMF + HDF5 files (requires h5py of course). It reuses the linear VTK connectivity but I had to write one XDMF Grid per group because I couldn't convince Paraview to index into arrays properly.

One nice benefit of these things is that HDF5 seems to be quite a bit better at compression. From running `test_meshmode.py::test_visualizers`:
```
>> ls -lSh
2.0M Dec 30 20:27 visualizer_vtk_simplex_3d_linear.vtu
846K Dec 30 20:27 visualizer_vtk_simplex_3d_lagrange.vtu
286K Dec 30 20:27 visualizer_xdmf_simplex_3d.h5
1.3K Dec 30 20:27 visualizer_xdmf_simplex_3d.xmf
```

### DIFF
--- a/.test-conda-env-py3.yml
+++ b/.test-conda-env-py3.yml
@@ -23,6 +23,9 @@ dependencies:
 # for pymetis
 - pybind11
 
+# for xdmf/hdf5 visualizer
+- h5py
+
 # Only needed to make pylint succeed
 - matplotlib
 
--- a/meshmode/discretization/visualization.py
+++ b/meshmode/discretization/visualization.py
@@ -812,9 +812,12 @@ class Visualizer:
 
         # {{{ write hdf5 + create xml tree
 
+        # NOTE: XDMF and/or XDMF in Paraview/VTK seems to be quite frail with
+        # a lot of setups found online just crashing it nonchalantly :(
+        #
         # The current setup writes a grid for each element group. The grids
         # are completely separate, i.e. each one gets its own subset of the
-        # nodes / connectivity / fields
+        # nodes / connectivity / fields. This seems to work reasonably well.
         #
         # This mostly works with the Xdmf3ReaderS (S for spatial) Paraview
         # plugin. It seems to also work with the XMDFReader (for Xdmf2) plugin,
@@ -850,17 +853,21 @@ class Visualizer:
             h5grid = h5.create_group("Grid")
 
             grids = []
+            node_nr_base = 0
             for igrp, (vgrp, gnodes) in enumerate(zip(connectivity.groups, nodes)):
                 grp_name = f"Group_{igrp:05d}"
                 h5grp = h5grid.create_group(grp_name)
 
+                # offset connectivity back to local numbering
+                visconn = vgrp.vis_connectivity.reshape(vgrp.nsubelements, -1) \
+                        - node_nr_base
+                node_nr_base += self.vis_discr.groups[igrp].ndofs
+
                 # hdf5 side
-                dset = h5grp.create_dataset("Nodes",
-                        data=gnodes.T)
+                dset = h5grp.create_dataset("Nodes", data=gnodes.T)
                 gnodes = DataArray.from_dataset(dset)
 
-                dset = h5grp.create_dataset("Connectivity",
-                        data=vgrp.vis_connectivity.reshape(vgrp.nsubelements, -1))
+                dset = h5grp.create_dataset("Connectivity", data=visconn)
                 gconnectivity = DataArray.from_dataset(dset)
 
                 # xdmf side
--- a/meshmode/discretization/visualization.py
+++ b/meshmode/discretization/visualization.py
@@ -1,4 +1,7 @@
-__copyright__ = "Copyright (C) 2014 Andreas Kloeckner"
+__copyright__ = """
+Copyright (C) 2014 Andreas Kloeckner
+Copyright (C) 2020 Alexandru Fikl
+"""
 
 __license__ = """
 Permission is hereby granted, free of charge, to any person obtaining a copy
@@ -25,7 +28,7 @@ from functools import singledispatch
 import numpy as np
 
 from pytools import memoize_method, Record
-from meshmode.dof_array import flatten, thaw
+from meshmode.dof_array import DOFArray, flatten, thaw
 
 from modepy.shapes import Shape, Simplex, Hypercube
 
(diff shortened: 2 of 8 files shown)


.test-conda-env-py3.yml
```diff
@@ -23,6 +23,9 @@ dependencies:
 # for pymetis
 - pybind11
 
+# for xdmf/hdf5 visualizer
+- h5py
```
inducer, line 27: Declare an "extra" dependency in `setup.py`?

test/test_meshmode.py
```diff
@@ -192,21 +198,23 @@ def test_visualizers(actx_factory, dim, group_cls):
...
+    basename = f"visualizer_vtk_{eltype}_{dim}d"
+    vis.write_vtk_file(f"{basename}_linear.vtu", names_and_fields, overwrite=True)
 
     with pytest.raises(RuntimeError):
         vis.write_vtk_file(f"{basename}_lagrange.vtu",
-                [("f", f)], overwrite=True, use_high_order=True)
+                names_and_fields, overwrite=True, use_high_order=True)
+
+    basename = f"visualizer_xdmf_{eltype}_{dim}d"
+    vis.write_xdmf_file(f"{basename}.xmf", names_and_fields, overwrite=True)
```
inducer, line 217: Skip if `h5py` is not available?

meshmode/discretization/visualization.py
```diff
@@ -701,6 +744,171 @@ def write_vtk_file(self, file_name, names_and_fields,
...
+        # are completely separate, i.e. each one gets its own subset of the
+        # nodes / connectivity / fields. This seems to work reasonably well.
+        #
+        # This mostly works with the Xdmf3ReaderS (S for spatial) Paraview
+        # plugin. It seems to also work with the XMDFReader (for Xdmf2) plugin,
+        # but that's not very tested.
+        #
+        # Tried
+        # * writing the nodes globally and then using a 'Reference' DataItem,
+        #   but that crashed Paraview on load
```
inducer, line 834: Add date/name to your observation. (Possibly also Vtk/paraview version numbers)

meshmode/discretization/visualization.py
```diff
@@ -701,6 +744,171 @@ def write_vtk_file(self, file_name, names_and_fields,
...
+                vtk.VTK_QUAD: TopologyType.Quadrilateral,
+                vtk.VTK_HEXAHEDRON: TopologyType.Hexahedron,
+                }[cell_type]
+
+    def write_xdmf_file(self, file_name, names_and_fields,
+            attrs=None, h5_file_options=None, dataset_options=None,
+            real_only=False, overwrite=False):
+        """Write an XDMF file (with an ``.xmf`` extension) containing the
+        arrays in *names_and_fields*. The heavy data is written to binary
+        HDF5 files with ``h5py``.
```
inducer, line 773: Add intersphinx link. Mention that it needs to be installed.

inducer, line 773: How does this interact with distributed-memory? If it's not yet implemented, say so.

inducer (commented):
Thanks! A few comments below.

meshmode/discretization/visualization.py
```diff
@@ -76,19 +80,75 @@ def separate_by_real_and_imag(names_and_fields, real_only):
...
+
+def resample_to_numpy(conn, vec, *, stack=False, by_group=False):
+    """
+    :arg stack: if *True* object arrays are stacked into a single
+        :class:`~numpy.ndarray`.
+    :arg by_group: if *True*, the per-group arrays in a :class:`DOFArray`
+        are flattened separately. This can be used to write each group as a
+        separate mesh (in supporting formats).
+    """
+    if isinstance(vec, np.ndarray) and vec.dtype.char == "O":
```
inducer, line 101:
```suggestion
    # "stack" exists as mainly as a workaround for Xdmf. See here:
    # https://github.com/inducer/pyvisfile/pull/12#discussion_r550959081
    # for (minimal) discussion.
    if isinstance(vec, np.ndarray) and vec.dtype.char == "O":
```

.test-conda-env-py3.yml
alexfikl: Added in 4f68c9b.

test/test_meshmode.py
alexfikl: Done in d92b88b.

meshmode/discretization/visualization.py
alexfikl: Done in 5bd70b8.

meshmode/discretization/visualization.py
alexfikl: Added in 3ca6889.

alexfikl: It shouldn't be too hard to add support for MPI. I'll look into it.

However, this will have a strong dependence on `mpi4py`, since that's what `h5py` uses internally.

alexfikl: I played around a bit with this and I'm not quite sure how to make it work nicely.

The current setup basically writes each `ElementGroup` as a separate XDMF `Grid`. In a distributed setting, there's even more groups for each local mesh, depending on how they get partitioned, but I'm not sure if it makes sense to write a grid per-group per-rank..

inducer: That's OK, not everything needs to be done on day 1. I think we should be able to handle that without too much interface upheaval (which is my main concern), so I think it'll be OK.

inducer: (And thanks for investigating!)

inducer (approved):
LGTM, and thanks! This'll go in once CI passes.

alexfikl:
> LGTM, and thanks! This'll go in once CI passes.

Sounds good! I've added a little comment for the parallel stuff in f049df8.

meshmode/discretization/visualization.py
```diff
@@ -701,6 +748,181 @@ def write_vtk_file(self, file_name, names_and_fields,
...
+            h5_file_options = {}
+
+        dataset_defaults = {"compression": "gzip", "compression_opts": 6}
+        if dataset_options is not None:
+            dataset_defaults.update(dataset_options)
+        dataset_options = dataset_defaults
+
+        if "comm" in h5_file_options \
+                or h5_file_options.get("driver", None) == "mpio":
+            raise ValueError("distributed memory visualization not supported")
```
inducer, line 802:
```suggestion
            raise NotImplementedError("distributed memory visualization")
```

meshmode/discretization/visualization.py
```diff
@@ -701,6 +748,181 @@ def write_vtk_file(self, file_name, names_and_fields,
...
+                vtk.VTK_HEXAHEDRON: TopologyType.Hexahedron,
+                }[cell_type]
+
+    def write_xdmf_file(self, file_name, names_and_fields,
+            attrs=None, h5_file_options=None, dataset_options=None,
+            real_only=False, overwrite=False):
+        """Write an XDMF file (with an ``.xmf`` extension) containing the
+        arrays in *names_and_fields*. The heavy data is written to binary
+        HDF5 files, which requires installing :ref:`h5py <h5py:install>`.
+        Distributed memory visualization is not supported.
```
inducer, line 778:
```suggestion
        Distributed memory visualization is not yet supported.
```

PR #99, opened 2021-01-08
Make DOFArrays picklable via global-state context manager

inducer:
Context: https://github.com/illinois-ceesd/mirgecom/pull/188/

### DIFF
--- a/meshmode/dof_array.py
+++ b/meshmode/dof_array.py
@@ -247,7 +247,7 @@ class DOFArray:
         if self.array_context is not actx:
             ary = thaw(actx, freeze(self))
 
-        return [ary_i.shape for ary_i in ary._data], actx.to_numpy(flatten(ary))
+        return [actx.to_numpy(ary_i) for ary_i in ary._data]
 
     def __setstate__(self, state):
         try:
@@ -259,10 +259,8 @@ class DOFArray:
             raise RuntimeError("DOFArray instances can only be unpickled while "
                     "array_context_for_pickling is active.")
 
-        group_shapes, np_array = state
-        res = _unflatten(actx, group_shapes, actx.from_numpy(np_array))
-        self._data = res._data
         self.array_context = actx
+        self._data = [actx.from_numpy(ary_i) for ary_i in state]
 
     # }}}
 
--- a/meshmode/dof_array.py
+++ b/meshmode/dof_array.py
@@ -27,6 +27,9 @@ from functools import partial
 from numbers import Number
 import operator as op
 import decorator
+import threading
+from contextlib import contextmanager
+
 
 from pytools import single_valued, memoize_in
 from pytools.obj_array import obj_array_vectorize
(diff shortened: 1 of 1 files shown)


inducer:
@alexfikl Do you think this a bad idea? It *does* introduce some global state.

alexfikl:
> @alexfikl Do you think this a bad idea? It _does_ introduce some global state.

Looking at the code in the linked PR, it looks pretty nice in practice.

You can probably remove the global state by moving the `from_numpy` and `to_numpy` outside of the `__state__` things, but I'm not sure it's worth it. It seems contained enough to me.

I was actually planning to use the XDMF/HDF5 stuff for checkpointing / restarting + visualization all at once to save on some space. This seems a lot more straightforward though.. Hm..

meshmode/dof_array.py
```diff
@@ -215,6 +230,42 @@ def __ror__(self, arg): return self._bop(operator.or_, arg, self)  # noqa: E704
...
+        if actx is None:
+            raise RuntimeError("DOFArray instances can only be pickled while "
+                    "array_context_for_pickling is active.")
+
+        ary = self
+
+        if self.array_context is not actx:
+            ary = thaw(actx, freeze(self))
+
+        return [ary_i.shape for ary_i in ary._data], actx.to_numpy(flatten(ary))
```
alexfikl, line 250: Why does this need to be flattened/unflattened?

inducer: Thanks for taking a look! Good catch, it doesn't. Saves a copy each way to not do that. 15ccf1e

inducer:
> I was actually planning to use the XDMF/HDF5 stuff for checkpointing / restarting + visualization all at once to save on some space. This seems a lot more straightforward though.. Hm..

I think that's desirable. The pickling stuff is mainly intended as a stopgap. It's simple and it works (hopefully), but there's no actually good reason to write out the same data twice. (Once for vis, once for restart.)

meshmode/dof_array.py
```diff
@@ -215,6 +230,40 @@ def __ror__(self, arg): return self._bop(operator.or_, arg, self)  # noqa: E704
...
+            actx = _ARRAY_CONTEXT_FOR_PICKLING_TLS.actx
+        except AttributeError:
+            actx = None
+
+        if actx is None:
+            raise RuntimeError("DOFArray instances can only be unpickled while "
+                    "array_context_for_pickling is active.")
+
+        self.array_context = actx
+        self._data = [actx.from_numpy(ary_i) for ary_i in state]
```
alexfikl, line 263:
```suggestion
        self._data = tuple([actx.from_numpy(ary_i) for ary_i in state])
```
?

inducer: Thx!

inducer:
> You can probably remove the global state by moving the `from_numpy` and `to_numpy` outside of the `__state__` things, but I'm not sure it's worth it. It seems contained enough to me.

I don't think that's possible: The closest representation of the array that can be pickled is numpy, and there isn't a way to get extra data into the pickle/unpickle methods other than by global state.

The benefit of enabling pickling is that you can then serialize (potentially) large/complex data structures containing `DOFArray`s without worrying about the details.

alexfikl:
> The benefit of enabling pickling is that you can then serialize (potentially) large/complex data structures containing `DOFArray`s without worrying about the details.

Oh yeah, I forgot about potentially nesting `DOFArray`s in something. Scratch that then! I can't think of any way around it either.

inducer:
OK, let's do this then. If it turns out to be a bad idea, we can always deprecate it.

PR #101, opened 2021-01-14
Add some in-place operators to DOFArray

alexfikl:
Added `iadd`, `isub`, `imul` and `itruediv`. There's some more in `cl.Array` that could be added, but I mostly just needed these.

### DIFF
--- a/meshmode/dof_array.py
+++ b/meshmode/dof_array.py
@@ -160,12 +160,58 @@ class DOFArray:
 
     # }}}
 
+    # {{{ ndarray interface
+
+    def _like_me(self, data):
+        return DOFArray(self.array_context, tuple(data))
+
     @property
     def shape(self):
         return (len(self),)
 
-    def _like_me(self, data):
-        return DOFArray(self.array_context, tuple(data))
+    @property
+    def size(self):
+        return len(self)
+
+    def copy(self):
+        return self._like_me([subary.copy() for subary in self])
+
+    def fill(self, value):
+        for subary in self:
+            subary.fill(value)
+
+    def conj(self):
+        return self._like_me([subary.conj() for subary in self])
+
+    conjugate = conj
+
+    @property
+    def real(self):
+        return self._like_me([subary.real for subary in self])
+
+    @property
+    def imag(self):
+        return self._like_me([subary.imag for subary in self])
+
+    # }}}
+
+    # {{{ arithmetic
+
+    def _ibop(self, f, arg):
+        if isinstance(arg, DOFArray):
+            if len(self) != len(arg):
+                raise ValueError("'DOFArray' objects in binary operator must "
+                        "have the same length: {len(self)} != {len(arg)}")
+
+            for i, subary in enumerate(self):
+                f(subary, arg[i])
+        elif isinstance(arg, Number):
+            for subary in self:
+                f(subary, arg)
+        else:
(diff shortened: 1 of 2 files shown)


meshmode/dof_array.py
```diff
@@ -160,12 +160,58 @@ def __iter__(self):
...
+    @property
+    def imag(self):
+        return self._like_me([subary.imag for subary in self])
+
+    # }}}
+
+    # {{{ arithmetic
+
+    def _ibop(self, f, arg):
+        if isinstance(arg, DOFArray):
```
inducer, line 201: Docstring?

meshmode/dof_array.py
```diff
@@ -208,11 +255,14 @@ def __pos__(self): return self  # noqa: E704
...
 
-    def conj(self): return self._like_me([self_i.conj() for self_i in self._data])  # noqa: E704, E501
-    @property
-    def real(self): return self._like_me([self_i.real for self_i in self._data])  # noqa: E704, E501
-    @property
-    def imag(self): return self._like_me([self_i.imag for self_i in self._data])  # noqa: E704, E501
+    def __iadd__(self, arg): return self._ibop(op.iadd, arg)            # noqa: E704
+    def __isub__(self, arg): return self._ibop(op.isub, arg)            # noqa: E704
+    def __imul__(self, arg): return self._ibop(op.imul, arg)            # noqa: E704
+    def __itruediv__(self, arg): return self._ibop(op.itruediv, arg)    # noqa: E704
```
inducer, line 261: Hey, at least complete the set.

inducer, line 261: Document them?

alexfikl: You mean with all the other stuff that's in [operator](https://docs.python.org/3/library/operator.html#in-place-operators)?

meshmode/dof_array.py
alexfikl: Added in 4a38de8.

meshmode/dof_array.py
alexfikl: Added some docs in 4a38de8. Let me know if they're ok!

inducer: Yeah.

alexfikl: Added them in d0bfbac, with the caveats
* `ipow` doesn't seem implemented in `cl.Array`
* `imod` (same as `mod`) is also missing
* didn't add the bitshift operators, since they weren't added for the out-of-place case either.

meshmode/dof_array.py
```diff
@@ -89,13 +89,35 @@ class DOFArray:
...
+    .. attribute:: imag
+
     This object supports arithmetic, comparisons, and logic operators.
 
     .. note::
 
         :class:`DOFArray` instances support elementwise ``<``, ``>``,
         ``<=``, ``>=``. (:mod:`numpy` object arrays containing arrays do not.)
 
+    Basic in-place operations are also supported.
```
inducer, line 111:
```suggestion
    Basic in-place operations are also supported. Note that not all arrays provided
    by :class:`meshmode.array_context.ArrayContext` implemenntations support
    in-place operations. Those based on lazy evaluation are a salient example.
```

inducer (approved):
Thx!

meshmode/dof_array.py
inducer: Fair enough, though the type isn't really specific to `cl.Array`.

PR #105, opened 2021-01-19
Test that connections that should be are actually permutations

inducer:
This adds a test that currently fails: Restriction to the boundary and opposite-face swap fall into the "interpolation matrix" code path [here](https://github.com/inducer/meshmode/blob/3af83a45f134014c1506bd46430cafc9641de9c9/meshmode/discretization/connection/direct.py#L325-L343) rather than the permutation matrix one, even though the matrices should be permutations. There's some crappy thresholding right before that to decide which is which. This is an efficiency problem, as operations that should merely be indirect memory access actually end up consuming considerable flops.

cc @lukeolson


test/test_connection.py
```diff
@@ -0,0 +1,103 @@
...
+            zip(conn.to_discr.groups, conn.groups)):
+        for i_batch, batch in enumerate(cgrp.batches):
+            if not len(batch.from_element_indices):
+                continue
+
+            point_pick_indices = conn._resample_point_pick_indices(
+                    actx, i_tgrp, i_batch)
+
+            assert point_pick_indices is not None
+
```
alexfikl, line 58:
```suggestion
            if point_pick_indices is None:
                return False
                
    return True
```

alexfikl:
I played around a bit with this locally and it can pass the tests if the [tolerance in here](https://github.com/inducer/meshmode/blob/3af83a45f134014c1506bd46430cafc9641de9c9/meshmode/discretization/connection/direct.py#L227) gets pushed up a bit.

Most of the fails seem to be from the opposite face connection, which also seems to have its [own tolerance](https://github.com/inducer/meshmode/blob/3af83a45f134014c1506bd46430cafc9641de9c9/meshmode/discretization/connection/opposite_face.py#L66) that's quite a bit larger. Matching them up a bit seems to make most of the tests pass.

test/test_connection.py
inducer: Good catch, thanks!

inducer:
Thanks for taking a look! Maybe be it'll be enough to twiddle tolerances to get this to go. I agree that the opposite-face-match tolerance and the permutation-matrix one should be roughly matched up, although they're really for different things. (nodes vs interpolation matrix entries) I'm a bit hesitant to dial down the Gauss-Newton tolerance in opposite-face matching without looking at the history of why it got dialed up that much in the first place. :)

alexfikl:
> I'm a bit hesitant to dial down the Gauss-Newton tolerance in opposite-face matching without looking at the history of why it got dialed up that much in the first place. :)

I tried to dial it down to `1.0e+3` and it started failing pretty fast at higher orders. Even `1.0e+4` started failing for orders 7-8. To be precise, there's an [assert in there](https://github.com/inducer/meshmode/blob/3af83a45f134014c1506bd46430cafc9641de9c9/meshmode/discretization/connection/opposite_face.py#L101) that started yelling, not the actual Newton iteration.

inducer:
Is it promising to try and build the interpolation matrix in quad precision? (`float128`) It's a one-time cost...

alexfikl:
> Is it promising to try and build the interpolation matrix in quad precision? (`float128`) It's a one-time cost...

I don't know if that will work, since `numpy.linalg` doesn't like `float128` from what I remember, so [modepy.resampling_matrix](https://github.com/inducer/modepy/blob/ed436e449826c723a29f210e8e1c27c54195c955/modepy/matrices.py#L145) will raise. Do you mean just casting it after the fact to the row sums?

inducer:
Thanks for checking! AFAIR, all we need is `linalg.solve`. [This crappy row echelon factorizer](https://github.com/illinois-scicomp/cs598apk-f19-binder/blob/master/dense_compute/m_echelon.py) might get us halfway there, we'd only need some triangular solves. (We know the matrices are invertible, so RREF=LU).

inducer:
I just spent a bit more time here, and I learned a few things.

*   First, when trying to find the permutation between unit nodes, there's kind of no point to constructing the interpolation matrix only to match it to the identity. We might as well compare the nodes directly. It's O(n^2) instead of O(n^3) to boot. That seems to  help some cases.
*   After that modification, the only cases still failing the assertion that the restriction to the boundary is a permutation are the higher-order warp-and-blend nodes. And the test might be correct: 2D facial restrictions of the 3D warp-and-blend nodes don't seem to be equivalent to 2D warp-and-blend nodes. (They are for all the other node types.) As a result, I think we need to introduce a separate element group type for 2D restrictions of 3D warp-and-blend.
*   Plenty of cases still fail the assertion that opposite-face-swap is a permutation. I picked 3D second-order equispaced as (what I thought was) the most egregious example to fail this test. What I found was that the Gauss-Newton objective function seems to have multiple minima, one at the "wrong answer", and one at the right answer. I added a debug snippet to the opposite-face finding (hence the flake8 failures) that proves the existence of those two minima:
    https://github.com/inducer/meshmode/blob/a3c689bc22ae66419d5b5305fe2b9ad7f5e0d106/meshmode/discretization/connection/opposite_face.py#L225-L264
    To see this, you need to change the `if 0` to `if 1` and run
    ```
    pycl test_connection.py 'test_bdry_restriction_is_permutation(_acf, PolynomialEquidistantSimplexGroupFactory, 3, 2)'
    ```
    which will print something like
    ```
    GOOD UNIT NODES
    [[-1. -1. -1.  0.  0.  1.]
     [-1.  0.  1. -1.  0. -1.]]
    BAD UNIT NODES
    [[-1.00000000e+00 -1.00000000e+00 -9.84191183e-01  1.65458682e-16
      -1.11094593e-15  1.00000000e+00]
     [-1.00000000e+00  5.28164206e-16  1.00000000e+00 -1.00000000e+00
      -2.07264860e-15 -1.00000000e+00]]

    FIXED RESIDUAL
    1.1102230246251565e-15
    UNFIXED RESIDUAL
    1.3322676295501878e-15
    ```
    Note the `-0.984` hanging out in a set of unit nodes that should only consist of -1s, 0s, and 1s. (Remember, equispaced, order 2!)

I'm out of steam for the night, but I hope this will help get us closer to tracking this down.

(cc @nchristensen since we had just discussed this in our meeting)

nchristensen:
Interesting. Thanks for looking!

inducer:
#225 is a version of this that's supposed to be mergeable right away, but with some tests still `xfail`'d.

inducer:
#225 and #226 should resolve the practical impact of this. There remains the mystery of why Gauss-Newton had two minima; I'd really like to know why. My plan is to update this after those two merge to leave this as an experimentation ground to investigate that question.

inducer:
Continued in #228.

PR #106, opened 2021-01-20
Make visualizer work with zero-length object arrays

majosm:
Currently, the visualizer code checks whether dataclass attributes are `None` or not, but it doesn't check for empty object arrays. This is needed by illinois-ceesd/mirgecom#196 for graceful handling of optional scalars. (Without the check, the "no scalars" case runs into an error inside `separate_by_real_and_imag` when it tries to determine the data type of `field[0]`.)


inducer:
LGTM so far(overlooked the 'draft' status).

majosm:
Now it's ready I think.

inducer:
Cool. In it goes. Thanks!

PR #97, opened 2020-12-25
Update refinement

alexfikl:
Follow up from #70 with some discussion in alexfikl/meshmode#3.

This PR mostly consolidates the refinement of elements backed by modepy shapes (from inducer/modepy#16). The actual algorithm is not changed (at least in spirit) from the existing one.

Main benefit is that it can now refine cubes, not just simplices!

PR #118, opened 2021-02-22
(WIP) Fix surface refinement

inducer:
This adds a failing test for the ability to refine toruses and spheres without large node-vertex inconsistencies.

I'm guessing this might have to do with the work in #97. I don't yet fully understand what's going on.

cc @alexfikl

### DIFF
--- a/test/test_refinement.py
+++ b/test/test_refinement.py
@@ -36,6 +36,7 @@ from meshmode.mesh.generation import (  # noqa: F401
         generate_icosahedron, generate_box_mesh, make_curve_mesh, ellipse)
 from meshmode.mesh.refinement.utils import check_nodal_adj_against_geometry
 from meshmode.mesh.refinement import Refiner, RefinerWithoutAdjacency
+import meshmode.mesh.generation as mgen
 
 from meshmode.mesh import SimplexElementGroup, TensorProductElementGroup
 from meshmode.discretization.poly_element import (
@@ -224,8 +225,7 @@ def test_refinement_connection(
             mesh = get_blob_mesh(mesh_par, mesh_order)
             h = float(mesh_par)
         elif mesh_name == "warp":
-            from meshmode.mesh.generation import generate_warped_rect_mesh
-            mesh = generate_warped_rect_mesh(dim, order=mesh_order, n=mesh_par,
+            mesh = mgen.generate_warped_rect_mesh(dim, order=mesh_order, n=mesh_par,
                     group_cls=group_cls)
             h = 1/mesh_par
         else:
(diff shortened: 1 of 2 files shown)


alexfikl:
What's supposed to fail? It works locally.

test/test_refinement.py
```diff
@@ -331,6 +330,30 @@ def test_conformity_of_uniform_mesh(refinement_rounds):
     assert is_boundary_tag_empty(mesh, BTAG_ALL)
 
 
+@pytest.parametrize("mesh_name", ["torus", "icosphere"])
```
alexfikl, line 333:
```suggestion
@pytest.mark.parametrize("mesh_name", ["torus", "icosphere"])
```

inducer:
I get
```
  File "/home/andreas/src/meshmode/test/test_refinement.py", line 360, in <module>
    exec(sys.argv[1])
  File "<string>", line 1, in <module>
  File "/home/andreas/src/meshmode/test/test_refinement.py", line 349, in test_refine_surfaces
    refined_mesh = refine_uniformly(mesh, 1)
  File "/home/andreas/src/meshmode/meshmode/mesh/refinement/__init__.py", line 991, in refine_uniformly
    refiner.refine_uniformly()
  File "/home/andreas/src/meshmode/meshmode/mesh/refinement/no_adjacency.py", line 61, in refine_uniformly
    return self.refine(flags)
  File "/home/andreas/src/meshmode/meshmode/mesh/refinement/no_adjacency.py", line 227, in refine
    new_mesh = Mesh(new_vertices, new_el_groups, is_conforming=(
  File "/home/andreas/src/meshmode/meshmode/mesh/__init__.py", line 779, in __init__
    assert _test_node_vertex_consistency(
  File "/home/andreas/src/meshmode/meshmode/mesh/__init__.py", line 983, in _test_node_vertex_consistency
    assert _test_node_vertex_consistency_resampling(mesh, mgrp, tol)
  File "/home/andreas/src/meshmode/meshmode/mesh/__init__.py", line 971, in _test_node_vertex_consistency_resampling
    assert max_el_vertex_error < tol*size, max_el_vertex_error
AssertionError: 9.901371953343942
```
Strangely, that seems to reproduce on my laptop and nowhere else.

alexfikl:
Hm, I've added a `refine_uniformly` to various surface tests in `test_meshmode.py` too and couldn't get any of them to fail (ran multiple times).

Does it fail for both the sphere and the torus?

inducer:
> Does it fail for both the sphere and the torus?

Yeah. The fact that CI passed is further indication that this might me something to do with my environment. I'll need to investigate. Thanks for helping!

inducer:
https://github.com/inducer/pytential/issues/63 is another thing that seems to be unique to my laptop. Kind of weird.

inducer:
Very strangely, this works OK on my laptop in an environment based on conda.

alexfikl:
> Very strangely, this works OK on my laptop in an environment based on conda.

I'm not using conda locally, just good old venv with master from pretty much everything.

inducer:
On my laptop: Rebuilt my venv with the existing source trees... same result. Freshly downloaded source with fresh venv... also same result.

On dunkel: Freshly downloaded source with fresh venv... no error.

PR #121, opened 2021-03-07
Do not copy results of reductions to host

kaushikcfd:


meshmode/array_context.py
```diff
@@ -339,16 +339,15 @@ def where_inner(inner_crit, inner_then, inner_else):
...
+        return cl_array.sum(a, dtype=dtype, queue=self._array_context.queue)
 
     def min(self, a):
         import pyopencl.array as cl_array
-        return cl_array.min(a, queue=self._array_context.queue).get()[()]
+        return cl_array.min(a, queue=self._array_context.queue)
 
     def max(self, a):
         import pyopencl.array as cl_array
-        return cl_array.max(a, queue=self._array_context.queue).get()[()]
```
inducer, line 351: This is a pretty big compatibility break. Want to elaborate why this is necessary?

kaushikcfd: 1. Forces the lazy code path to eagerly compute, as there is an implicit freeze here.
2. The implicit device->host transfers might become evident if in the future we also support `axis`.

kaushikcfd: This is breaking a lot of tests :astonished:. Maybe for the special case when the shape of the array to be returned is (), we can just return the value.

inducer: You know, I'm coming around to this. Copying reduction results *is* inconsistent, and forcing a lazy freeze is not desirable. I suspect we just need to bite the bullet and make this incompatible change. It's not great, but the current state is arguably just incorrect.

meshmode/array_context.py
```diff
@@ -339,16 +339,15 @@ def where_inner(inner_crit, inner_then, inner_else):
 
     def sum(self, a, dtype=None):
```
inducer, line 340: `la.norm` probably needs the same treatment.

meshmode/array_context.py
inducer, line 351: We *could* do a backward compatibility job here with a mode switch, but I don't think that's a good idea. Do you agree?

meshmode/array_context.py
kaushikcfd: la.norm in meshmode is realized as calls to actx.np.linalg.(sum|max). So the current implemenation should be fine.

meshmode/array_context.py
kaushikcfd: Difficult to make the claim that breaking back-compatibility is a good idea. But, it was broken when compared to other array ops provided by ArrayContext.

kaushikcfd:
@inducer: Another change that had to be done: actx.np.linalg.norm of DOFArray just returns a DOFArray of the thawed norms. I have un-deprecated `flat_norm` that still takes a DOFArray and returns the norm of the frozen norm of it's sub arrays. 

(The tests are failing, as this breaks back-compat. If there's no objection to this direction, I can open PRs to apply the necessary changes to the downstream projects.)

meshmode/array_context.py
inducer: OIC.

:+1:

inducer:
> actx.np.linalg.norm of DOFArray just returns a DOFArray of the thawed norms.

I don't understand. Is that the state you would like, or the current state, or...? It doesn't seem desirable, unless I'm missing something.

kaushikcfd:
Closing, as it was handled by https://github.com/inducer/arraycontext/pull/43.

Issue #123, opened 2021-03-09
Better support moving geometry


inducer:
Some problems that currently crop up:

- [x] Recreating meshes is awkward. @alexfikl proposes `Discretization.copy(mesh=...)` (or `.replace(mesh=...)`). (which?) That seems easy and useful, but it requires remembering the group factory (probably OK). 
- `Mesh`es can't be created efficiently. Their data structure involves `numpy` at the moment. Maybe meshes should start using array contexts, too?
- [ ] Group factories need to recompute discretization node arrays. If the mesh already uses the right unit nodes, this can probably be avoided. (#135 )

cc @alexfikl

PR #124, opened 2021-03-09
Add a copy method to Discretization

alexfikl:
Address the first (and simplest) point from #123.

I went with copy instead of replace because similar classes are based on `pytools.Record` and it uses copy.

### DIFF
--- a/meshmode/discretization/__init__.py
+++ b/meshmode/discretization/__init__.py
@@ -216,6 +216,15 @@ class Discretization:
                 }[self.real_dtype.type])
 
         self._setup_actx = actx
+        self._group_factory = group_factory
+
+    def copy(self, actx=None, mesh=None, group_factory=None, real_dtype=None):
+        return type(self)(
+                self._setup_actx if actx is None else actx,
+                self.mesh if mesh is None else mesh,
+                self._group_factory if group_factory is None else group_factory,
+                self.real_dtype if real_dtype is None else real_dtype,
+                )
 
     @property
     def dim(self):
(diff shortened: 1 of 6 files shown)


meshmode/discretization/connection/face.py
```diff
@@ -340,9 +340,10 @@ def make_face_restriction(actx, discr, group_factory, boundary_tag,
 
     bdry_mesh = Mesh(bdry_vertices, bdry_mesh_groups)
 
-    from meshmode.discretization import Discretization
-    bdry_discr = Discretization(
-            actx, bdry_mesh, group_factory)
+    bdry_discr = discr.copy(
+            actx=actx,
+            mesh=bdry_mesh,
+            group_factory=group_factory)
```
alexfikl, line 346: These changes (and other similar ones) are a bit unrelated, but I used it here instead of adding a proper test. Is that ok?

alexfikl: Actually this could be useful if someone subclasses `Discretization`, so not completely useless!

meshmode/discretization/__init__.py
```diff
@@ -216,6 +216,15 @@ def __init__(self, actx: ArrayContext, mesh, group_factory,
                 }[self.real_dtype.type])
 
         self._setup_actx = actx
+        self._group_factory = group_factory
+
+    def copy(self, actx=None, mesh=None, group_factory=None, real_dtype=None):
```
inducer, line 221: Docs! (Otherwise LGTM)

alexfikl: Added some docs in 0357de4.

inducer:
Thx!

PR #125, opened 2021-03-09
Test for mesh with interior nodes

alexfikl:
Included changes:

* Adds a test that uses interior quadrature nodes in the `Mesh`.
* Adds a `unit_nodes` argument to `generate_icosphere` and `generator_torus`.

It seems to be working as expected. The only hiccup is that it requires a lot more elements to actually pass `_test_node_vertex_consistency`.


inducer:
Thx!

PR #128, opened 2021-03-10
Slightly adjust test tolerance

thomasgibson:
Does exactly what it suggests :smile: 

This is based on some testing using the conda environment created via https://github.com/illinois-ceesd/emirge, and a conversation with @inducer.

It appears that using OpenBLAS from conda forge produces slightly different computed values in `test_meshmode.py`, causing test failures when using the conda environment for Mirgecom.


inducer:
Thx!

PR #131, opened 2021-03-11
Introduce nodal/modal element groups and connections

thomasgibson:
This PR revamps the element group class hierarchy and introduces two new connections, which allows for compact mapping to/from nodal and modal representations of a finite element function.


meshmode/discretization/__init__.py
```diff
@@ -1,4 +1,7 @@
-__copyright__ = "Copyright (C) 2013-2020 Andreas Kloeckner"
+__copyright__ = """
+Copyright (C) 2013-2021 Andreas Kloeckner
+Copyright (C) 2021 Thomas Gibson
```
inducer, line 3:
```suggestion
Copyright (C) 2021 University of Illinois Board of Trustees
```
https://otm.illinois.edu/disclose-protect/open-source-licensing

(Here and elsewhere)

meshmode/discretization/modal.py
```diff
@@ -0,0 +1,72 @@
...
+
+
+class ModalDiscretization(DiscretizationBase):
+    """An unstructured composite modal discretization."""
+
+    @memoize_method
+    def modes(self):
+        r"""
+        :returns: object array of shape ``(ambient_dim,)`` containing
+            :class:`~meshmode.dof_array.DOFArray`\ s of modal coefficients.
```
inducer, line 44: Modal coefficients of what? (Your getting modal coefficients of the coordinate field here, but why? and for what? Is that important enough to warrant a method?)

inducer (commented):
Thanks! A thing that would maybe make this feel less empty (and make it more testable) would be to make connections that go to and from modal (from various versions of nodal).

Basically, my thought is that nobody would create a stand-alone modal discretization. Instead, you would basically always need to have a nodal discretization along for the ride, e.g. to evaluate functions with.

inducer:
To cut down on noise in my email, I'm unsubscribing from this PR. When it next needs my attention, please @-mention me or hit the "request review" button. Otherwise, I may not see your messages in a timely manner.

meshmode/discretization/__init__.py
```diff
@@ -292,90 +359,9 @@ def empty_like(self, array: _DOFArray):
     def zeros_like(self, array: _DOFArray):
         return self.zeros(array.array_context, dtype=array.entry_dtype)
 
-    def num_reference_derivative(self, ref_axes, vec):
```
thomasgibson, line 295: :warning: This code is just moved to `meshmode/discretization/nodal.py`

meshmode/discretization/nodal.py
```diff
@@ -0,0 +1,169 @@
...
+# underscored because it shouldn't be imported from here.
+from meshmode.dof_array import DOFArray as _DOFArray
+
+from meshmode.discretization import DiscretizationBase
+from meshmode.discretization.poly_element import (
+        ModalSimplexGroupFactory,
+        ModalTensorProductGroupFactory)
+
+
+class NodalDiscretization(DiscretizationBase):
```
thomasgibson, line 42: :warning: This code is functionally equivalent to what was previously moved from `meshmode/discretization/__init__.py` and is unchanged.

meshmode/discretization/__init__.py
```diff
@@ -292,90 +359,9 @@ def empty_like(self, array: _DOFArray):
...
-                for grp in self.groups))
-
-    @memoize_method
-    def nodes(self):
-        r"""
-        :returns: object array of shape ``(ambient_dim,)`` containing
-            :class:`~meshmode.dof_array.DOFArray`\ s of node coordinates.
-        """
 
-        actx = self._setup_actx
```
thomasgibson, line 354: :warning:

meshmode/discretization/connection/modal.py
```diff
@@ -0,0 +1,263 @@
...
+                # to evaluate the Vandermonde matrix for non-interpolatory
+                # element groups
+                basis = mgrp.orthonormal_basis()
+                result = self._quadrature_projection(
+                    actx, ary, result, grp, basis)
+            else:
+                result = self._invert_vandermonde(
+                    actx, ary, result, grp)
+
+        return result
```
thomasgibson, line 183: @inducer Please could you have a look at this more carefully so we can discuss if this is more or less what we want to do?

meshmode/discretization/connection/modal.py
```diff
@@ -0,0 +1,263 @@
...
+            # Load Vandermonde in the array context
+            vdm = actx.from_numpy(V)
+
+            actx.call_loopy(
+                    keval(),
+                    result=result[grp.index],
+                    vdm=vdm,
+                    coefficients=ary[grp.index])
+
+        return result
```
thomasgibson, line 263: @inducer I think this is correct `(modal -> nodal)` but I would really appreciate it if you could please take a look so we can discuss this.

thomasgibson (commented):
@inducer I highlighted two key areas I would like some input on.

meshmode/discretization/connection/modal.py
```diff
@@ -0,0 +1,263 @@
...
+                        shape=("nelements", "n_to_dofs")),
+                    "..."
+                    ],
+                name="quadrature_proj_eval_knl")
+
+        for ibasis, basis_fn in enumerate(modal_basis_fns):
+
+            # Move data to array context for use in
+            # generated loopy kernels
+            basis = actx.from_numpy(basis_fn(grp.unit_nodes))
```
inducer, line 97: That's also a Vandermonde (and should be built using that function for consistency).

inducer, line 97: You could probably premultiply the quadrature weights into the matrix as well.

meshmode/discretization/connection/modal.py
```diff
@@ -0,0 +1,263 @@
...
+                    lp.GlobalArg("result", None,
+                        shape=("nelements", "n_to_dofs")),
+                    "..."
+                    ],
+                name="quadrature_proj_eval_knl")
+
+        for ibasis, basis_fn in enumerate(modal_basis_fns):
+
+            # Move data to array context for use in
+            # generated loopy kernels
```
inducer, line 96:
```suggestion
```

meshmode/discretization/connection/modal.py
```diff
@@ -0,0 +1,263 @@
...
+
+                # Use the modal (orthonormal basis) from the `to_discr`
+                # to evaluate the Vandermonde matrix for non-interpolatory
+                # element groups
+                basis = mgrp.orthonormal_basis()
+                result = self._quadrature_projection(
+                    actx, ary, result, grp, basis)
+            else:
+                result = self._invert_vandermonde(
+                    actx, ary, result, grp)
```
inducer, line 181: I'm skeptical of making this the fallback path. It only applies if the group is unisolvent and from/to have the same space. I.e. I think it would be better to make the `else` branch, `"I don't know how to deal with {from_group_type} and {to_group_type}"`.

meshmode/discretization/connection/modal.py
```diff
@@ -0,0 +1,263 @@
...
+        if ary.shape != (len(self.from_discr.groups),):
+            raise ValueError("Invalid shape of incoming nodal data")
+
+        actx = ary.array_context
+
+        result = self.to_discr.zeros(actx, dtype=ary.entry_dtype)
+
+        for igrp, grp in enumerate(self.from_discr.groups):
+
+            if isinstance(grp, QuadratureSimplexElementGroup):
```
inducer, line 165: This only applies if the basis is orthonormal. It'd be good to have this condition in the `if`, for clarity.

meshmode/discretization/connection/modal.py
```diff
@@ -0,0 +1,263 @@
...
+
+        if not isinstance(from_discr, ModalDiscretization):
+            raise ValueError("from_discr must be a ModalDiscretization "
+                    "object to use this connection.")
+
+        if not isinstance(to_discr, NodalDiscretization):
+            raise ValueError("to_discr must be a NodalDiscretization "
+                    "object to use this connection.")
+
+        super().__init__(
```
inducer, line 198: Should check that these are based on the same mesh.

meshmode/discretization/connection/modal.py
```diff
@@ -0,0 +1,263 @@
...
+
+        if not isinstance(from_discr, NodalDiscretization):
+            raise ValueError("from_discr must be a NodalDiscretization "
+                    "object to use this connection.")
+
+        if not isinstance(to_discr, ModalDiscretization):
+            raise ValueError("to_discr must be a ModalDiscretization "
+                    "object to use this connection.")
+
+        super().__init__(
```
inducer, line 55: Should check that these are based on the same mesh.

meshmode/discretization/connection/modal.py
```diff
@@ -0,0 +1,263 @@
...
+            if (isinstance(grp, QuadratureSimplexElementGroup) or
+                grp.order != self.from_discr.groups[igrp].order):
+                # Use the modal (orthonormal basis) from the `from_discr`
+                # to evaluate the (possibly non-square) Vandermonde matrix
+                # for non-interpolatory element groups
+                basis = self.from_discr.groups[igrp].orthonormal_basis()
+            else:
+                # Otherwise, we use the basis of the target space
+                # (in the `to_discr`)
+                basis = grp.basis()
```
inducer, line 250: Using the basis from the target is not necessarily correct. (Why not always use the from-basis?)

meshmode/discretization/connection/modal.py
```diff
@@ -0,0 +1,263 @@
...
+            else:
+                result = self._invert_vandermonde(
+                    actx, ary, result, grp)
+
+        return result
+
+
+class ModalInverseDiscretizationConnection(DiscretizationConnection):
+
+    def __init__(self, from_discr, to_discr):
```
inducer, line 188: Document (and check! if possible) the assumptions that must be true in order to use this:
- same mesh

test/test_modal.py
```diff
@@ -0,0 +1,129 @@
...
+
+@pytest.mark.parametrize("nodal_group_factory", [
+    InterpolatoryQuadratureSimplexGroupFactory,
+    PolynomialWarpAndBlendGroupFactory,
+    partial(PolynomialRecursiveNodesGroupFactory, family="lgl"),
+    PolynomialEquidistantSimplexGroupFactory,
+    LegendreGaussLobattoTensorProductGroupFactory,
+    QuadratureSimplexGroupFactory
+    ])
+def test_inverse_modal_connections(actx_factory, nodal_group_factory):
```
inducer, line 78: Also add a test for truncation order of modal truncation.

inducer (commented):
Thanks for workig on this. Mathematically, I have few complaints, but I did spot plenty of nitpicky things. :) Many of my comments apply generally, but I only pointed them out once or twice.

meshmode/discretization/connection/modal.py
thomasgibson: @inducer `QuadratureSimplexElementGroup` instances do not have bases --- rather the code currently returns this:
```python
    def basis(self):
        raise NoninterpolatoryElementGroupError("'{}' "
                "is not equipped with a unisolvent function space "
                "and therefore cannot be used for interpolation"
                .format(self.__class__.__name__))
```

What basis should I use here? Should I add a method to `QuadratureSimplexElementGroup` like:
```python
def basis(self):
    meg = self.mesh_el_group
    meg_space = mp.PN(meg.dim, meg.order)
    return mp.basis_for_space(meg_space, self._shape).functions
```
The groups associated with `self.to_discr` are always modal. So they are orthonormal by construction. Are you asking for me to add something like this:
```python
mgrp = self.to_discr.groups[igrp]
if (isinstance(grp, QuadratureSimplexElementGroup) and mgrp.is_orthonormal_basis()):
    ...
```
?

meshmode/mesh/__init__.py
```diff
@@ -282,6 +282,19 @@ def __eq__(self, other):
...
+        return hash((self.__class__,
+                     self.order,
+                     str(self.vertex_indices),
+                     str(self.nodes),
+                     str(self.unit_nodes),
+                     self.element_nr_base,
+                     self.node_nr_base))
+
+    def __hash__(self):
+        return self._hash()
```
thomasgibson, line 296: @inducer Are you happy with making `MeshElementGroup` hashable in this way? The reason is that I need to make `ElementGroupBase` (not to be confused with `MeshElementGroup` here....) hashable for caching of the Vandermonde matrix inverse. I do this by using the mesh element group attribute to generate a hash key (and in meshmode main, `MeshElementGroup` is currently not hashable).

For reference, see this comment: https://github.com/inducer/meshmode/pull/131/commits/dde334d987d14a5419906a3069544c682a0fa810

meshmode/discretization/nodal.py
```diff
@@ -161,9 +161,13 @@ def prg():
...
+class Discretization(NodalDiscretization):
+
+    def __init__(self, *args, **kwargs):
+
+        from warnings import warn
+        warn("`Discretization` will be dropped in favor of "
+             "`NodalDiscretization` in version 2022.x.",
+             DeprecationWarning, stacklevel=2)
+
+        super().__init__(*args, **kwargs)
```
thomasgibson, line 173: @inducer Are you happy with this approach?

meshmode/mesh/__init__.py
inducer: I don't think we need to hash all that data. (And I don't think we need the whole `ElementGroup` to be hashable, either.) What I would propose is to define a method that returns a unique identifier of the meaning of the DOFs. So, for most element groups, a tuple `(GroupClass, self.order)` should suffice. That's then trivially hashable, and it suffices as a hash key for the matrix.

meshmode/discretization/connection/modal.py
```diff
@@ -0,0 +1,246 @@
...
+    QuadratureSimplexElementGroup, InterpolatoryQuadratureSimplexElementGroup)
+from meshmode.discretization.connection.direct import DiscretizationConnection
+from meshmode.discretization.modal import ModalDiscretization
+from meshmode.discretization.nodal import NodalDiscretization
+
+from pytools import memoize_in
+from pytools.obj_array import obj_array_vectorized_n_args
+
+
+class ModalDiscretizationConnection(DiscretizationConnection):
```
inducer, line 44:
```suggestion
class NodalToModalDiscretizationConnection(DiscretizationConnection):
```

meshmode/discretization/connection/modal.py
```diff
@@ -0,0 +1,246 @@
...
+                    "%s to %s" % (grp.__class__.__name__,
+                                  mgrp.__class__.__name__)
+                    )
+
+            result_data += (output['result'],)
+
+        return DOFArray(actx, data=result_data)
+
+
+class ModalInverseDiscretizationConnection(DiscretizationConnection):
```
inducer, line 185:
```suggestion
class ModalToNodalDiscretizationConnection(DiscretizationConnection):
```

alexfikl:
The grudge failure is my fault. Sorry for that! There's some details in https://github.com/inducer/meshmode/pull/136#issuecomment-803209186

thomasgibson:
> The grudge failure is my fault. Sorry for that! There's some details in [#136 (comment)](https://github.com/inducer/meshmode/pull/136#issuecomment-803209186)

No problem, thank you for letting me know!

test/test_modal.py
thomasgibson: I've added one. I've opted to _not_ add another discretization connection (despite us talking about it) because: (1) I don't want to make this PR longer than it already is, and (2) I think there is a debate to be had about where it should belong. The truncation operation is really a type of (bad) spectral filter. There is a much nicer interface that @MTCam started in https://github.com/illinois-ceesd/mirgecom/pull/92/, which this PR could help simplify immediately out of the box.

meshmode/discretization/connection/modal.py
```diff
@@ -0,0 +1,349 @@
...
+                                            * coefficients[iel, ibasis])
+                """,
+                name="modinv_evaluation_knl")
+
+        result_data = []
+        for igrp, grp in enumerate(self.to_discr.groups):
+
+            basis_fns = self.from_discr.groups[igrp].orthonormal_basis()
+            vdm = mp.vandermonde(basis_fns, grp.unit_nodes)
+            vdm = actx.from_numpy(vdm)
```
inducer, line 342: See #138 for a caching structure to use. (here and above)

meshmode/discretization/__init__.py
```diff
@@ -107,9 +101,68 @@ def ndofs(self):
...
+
+    @abstractmethod
+    def basis(self):
+        """The basis spanning the underlying polynomial space."""
+
+    @abstractmethod
+    def grad_basis(self):
+        """The gradients of the basis functions spanning the
+        underlying polynomial space.
+        """
```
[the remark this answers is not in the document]
thomasgibson: @inducer So do you prefer not using abstract methods to decorate functions that need concrete implementations in derived classes?

meshmode/discretization/connection/modal.py
thomasgibson: @inducer Let me know when your pytools feature is merged, and I'll cherry-pick your commit for caching the inverse. And I guess based on this comment here that you also want to cache the Vandermonde matrix here (in addition to caching inverses as above)?

meshmode/discretization/__init__.py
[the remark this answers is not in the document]
inducer: No, I like the `@abstractmethod` decorator. I haven't been using it, but I agree that it's a good idea. As far as I understand the [documentation](https://docs.python.org/3/library/abc.html#abc.abstractmethod) though, it does *not* prevent anyone from calling this method. Maybe I'm missing something?

meshmode/discretization/connection/modal.py
inducer: > And I guess based on this comment here that you also want to cache the Vandermonde matrix here (in addition to caching inverses as above)?

Yes. I think this caching strategy makes sense, and I'd say we should use it everywhere. (`num_reference_diff`, other connections, ...) But that's likely best done in another PR.

meshmode/discretization/__init__.py
```diff
@@ -107,9 +101,68 @@ def ndofs(self):
...
+    def __hash__(self):
+        return hash((self.__class__,
+                     self.order,
+                     self.index))
+
+    def __eq__(self, other):
+        return (other.__class__ == self.__class__
+                and other.mesh_el_group == self.mesh_el_group
+                and other.order == self.order
+                and other.index == self.index)
```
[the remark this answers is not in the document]
inducer: > Once your pytools feature is merged

It's more than that. Since pytools is a pyopencl depedency (which is installed from the conda package here), it's only usable here after being merged, released to pypi and packaged for conda. (which sounds like a lot, but it's about 15 minutes end-to-end)

meshmode/discretization/connection/modal.py
thomasgibson: Sounds good, and I agree! I'll save that for a later PR

meshmode/discretization/__init__.py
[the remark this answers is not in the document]
thomasgibson: It's actually even stricter than that. Decorating abstract class methods with `@abstractmethod` will prevent accidental instantiation of derived classes that do not provide a concrete implementation. Here is an example of what I mean:
```
In [3]: class AbstractFoo(object, metaclass=abc.ABCMeta):
   ...:     @abc.abstractmethod
   ...:     def method1(self):
   ...:         pass
   ...: 

In [4]: class Foo(AbstractFoo):
   ...:     def method2(self):
   ...:         print("bar")
   ...: 

In [5]: f = Foo()
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-5-79a961337674> in <module>
----> 1 f = Foo()

TypeError: Can't instantiate abstract class Foo with abstract methods method1
```

meshmode/discretization/__init__.py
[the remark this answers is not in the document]
inducer: This should be all done now. (https://github.com/conda-forge/pytools-feedstock/pull/35)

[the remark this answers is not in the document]
thomasgibson: Awesome! Updating pytools and will cherry-pick your commits

meshmode/discretization/__init__.py
```diff
@@ -107,88 +106,208 @@ def ndofs(self):
...
+        associated with a single element.
+        """
+        return self.space.space_dim
+
+    @property
+    @memoize_method
+    def _orthonormal_basis(self):
+        import modepy as mp
+        return mp.orthonormal_basis_for_space(self.space,
+                                              self.shape)
```
inducer, line 260: I'd prefer to leave this abstract, with the actual basis specified by a subclass.

meshmode/discretization/__init__.py
```diff
@@ -107,88 +106,208 @@ def ndofs(self):
...
+
+    def grad_orthonormal_basis(self):
+        """Returns a :class:`tuple` of functions, each of which
+        accepts arrays of shape *(dims, npts)* and returns a
+        :class:`tuple` of length *dims* containing the
+        derivatives along each axis as an array of size
+        *npts*.  'Scalar' evaluation, by passing just one
+        vector of length *dims*, is also supported.
+        """
+        return self._orthonormal_basis.gradients
```
inducer, line 284: I find it weird that we have two different spots where bases are introduced, with two different interfaces. How about this hierarchy instead?

image

Graphviz in case you'd like to counterpropose:
```
digraph {
  ElementGroupBase -> ElementGroupWithBasis;
  ElementGroupWithBasis -> ModalElementGroup;
  ElementGroupBase -> NodalElementGroup;
  ElementGroupWithBasis -> InterpolatoryElementGroup;
  NodalElementGroup -> InterpolatoryElementGroup;
}
```

meshmode/discretization/__init__.py
```diff
@@ -390,4 +520,13 @@ def resample_mesh_nodes(grp, iaxis):
                 ]))
             for iaxis in range(self.ambient_dim)])
 
+    def nodes(self):
+        from warnings import warn
+        warn("`Discretization.nodes` will be dropped in version 2022.x. "
+             "use `Discretization.coordinates` instead.",
+             DeprecationWarning, stacklevel=2)
+        return self.coordinates()
```
inducer, line 528: I propose we keep this, but make it error if any element group is non-nodal. That way, you can (somewhat) safely) use the output of `nodes()` in nonlinear expressions, something that is not true of `coordinates`.

inducer (commented):
Thanks! I like how this is shaping up. A few thoughts below, but I think we're close.

inducer:

meshmode/discretization/__init__.py
thomasgibson: @inducer I like this idea a lot. I've gone ahead and implemented such a hierarchy. I completely agree that it doesn't make much sense to have two different basis interfaces, and I think this approach is a clean solution (b68083da79d7c1b0c453e2d5f7eaea93c0a98b94).

Could you check the associated docstrings for these different base classes? I wanted to be concrete about what `ElementGroupWithBasis`/`NodalElementGroup`/`ModalElementGroup`/etc. actually are and I would appreciate your feedback!

meshmode/discretization/__init__.py
thomasgibson: Done in b68083da79d7c1b0c453e2d5f7eaea93c0a98b94

meshmode/discretization/__init__.py
```diff
@@ -83,111 +83,249 @@ def __init__(self, mesh_el_group, order, index):
...
+
+# {{{ Nodal element group base
+
+class NodalElementGroupBase(ElementGroupBase):
+    """Base class for nodal element groups, defined as
+    finite elements equipped with a quadrature rule. Nodal
+    element groups may not necessarily have the same number
+    of quadrature nodes as the dimension of
+    :attr:`~ElementGroupBase.space`, and hence is not generally
+    usable for interpolation.
```
thomasgibson, line 158: Upon reading this docstring this morning, I'm not quite satisfied with what I wrote. I'm thinking perhaps a better description would be to explicitly define what we mean by "node." I'm thinking something along the lines of this:
```python
    """Base class for nodal element groups, defined as
    finite elements equipped with nodes. Nodes are specific
    locations defined on the reference element (:attr:~ElementGroupBase.shape`)
    for which a degree of freedom is defined. Such element groups
    have an associated quadrature rule to perform numerical integration,
    but are not necessarily usable for interpolation.
    """
```

thomasgibson: @inducer

meshmode/discretization/__init__.py
```diff
@@ -83,111 +83,249 @@ def __init__(self, mesh_el_group, order, index):
...
-            vector of length *dims*, is also supported.
+# {{{ Element groups suitable for interpolation
 
-    .. method:: diff_matrices()
+class InterpolatoryElementGroupBase(NodalElementGroupBase,
+                                    ElementGroupWithBasis):
+    """An element group equipped with both an explicit basis and
+    quadrature rule. These element groups have the same number of
+    interpolation nodes and basis function, making them suitable for
+    interpolation and differentiation.
```
thomasgibson, line 247: @inducer How about this?
```python
"""An element group equipped with both an explicit basis for the
underlying :attr:`~ElementGroupBase.space`, and a set of nodal
locations on the :attr:`~ElementGroupBase.shape`. These element
groups are unisolvent in the Ciarlet sense, meaning the dimension
of :attr:`~ElementGroupBase.space` matches the number of
interpolatory nodal locations. These element groups are therefore
suitable for interpolation and differentiation.
"""
```

thomasgibson: Maybe `CiarletElementGroupBase` could be an alternative name? :) But to be honest, I'm quite okay with leaving it as is, as I don't want to cause too large of a ripple-effect for any meshmode dependencies.

meshmode/discretization/__init__.py
inducer: > have an associated quadrature rule to perform numerical integration

I hadn't thought of this as a necessary feature of a nodal group. (If it is, "having a quadrature" would seem like the defining feature and maybe should end up in the name.) I could imagine just plopping some nodes onto an element without specifying a quadrature. I don't know that that's a useful thing though...

inducer: (I could see renaming nodal -> quadrature? What do you think?)

thomasgibson: @inducer I feel like an element with nodes and no notion of quadrature feels... empty? And since `QuadratureSimplexElementGroup` inherits from this directly, it makes sense to potentially rename `NodalElementGroupBase` to `QuadratureElementGroupBase`. It also makes sense for `InterpolatoryElementGroupBase`, since it's basically an element group with both an explicitly defined basis and a quadrature rule. So it fits within the proposed class hierarchy well I think.

Alternatively, I feel like a class hierarchy like below is a bit strange, but I could be convinced otherwise.
graphviz

Thoughts?

thomasgibson: Versus this: 
graphviz (1)

thomasgibson: @inducer What about `NodalElementGroup` -> `ElementGroupWithQuadrature`? (sort of in line with `ElementGroupWithBasis`).

Related question: Do we need `NodalElementGroup` at all if we have both `ElementGroupWithBasis` and `ElementGroupWithQuadrature`? I think this pretty much covers the extent of what meshmode cares about when building discretizations.

In my mind, it doesn't make sense to have an element group with nodes _and_ no specified basis or quadrature rule. Unless perhaps we want to handle point-wise defined fields? (Thinking in terms of physics models defined on grids). In which case, "node" is implicitly being defined at a coordinate on a cell. This is, however, stepping out of scope for a finite element, because `ElementGroupBase` is contracting out to subclasses that we have a function space and degrees of freedom (which the subclasses define concretely, as finite element DOFS are defined at nodal locations).

Perhaps we're overthinking this :)

meshmode/discretization/connection/modal.py
```diff
@@ -0,0 +1,373 @@
...
+    be used with both unisolvent and non-unisolvent element groups.
+
+    .. note::
+
+        This connection requires that both nodal and modal discretizations
+        are defined on the *same* mesh.
+
+    .. attribute:: from_discr
+
+        an instance of :class:`meshmode.discretization.Discretization` containing
```
alexfikl, line 55:
```suggestion
        An instance of :class:`meshmode.discretization.Discretization` containing
```
And elsewhere?

meshmode/discretization/connection/modal.py
```diff
@@ -0,0 +1,373 @@
...
+        :class:`meshmode.discretization.NodalElementGroupBase` element groups
```
alexfikl, line 56:
```suggestion
        :class:`~meshmode.discretization.NodalElementGroupBase` element groups
```
? These names can get a bit long when looking through the generated docs.

meshmode/discretization/connection/modal.py
```diff
@@ -0,0 +1,373 @@
...
+        an instance of :class:`meshmode.discretization.Discretization` containing
+        :class:`meshmode.discretization.ModalElementGroupBase` element groups
+
+    .. attribute:: groups
+
+        a list of :class:`DiscretizationConnectionElementGroup`
+        instances, with a one-to-one correspondence to the groups in
+        :attr:`to_discr`.
+
+    .. automethod:: __call__
```
alexfikl, line 69:
```suggestion
    .. automethod:: __init__
    .. automethod:: __call__
```
?

meshmode/discretization/connection/modal.py
```diff
@@ -0,0 +1,373 @@
...
+                    for grp in from_discr.groups]):
+            raise ValueError("`from_discr` must be defined on nodal "
+                             "element groups to use this connection.")
+
+        if not all([isinstance(grp, ModalElementGroupBase)
+                    for grp in to_discr.groups]):
+            raise ValueError("`to_discr` must be defined on modal "
+                             "element groups to use this connection.")
+
+        if to_discr.mesh != from_discr.mesh:
```
alexfikl, line 93:
```suggestion
        if to_discr.mesh is not from_discr.mesh:
```
? Otherwise this calls [\_\_eq\_\_](https://github.com/inducer/meshmode/blob/ac37ba5bc53223f40c195e25b23778cf6a004329/meshmode/mesh/__init__.py#L906-L916). Which one was intended?

meshmode/discretization/connection/modal.py
```diff
@@ -0,0 +1,373 @@
...
+            # we use an orthonormal basis and a quadrature rule
+            # to compute the modal coefficients.
+            if isinstance(grp, QuadratureSimplexElementGroup):
+
+                if (
+                    grp._quadrature_rule().exact_to < 2*mgrp.order
+                    and not self._allow_approximate_quad
+                ):
+                    raise ValueError("Quadrature rule is not exact, please "
+                                     "set `allow_approximate_quad=True`")
```
alexfikl, line 233: These look like two different errors? Even if `allow_approximate_quad` is True, it can still fail?

alexfikl (commented):
Found some small style issues and nitpicked a bit too. Feel free to close any of the issues, since I'm not sure how many actually apply.

The actual functionality looks good to me!

meshmode/discretization/connection/modal.py
thomasgibson: Ah, I see what you mean. The intention for this was to allow the user to basically say "I know the quadrature rule is inexact, but I want to proceed anyways via a quadrature projection." I will have a think about how I can make this a bit more clear!

meshmode/discretization/connection/modal.py
thomasgibson: I think calling `__eq__` is probably sufficient, but I like your suggestion as it's a stronger assertion. Will update :)

meshmode/discretization/connection/modal.py
alexfikl: Ah, I think I misunderstood the condition the first time I looked through it.

Your explanation makes sense to me, so feel free to close this!

meshmode/discretization/__init__.py
```diff
@@ -83,111 +83,249 @@ def __init__(self, mesh_el_group, order, index):
...
+        should cover all parts of the
+        `Ciarlet Triple <https://finite-element.github.io/L2_fespaces.html>`__:
+        reference element, shape functions, and the linear functionals defining
+        the degrees of freedom.) The object should be independent, however, of
+        the (global) elements that make up the group.
+
+        The structure of the element is not specified, but it must be globally
+        unique to this element group.
+        """
+        return (type(self), self.order)
```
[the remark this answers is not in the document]
thomasgibson: I don't think it can hurt, but I'm not sure it's strictly necessary here to unique define an element (since it's mostly agnostic whether you're in 2D, 3D, etc). At least in terms of defining a finite element abstractly.

Actually, now that I think about it, I'm wondering if we should include `shape` and/or `space`. One could argue that's already handled in `type(self)` since the classes themselves already inherently define both the space and shape. @inducer thoughts?

meshmode/discretization/__init__.py
```diff
@@ -296,6 +434,12 @@ def num_reference_derivative(self, ref_axes, vec):
         actx = vec.array_context
         ref_axes = list(ref_axes)
 
+        if not all([isinstance(grp, InterpolatoryElementGroupBase)
```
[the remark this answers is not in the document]
thomasgibson: Good idea! Will add this

meshmode/discretization/__init__.py
```diff
@@ -83,111 +83,249 @@ def __init__(self, mesh_el_group, order, index):
...
-        """The total number of (for now: nodal) degrees of freedom ("DOFs")
-        associated with the element group.
+        """Returns the total number of degrees of freedom ("DOFs")
+        associated with the entire element group.
         """
         return self.nunit_dofs * self.nelements
 
     @property
     def dim(self):
+        """Returns the geometric dimension of the element group."""
```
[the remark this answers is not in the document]
thomasgibson: Good catch, I've updated the description!

meshmode/discretization/__init__.py
```diff
@@ -31,45 +35,41 @@
...
+.. autoclass:: NodalElementGroupBase
+.. autoclass:: ElementGroupWithBasis
 .. autoclass:: InterpolatoryElementGroupBase
+.. autoclass:: ModalElementGroupBase
 .. autoclass:: Discretization
 """
 
 
-# {{{ element group base
+class ElementGroupError(TypeError):
```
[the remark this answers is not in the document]
thomasgibson: Done :)

meshmode/discretization/__init__.py
thomasgibson: @inducer I've made my decision to leave `NodalElementGroupBase` as is for now and clarified in the docstring.

meshmode/discretization/__init__.py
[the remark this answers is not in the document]
alexfikl: I was thinking in that direction too, i.e. `type(self)` should define the shape and the space classes (e.g. `Simplex` and `PN`) and ``(dims, order)`` should define exactly which ones (since the constructor is `PN(dims, order)`).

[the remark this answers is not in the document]
thomasgibson: I'm sold :) I'll update the key definition

thomasgibson:
> Found some small style issues and nitpicked a bit too. Feel free to close any of the issues, since I'm not sure how many actually apply.
> 
> The actual functionality looks good to me!

Your comments were very helpful, and I'm quite happy now with how this PR looks

meshmode/discretization/__init__.py
```diff
@@ -83,111 +100,247 @@ def __init__(self, mesh_el_group, order, index):
...
-        shape ``(nunit_nodes, nunit_nodes)``, each of which,
-        when applied to an array of nodal values, take derivatives
-        in the reference (r,s,t) directions.
+class ModalElementGroupBase(ElementGroupWithBasis):
+    """An element group equipped with a function space
+    and a hierarchical basis that is orthonormal with
+    respect to the :math:`L^2` inner product.
+
+    .. autoattribute:: nunit_dofs
+    .. autoattribute:: is_orthogonal_basis
```
inducer, line 300: IMO: No need to document functions that are just being implemented here.

meshmode/discretization/__init__.py
```diff
@@ -83,111 +100,247 @@ def __init__(self, mesh_el_group, order, index):
...
+        return True
+
 # }}}
 
 
+# {{{ discretization
+
 class Discretization:
-    """An unstructured composite discretization.
+    """An unstructured composite discretization class.
```
inducer, line 324:
```suggestion
    """An unstructured composite discretization.
```

meshmode/discretization/__init__.py
```diff
@@ -248,6 +401,24 @@ def ambient_dim(self):
...
         return sum(grp.ndofs for grp in self.groups)
 
+    @property
+    @memoize_method
+    def is_nodal(self):
+        """A :class:`bool` indicating whether the :class:`Discretization`
+        is defined over element groups subclasses of :class:`NodalElementGroupBase`.
+        """
+        return all([isinstance(grp, NodalElementGroupBase)
+                    for grp in self.groups])
```
inducer, line 411:
```suggestion
        return all(isinstance(grp, NodalElementGroupBase)
                   for grp in self.groups)
```

meshmode/discretization/__init__.py
```diff
@@ -248,6 +401,24 @@ def ambient_dim(self):
...
+
+    @property
+    @memoize_method
+    def is_modal(self):
+        """A :class:`bool` indicating whether the :class:`Discretization`
+        is defined over element groups subclasses of :class:`ModalElementGroupBase`.
+        """
+        return all([isinstance(grp, ModalElementGroupBase)
+                    for grp in self.groups])
```
inducer, line 420:
```suggestion
        return all(isinstance(grp, ModalElementGroupBase)
                   for grp in self.groups)
```

meshmode/discretization/__init__.py
```diff
@@ -296,6 +467,12 @@ def num_reference_derivative(self, ref_axes, vec):
         actx = vec.array_context
```
inducer, line 467: Could refactor this guy into a function as well, replacing this method with a deprecated wrapper?

meshmode/discretization/connection/modal.py
```diff
@@ -0,0 +1,368 @@
...
+
+        super().__init__(
+                from_discr=from_discr,
+                to_discr=to_discr,
+                is_surjective=True)
+
+        self._allow_approximate_quad = allow_approximate_quad
+
+    def _project_via_quadrature(self, actx, ary, grp, mgrp):
+
```
inducer, line 101:
```suggestion
```

meshmode/discretization/connection/modal.py
```diff
@@ -0,0 +1,368 @@
...
+    def __init__(self, from_discr, to_discr, allow_approximate_quad=False):
+        """
+        :arg from_discr: a :class:`meshmode.discretization.Discretization`
+            containing :class:`~meshmode.discretization.NodalElementGroupBase`
+            element groups.
+        :arg to_discr: a :class:`meshmode.discretization.Discretization`
+            containing :class:`~meshmode.discretization.ModalElementGroupBase`
+            element groups.
+        """
```
inducer, line 79: Document `allow_approximate_quad`.

meshmode/discretization/connection/modal.py
```diff
@@ -0,0 +1,368 @@
...
+
+            y = V^{-1} [\text{nodal basis coefficients}]
+
+        where :math:`V_{i,j} = \phi_j(x_i)` is the generalized
+        Vandermonde matrix, :math:`\phi_j` is a nodal basis,
+        and :math:`x_i` are nodal points on the reference
+        element defining the nodal discretization.
+
+        For non-interpolatory element groups (for example,
+        :class:`~meshmode.discretization.poly_element.QuadratureSimplexElementGroup`),
```
inducer, line 187: `NodalElementGroup` would probably be better here?

meshmode/discretization/connection/modal.py
```diff
@@ -0,0 +1,368 @@
...
+
+        output = actx.call_loopy(vinv_keval(),
+                                 vdm_inv=vandermonde_inverse(grp),
+                                 nodal_coeffs=ary[grp.index])
+
+        return output
+
+    @obj_array_vectorized_n_args
+    def __call__(self, ary):
+        r"""Computes modal coefficients data from a functions
```
inducer, line 173: Not sure: Maybe move this to the class docstring?

meshmode/discretization/connection/modal.py
```diff
@@ -0,0 +1,368 @@
...
+                             "the same mesh.")
+
+        super().__init__(
+                from_discr=from_discr,
+                to_discr=to_discr,
+                is_surjective=True)
+
+    @obj_array_vectorized_n_args
+    def __call__(self, ary):
+        r"""Computes nodal coefficients from modal data via
```
inducer, line 311: As above: move to class docstring?

meshmode/discretization/connection/modal.py
```diff
@@ -0,0 +1,368 @@
...
+            return actx.from_numpy(vdm)
+
+        result_data = []
+        for igrp, grp in enumerate(self.to_discr.groups):
+            output = actx.call_loopy(keval(),
+                                     vdm=matrix(grp, self.from_discr.groups[igrp]),
+                                     coefficients=ary[grp.index])
+            result_data.append(output["result"])
+
+        return DOFArray(actx, data=tuple(result_data))
```
inducer, line 368: Use a generator expression to package up this whole thing into one expression.

inducer (commented):
Thanks! A few (mostly documentation) comments below.

meshmode/discretization/connection/modal.py
thomasgibson: I actually prefer to use a concrete element group example rather than an abstract base class.

meshmode/discretization/connection/modal.py
thomasgibson: Done.

inducer:

meshmode/discretization/__init__.py
```diff
@@ -83,111 +100,247 @@ def __init__(self, mesh_el_group, order, index):
...
+        basis spanning the underlying :attr:`~ElementGroupBase.space`.
+        """
+
+    @abstractmethod
+    def diff_matrices(self):
+        """Return a :attr:`~ElementGroupBase.dim`-long :class:`tuple` of
+        :class:`numpy.ndarray` of shape ``(nunit_nodes, nunit_nodes)``,
+        each of which, when applied to an array of nodal values, take
+        derivatives in the reference :math:`(r, s, t)` directions.
+        """
```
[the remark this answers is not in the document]
thomasgibson: I don't disagree with your point here at all. However, these refactoring changes (and similar ones you've mentioned) I think are best left for another PR IMO. This PR has already introduced a significant amount of new lines (mainly documentation), while not fundamentally doing any significant refactoring or functional changes.

meshmode/discretization/__init__.py
thomasgibson: See previous comment about refactoring

meshmode/discretization/__init__.py
```diff
@@ -83,111 +100,247 @@ def __init__(self, mesh_el_group, order, index):
...
+    .. automethod:: grad_basis
+    """
 
-        Return an immutable sequence of opaque (hashable) mode identifiers,
+    @abstractmethod
+    def mode_ids(self):
+        """Return an immutable sequence of opaque (hashable) mode identifiers,
         one per element of the :meth:`basis`. The meaning of the mode
         identifiers is defined by the concrete element group.
+        """
```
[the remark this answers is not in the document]
thomasgibson: @inducer The only advantage (from my perspective) is that this only really cuts down on the number of methods (which I can totally understand). On the other hand, these methods also seem like completely reasonable things to ask of an element group. Could you give me an explicit argument for why you think these methods should go?

meshmode/discretization/__init__.py
```diff
@@ -83,111 +100,247 @@ def __init__(self, mesh_el_group, order, index):
...
-    function space.
+class ElementGroupWithBasis(ElementGroupBase):
+    """Base class for element groups which possess an
+    explicit basis for the underlying function space
+    :attr:`~ElementGroupBase.space`.
 
-    .. method:: mode_ids()
+    .. automethod:: mode_ids
+    .. automethod:: basis
+    .. automethod:: is_orthogonal_basis
```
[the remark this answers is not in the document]
thomasgibson: Sure, I agree. I can revise/replace with an orthonormal check and deprecate the existing one.

meshmode/discretization/__init__.py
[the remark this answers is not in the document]
inducer: OK.

meshmode/discretization/__init__.py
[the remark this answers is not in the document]
inducer: Maybe you're right, and this is unnecessary churn. My impulse was that there are now conventions for obtaining this info (the one in `Basis`, which came later, and this one), and it seemed worthwhile to unify them. OK to resolve, depending on what you think.

meshmode/discretization/__init__.py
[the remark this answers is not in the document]
inducer: #141

meshmode/discretization/__init__.py
inducer: #141

meshmode/discretization/__init__.py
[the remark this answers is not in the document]
thomasgibson: @inducer I'll push a commit that exposes the `Basis` guy and we can take a look to see if we like it! :)

meshmode/discretization/__init__.py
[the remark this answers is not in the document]
thomasgibson: Done in https://github.com/inducer/meshmode/pull/131/commits/a21c9ac2ba9ac5a078fd3adbf4b660380b63bd6d

meshmode/discretization/__init__.py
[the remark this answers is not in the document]
thomasgibson: @inducer I've implemented the suggestion in this commit https://github.com/inducer/meshmode/pull/131/commits/a9cc3f26f947863975bbee1020dd82de372fd9d9

I actually don't mind it, what do you think?

meshmode/discretization/__init__.py
```diff
@@ -83,111 +104,264 @@ def __init__(self, mesh_el_group, order, index):
...
+        """Returns a :class:`bool` flag that is *True* if the
+        basis corresponding to the element group is orthonormal
+        with respect to the :math:`L^2` inner-product.
+        """
+        import modepy as mp
+        try:
+            # Check orthonormality weight
+            # FIXME: Modepy expects an `rvec` argument for simplicial
+            # bases, while tensor product bases do not.
+            return self.get_basis().orthonormality_weight([]) == 1
```
thomasgibson, line 264: Will remove this ugly hack once https://github.com/inducer/modepy/pull/26 is merge :)

meshmode/discretization/poly_element.py
```diff
@@ -396,25 +457,22 @@ def __init__(self, mesh_el_group, order, index, *, basis, unit_nodes):
...
-            # with weight 1, which is why this check is stricter than expected.
-            return self._basis.orthonormality_weight() == 1
+            # The reason this function is repeated here is
+            # that `orthonormality_weight` does not require
+            # an `rvec` argument, while simplicial bases do.
+            # FIXME: This needs to be fixed if we want to avoid
+            # duplicating this method.
+            return self.get_basis().orthonormality_weight() == 1
         except mp.BasisNotOrthonormal:
             return False
```
thomasgibson, line 471: This function can be removed once https://github.com/inducer/modepy/pull/26, since the base class `ElementGroupWithBasis` will have a generic method that will work for all basis types

thomasgibson:
This PR seems to be getting the following error when running the CI examples (Conda Py3): `CondaHTTPError: HTTP 502 BAD GATEWAY for url <https://conda.anaconda.org/conda-forge/noarch/prompt-toolkit-3.0.18-pyha770c72_0.tar.bz2>`

Any ideas @inducer ?

thomasgibson:
Nevermind. Appears that the ol' "turn it off and back on again" approach (cancel -> rerun) cleared up the problem.

inducer (commented):
I noticed just one minor thing that I'll fix myself, the rest of it looked great. Thanks for working on this!

PR #133, opened 2021-03-15
Deprecate n parameter in generate_*_rect_mesh

majosm:


majosm:
@inducer Any idea what's up with the install failures in the Firedrake checks? ([example](https://github.com/inducer/meshmode/pull/133/checks?check_run_id=2142540168))

inducer:
I think that was me. I made some [changes to loopy](https://github.com/inducer/loopy/pull/242/), which required changes in sumpy and pytential. I'll take care of that.

inducer:
> Could not terminate the child.

That CI failure happens now and then. It's something to do with Octave. I don't know how to fix it.

meshmode/mesh/generation.py
```diff
@@ -949,27 +949,55 @@ def generate_box_mesh(axis_coords, order=1, coord_dtype=np.float64,
...
+                               n=None,
                                ):
     """Create a semi-structured rectangular mesh with equispaced elements.
 
     :param a: the lower left hand point of the rectangle
     :param b: the upper right hand point of the rectangle
-    :param n: a tuple of integers indicating the total number of points
+    :param nel: an optional tuple of integers indicating the total number of elements
+      on [a,b].
+    :param npts: an optional tuple of integers indicating the total number of points
```
inducer, line 964: How about `nelements_1d` and `npoints_1d`?

(As it is, it's inconsistent for whether it includes a plural-"s".)

majosm:
@inducer Have you ever seen the tests hang like [this](https://github.com/inducer/meshmode/pull/133/checks?check_run_id=2247138727) before? It's happened a couple of times on this PR so far. I think it's intermittent... not sure what could be causing it. I'll try running the CI again.

inducer:
> @inducer Have you ever seen the tests hang like [this](https://github.com/inducer/meshmode/pull/133/checks?check_run_id=2247138727) before? It's happened a couple of times on this PR so far. I think it's intermittent... not sure what could be causing it. I'll try running the CI again.

I can't see those logs any more... Github replaces the logs (at the same URL) when you rerun. (:facepalm:) FWIW, whenever it's happened to me, it had to do with some Github outage. https://www.githubstatus.com/ Turns out there's something going on ATM:

grafik

majosm:
Ran CI again today, and it's still doing it. Hmm...

meshmode/mesh/generation.py
```diff
@@ -949,27 +949,58 @@ def generate_box_mesh(axis_coords, order=1, coord_dtype=np.float64,
...
+    else:
+        if npoints_per_axis is not None:
+            if nelements_per_axis is not None:
+                raise ValueError("cannot specify both nelements_per_axis and "
+                    "npoints_per_axis")
+        else:
+            if nelements_per_axis is not None:
+                npoints_per_axis = tuple(nel_i+1 for nel_i in nelements_per_axis)
+            else:
+                npoints_per_axis = (5, 5)
```
inducer, line 997: ?

Not a fan.

meshmode/mesh/generation.py
```diff
@@ -981,15 +1012,36 @@ def generate_regular_rect_mesh(a=(0, 0), b=(1, 1), n=(5, 5), order=1,
...
+            if nelements_side is not None:
+                raise ValueError("cannot specify both nelements_side and "
+                    "npoints_side")
+        elif nelements_side is not None:
+            npoints_side = nelements_side + 1
 
     assert dim in [1, 2, 3]
+
+    npoints_per_axis = (npoints_side,)*dim if npoints_side is not None else None
+
```
inducer, line 1041: Why not pass all these on to `generate_regular_rect_mesh` to avoid duplicating the logic?

meshmode/mesh/generation.py
```diff
@@ -981,15 +1012,36 @@ def generate_regular_rect_mesh(a=(0, 0), b=(1, 1), n=(5, 5), order=1,
 
 # {{{ generate_warped_rect_mesh
 
-def generate_warped_rect_mesh(dim, order, n, *, group_cls=None):
+def generate_warped_rect_mesh(dim, order, *, nelements_side=None,
+        npoints_side=None, group_cls=None, n=None):
```
inducer, line 1016: Why not go with the `per_axis` names here, too? I'll grant you that it's slightly weird, but I'd prefer the consistency, I think.

inducer (commented):
Thanks! A few simple things below.

meshmode/mesh/generation.py
majosm: That was the original default value for n, which I left in for compatibility. If we're going to break compatibility anyway with the `*` change, I can change this too.

meshmode/mesh/generation.py
majosm: Despite appearing similar, the `n` parameters in `generate_regular_rect_mesh` and `generate_warped_rect_mesh` are/were different. In the regular case it's a tuple, and here it's just a number (so all sides end up having the same number of points). That's also why I named the new parameters for the warped version `*_side` instead of `*_per_axis`. I could possibly change `generate_warped_rect_mesh` to take tuples instead. Should I?

majosm: Though `generate_warped_rect_mesh` also doesn't have `a` and `b` parameters, so the interfaces still wouldn't be the same afterwards. 🤷‍♂️

meshmode/mesh/generation.py
majosm: I removed it.

meshmode/mesh/generation.py
inducer: Why not make them outright the same? Taking `a` and `b` doesn't make sense IMO, because the mapping used in the function is only good (invertible) for a certain range of values.

majosm: I guess I didn't see much of a point in trying to make the parameters look the same when the functions' interfaces were going to be different anyway due to the other parameters.

FWIW, I made an attempt at unifying the two interfaces: https://github.com/majosm/meshmode/commit/117936c579cd57091c7d1e8a97c9b04bbb74b341

It does the warping using the original (-0.5, 0.5)^dim mesh and then translates/scales it to proportionally match `a` and `b`. Here's what it looks like:

Default extents (blue=regular, red=warped)
----------------------------------------------
default_a_b

Tweaked extents (blue=regular, red=warped)
------------------------------------------------
custom_a_b

I wasn't sure how particular the form of the warping function was, so I tried not to mess with it. (If it's not that specific, I could probably tweak it to make the warped/unwarped mesh sizes match up more closely too.)

Thoughts?

majosm: I suppose I could also just rescale such that the lower and upper corners of the warped mesh end up at `a` and `b`:

normalized

([code](https://github.com/majosm/meshmode/commit/b1432a30bbbe789ec72b9883f5006719b67f919b))

Not sure if any of the tests depend on the original warped mesh size, I need to check.

inducer:
You know, the current state works for me. We've got bigger fish to fry. Thanks for working on this, and for humoring me on the warped function stuff.

inducer (approved):

PR #135, opened 2021-03-19
Skip node resampling in discretization

alexfikl:
This feels a bit dumb, but it works for my use case. Thoughts?

Work towards last point in #123

### DIFF
--- a/meshmode/discretization/__init__.py
+++ b/meshmode/discretization/__init__.py
@@ -366,16 +366,26 @@ class Discretization:
                     """,
                 name="nodes")
 
+        def resample_mesh_nodes(grp, iaxis, tol=1.0e-15):
+            # TODO: would be nice to have the mesh use an array context already
+            nodes = actx.from_numpy(grp.mesh_el_group.nodes[iaxis])
+
+            grp_unit_nodes = grp.unit_nodes
+            meg_unit_nodes = grp.mesh_el_group.unit_nodes
+            if (grp_unit_nodes.shape == meg_unit_nodes.shape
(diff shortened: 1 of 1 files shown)


meshmode/discretization/__init__.py
```diff
@@ -366,16 +366,26 @@ def prg():
...
                 name="nodes")
 
+        def resample_mesh_nodes(grp, iaxis, tol=1.0e-15):
+            # TODO: would be nice to have the mesh use an array context already
+            nodes = actx.from_numpy(grp.mesh_el_group.nodes[iaxis])
+
+            grp_unit_nodes = grp.unit_nodes
+            meg_unit_nodes = grp.mesh_el_group.unit_nodes
+            if (grp_unit_nodes.shape == meg_unit_nodes.shape
+                    and np.linalg.norm(grp_unit_nodes - meg_unit_nodes) < tol):
```
inducer, line 376: Maybe flatten this into 1D before applying the norm? Otherwise you're (morally) applying a matrix norm, even if it's just Frobenius.

meshmode/discretization/__init__.py
```diff
@@ -366,16 +366,26 @@ def prg():
                     """,
                 name="nodes")
 
+        def resample_mesh_nodes(grp, iaxis, tol=1.0e-15):
```
inducer, line 369: Get the tol via `finfo`?

meshmode/discretization/__init__.py
alexfikl: Done in f24f19b.

meshmode/discretization/__init__.py
alexfikl: Done in f24f19b

PR #136, opened 2021-03-19
Add simple copy to Visualizer

alexfikl:
This tries to avoid rebuilding all the connectivity when doing IO for moving meshes (see #123).

I started doing a proper copy, but it seemed like too many things could get inconsistent. The current version just replaces the underlying discretization.

### DIFF
--- a/meshmode/discretization/visualization.py
+++ b/meshmode/discretization/visualization.py
@@ -185,20 +185,11 @@ class _VisConnectivityGroup(Record):
 
 
 def _check_discr_same_connectivity(discr, other):
-    def _cmp(grp, ogrp):
-        # NOTE: include more things here if they become necessary
-        return (
-                type(grp) == type(ogrp)
-                and grp.order == ogrp.order
-                and grp.nelements == ogrp.nelements
-                and grp.nunit_dofs == ogrp.nunit_dofs
-                and np.array_equal(grp.unit_nodes, ogrp.unit_nodes)
-                )
-
     if len(discr.groups) != len(other.groups):
         return False
 
-    if not all(_cmp(sg, og) for sg, og in zip(discr.groups, other.groups)):
+    if not all(sg.discretization_key() == og.discretization_key()
+            for sg, og in zip(discr.groups, other.groups)):
         return False
 
     return True
--- a/meshmode/discretization/visualization.py
+++ b/meshmode/discretization/visualization.py
@@ -183,6 +183,26 @@ class _VisConnectivityGroup(Record):
     def primitive_element_size(self):
         return self.vis_connectivity.shape[2]
 
+
+def _check_discr_same_connectivity(discr, other):
+    def _cmp(grp, ogrp):
+        # NOTE: include more things here if they become necessary
+        return (
+                type(grp) == type(ogrp)
+                and grp.order == ogrp.order
+                and grp.nelements == ogrp.nelements
+                and grp.nunit_dofs == ogrp.nunit_dofs
+                and np.array_equal(grp.unit_nodes, ogrp.unit_nodes)
+                )
+
+    if len(discr.groups) != len(other.groups):
+        return False
+
+    if not all(_cmp(sg, og) for sg, og in zip(discr.groups, other.groups)):
+        return False
+
+    return True
+
 # }}}
 
 
@@ -445,7 +465,11 @@ class Visualizer:
     """
 
     def __init__(self, connection,
-            element_shrink_factor=None, is_equidistant=False):
+            element_shrink_factor=None,
+            is_equidistant=False,
+            _vtk_connectivity=None,
+            _vtk_lagrange_connectivity=None):
+
         self.connection = connection
         self.discr = connection.from_discr
         self.vis_discr = connection.to_discr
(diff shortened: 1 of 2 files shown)


meshmode/discretization/visualization.py
```diff
@@ -455,6 +479,28 @@ def __init__(self, connection,
         self.element_shrink_factor = element_shrink_factor
         self.is_equidistant = is_equidistant
 
+        self._cached_vtk_connectivity = _vtk_connectivity
+        self._cached_vtk_lagrange_connectivity = _vtk_lagrange_connectivity
+
+    def copy_with_same_connectivity(self, actx, discr, skip_tests=False):
```
inducer, line 485: Document?

alexfikl: Added some docs in 21f9204.

alexfikl:
I can reproduce the grudge failure locally, but no idea currently what broke (maybe related to #135?)

alexfikl:
Maybe it's worth adding a `1.0e-16` in `EOCRecorder` too so that it doesn't do this?

alexfikl:
@inducer grudge should hopelly be fixed with inducer/grudge#59. Rerun?

test/test_meshmode.py
```diff
@@ -238,6 +238,52 @@ def test_visualizers(actx_factory, dim, group_cls):
...
+            PolynomialWarpAndBlendGroupFactory(target_order))
+
+    from meshmode.discretization.visualization import make_visualizer
+    vis = make_visualizer(actx, discr, target_order, force_equidistant=True)
+    assert vis._vtk_connectivity
+    assert vis._vtk_lagrange_connectivity
+
+    translated_vis = vis.copy_with_same_connectivity(actx, translated_discr)
+    assert translated_vis._cached_vtk_connectivity is not None
+    assert translated_vis._cached_vtk_lagrange_connectivity is not None
```
inducer, line 277: You could assert that these are the same objects as on the un-translated one. (in the `a is b` sense)

inducer (commented):
Thanks! A few things below.

meshmode/discretization/visualization.py
```diff
@@ -183,6 +183,26 @@ def nsubelements_per_element(self):
...
+                and grp.nelements == ogrp.nelements
+                and grp.nunit_dofs == ogrp.nunit_dofs
+                and np.array_equal(grp.unit_nodes, ogrp.unit_nodes)
+                )
+
+    if len(discr.groups) != len(other.groups):
+        return False
+
+    if not all(_cmp(sg, og) for sg, og in zip(discr.groups, other.groups)):
+        return False
```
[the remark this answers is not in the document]
alexfikl: Added a check that they're the same type in 9299246.

test/test_meshmode.py
alexfikl: Done in 9299246.

meshmode/discretization/visualization.py
```diff
@@ -183,6 +183,26 @@ def nsubelements_per_element(self):
...
+
+def _check_discr_same_connectivity(discr, other):
+    def _cmp(grp, ogrp):
+        # NOTE: include more things here if they become necessary
+        return (
+                type(grp) == type(ogrp)
+                and grp.order == ogrp.order
+                and grp.nelements == ogrp.nelements
+                and grp.nunit_dofs == ogrp.nunit_dofs
+                and np.array_equal(grp.unit_nodes, ogrp.unit_nodes)
```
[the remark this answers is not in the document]
alexfikl: Sure, I'm fine with waiting for that to finish up.

meshmode/discretization/visualization.py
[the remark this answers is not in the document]
thomasgibson: #131 no longer adds a `Discretization` hierarchy (and I'm pretty sure we will want to keep it this way), so I think you can safely remove the type check on `discr` and `other`!

meshmode/discretization/visualization.py
```diff
@@ -183,6 +183,29 @@ def nsubelements_per_element(self):
...
```
thomasgibson, line 195: Once #131 lands (hopefully soon), I wonder if the following would be sufficient:
```python
return (
        grp.discretization_key() == ogrp.discretization_key()
        and np.array_equal(grp.unit_nodes, ogrp.unit_nodes)
        )
```
I am pretty sure checking the discretization key (in its current form in #131) is not sufficient for what you want to check for, right? Because you also want to make sure you have the same nodes

thomasgibson (commented):
Some quick comments!

meshmode/discretization/visualization.py
inducer: To my mind, the way [`discretization_key` is defined](https://github.com/inducer/meshmode/blob/35d4c2546b5fb9db300ef621d2034e031f3a57da/meshmode/discretization/__init__.py#L149-L161), "same discretization key" implies "same nodes".

thomasgibson: I think so too probably. Unless we try to visualize using one of the `TensorProductElementGroups` with a `NotImplementedError` for discretization key. But I think we can safely avoid that problem since we do have other quad/hex element groups that do have well-defined discretization keys.

meshmode/discretization/visualization.py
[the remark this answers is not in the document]
alexfikl: Updated the previous commit to not include that change.

meshmode/discretization/visualization.py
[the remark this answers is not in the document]
inducer: Now merged. :tada:

meshmode/discretization/visualization.py
alexfikl: Agreed! Resolving this in favor of https://github.com/inducer/meshmode/pull/136#discussion_r598220229.

meshmode/discretization/visualization.py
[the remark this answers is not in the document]
alexfikl: 8e63381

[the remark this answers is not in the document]
inducer: Yup, that looks good. Thanks!

PR #137, opened 2021-03-22
Add moving mesh example

alexfikl:
Adds a simple example with a moving mesh to track some progress in #123.

### DIFF
--- a/meshmode/discretization/poly_element.py
+++ b/meshmode/discretization/poly_element.py
@@ -23,7 +23,7 @@ THE SOFTWARE.
 
 import numpy as np
 #import numpy.linalg as la
-from pytools import memoize_method, memoize
+from pytools import memoize_method
 from meshmode.mesh import (
         SimplexElementGroup as _MeshSimplexElementGroup,
         TensorProductElementGroup as _MeshTensorProductElementGroup)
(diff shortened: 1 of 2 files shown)


meshmode/discretization/poly_element.py
```diff
@@ -234,6 +234,11 @@ def weights(self):
                 np.ones(len(self.basis())))
 
 
+@memoize
+def _warp_and_blend_nodes(dim, order):
+    return mp.warp_and_blend_nodes(dim, order)
```
alexfikl, line 239: Caching the unit nodes here seems to be the biggest added gain. This is hack at the moment, but would be nice to figure something out.

Here's some profiling logs for with/without (both are html files)
[pyinstrument.zip](https://github.com/inducer/meshmode/files/6184163/pyinstrument.zip)

meshmode/discretization/__init__.py
```diff
@@ -368,7 +368,9 @@ def prg():
 
         def resample_mesh_nodes(grp, iaxis):
             # TODO: would be nice to have the mesh use an array context already
-            nodes = actx.from_numpy(grp.mesh_el_group.nodes[iaxis])
+            nodes = grp.mesh_el_group.nodes[iaxis]
+            if isinstance(nodes, np.ndarray):
+                nodes = actx.freeze(actx.from_numpy(nodes))
```
inducer, line 373: What other type is being accepted here? Could you make that explicit? I find the code hard to read without that.

inducer (commented):
Thanks! One question about types, otherwise good to go.

meshmode/discretization/__init__.py
alexfikl: I hacked together a `mesh_to_device` function while profiling a bit to see what that would save. This is just some remnant from that, so I went ahead and removed it.

meshmode/discretization/poly_element.py
alexfikl: @inducer Before merging, this should be removed too and leave it for another PR. 

Maybe figure out some more general caching for the unit nodes? Although there aren't many objects that survive when moving the discretization..

alexfikl: Removed in 3af9d7d

inducer:
Thanks!

PR #139, opened 2021-03-27
Add a generic ArrayContainer

alexfikl:
* Added a barebones `ArrayContainer` class that just defines a serialization protocol for its components.
* Implemented a `array_container_vectorize`, `freeze` and `thaw` that seem to work nice enough.

### DIFF
--- a/meshmode/array_context.py
+++ b/meshmode/array_context.py
@@ -20,15 +20,18 @@ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 THE SOFTWARE.
 """
 
-from typing import Union, Sequence
-from functools import partial
 import operator
+from functools import partial
+from abc import ABC, abstractmethod
+from typing import Any, Callable, Iterable, Sequence, Tuple, Union
+
 import numpy as np
+
 import loopy as lp
 from loopy.version import MOST_RECENT_LANGUAGE_VERSION
+
 from pytools import memoize_method
 from pytools.tag import Tag
-from abc import ABC, abstractmethod
 
 
 __doc__ = """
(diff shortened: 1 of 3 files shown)


meshmode/array_context.py
```diff
@@ -160,6 +160,16 @@ def loopy_implemented_elwise_func(*args):
         else:
             raise AttributeError(name)
 
+    def _new_like(self, x, alloc):
+        from meshmode.dof_array import obj_or_dof_array_vectorize
+        return obj_or_dof_array_vectorize(alloc, x)
+
+    def empty_like(self, x):
+        return self._new_like(x, self._array_context.empty_like)
```
alexfikl, line 168: This is maybe a bit redundant with `Discretization.empty_like`, but there are a few places where it could be useful, e.g. pytential has some `DOFArray`s that don't have the same `nunit_dofs` as the discretization.

meshmode/array_context.py
```diff
@@ -160,6 +160,16 @@ def loopy_implemented_elwise_func(*args):
         else:
             raise AttributeError(name)
 
+    def _new_like(self, x, alloc):
+        from meshmode.dof_array import obj_or_dof_array_vectorize
+        return obj_or_dof_array_vectorize(alloc, x)
```
alexfikl, line 165: This is not consistent with `np.zeros_like`, which returns `array([0, 0, 0], dtype=object)` when applied to an object array. Thoughts?

alexfikl:
Failure is in `test_nodal_dg_interop` again. This seems to be happening a lot more all of a sudden.

meshmode/array_context.py
```diff
@@ -335,22 +349,52 @@ def where(self, criterion, then, else_):
...
     def max(self, a):
         import pyopencl.array as cl_array
-        return cl_array.max(a, queue=self._array_context.queue).get()[()]
+        return self._reduce(a, max, partial(cl_array.max, queue=self._queue))
+
+    def dot(self, x, y):
+        if isinstance(x, np.ndarray) and x.dtype.char == "O":
+            return x @ y
+        else:
+            return self.sum(x * y)
```
inducer, line 390: This isn't correct in general (I think), and the general case is also depressingly difficult on GPUs, which is why there's no implementation in PyOpenCL. I don't think we can offer this with a straight face.

alexfikl: Agreed it's bit fishy, so it's removed in the latest version.

By "general" do you mean for arbitrary multidimensional arrays?

alexfikl:
For a bit of context, I had some of these laying around and added them with ecisneros8/pyrometheus#18 in mind. I would say that this can be left open until ecisneros8/pyrometheus#18 clears up and you see what exactly is needed.

On my side, I would mostly like to keep extending the reductions to `DOFArrays`.

meshmode/array_context.py
inducer: Yeah. https://numpy.org/doc/stable/reference/generated/numpy.dot.html

If you meant "for 1D", I could go with that, as long as you have `if len(x.shape) !=1 and ...: raise NotImplementedError`.

inducer:
> Failure is in `test_nodal_dg_interop` again. This seems to be happening a lot more all of a sudden.

Yeah. Seems to have been a side effect of Github CI upgrading to Ubuntu 20.x. The same issue had been happening on Gitlab for much longer than that, so it seems to be a legitimate issue with "modern" Linux. I've been too lazy to figure out what's happening though.

meshmode/array_context.py
inducer: Kind of a no-go. Within this namespace, numpy is king, even if it's inconvenient/bad, especially when applied to numpy input. If you don't want to reimplement, `NotImplementedError` is always an option.

meshmode/array_context.py
inducer: That's OK. Thanks for asking!

meshmode/array_context.py
```diff
@@ -335,22 +349,47 @@ def where(self, criterion, then, else_):
 
         def where_inner(inner_crit, inner_then, inner_else):
             return cl_array.if_positive(inner_crit != 0, inner_then, inner_else,
-                    queue=self._array_context.queue)
+                    queue=self._queue)
 
         return obj_or_dof_array_vectorize_n_args(where_inner, criterion, then, else_)
 
+    def _reduce(self, x, router, rinner):
```
inducer, line 356: Explain `router` and `rinner` in a docstring?

inducer, line 356: `router` reads poorly for me.

meshmode/array_context.py
```diff
@@ -335,22 +349,47 @@ def where(self, criterion, then, else_):
...
 
     def max(self, a):
         import pyopencl.array as cl_array
-        return cl_array.max(a, queue=self._array_context.queue).get()[()]
+        return self._reduce(a, max, partial(cl_array.max, queue=self._queue))
+
+    def ones_like(self, a):
+        from pytools.obj_array import obj_array_vectorize
+        b = self._array_context.np.zeros_like(a)
+        obj_array_vectorize(lambda bi: bi.fill(1), b)
```
inducer, line 390: That's an "innovative" use of `obj_array_vectorize`. I'm not objecting, but maybe add an apologetic comment?

inducer (commented):
Thanks! Some comments below.

meshmode/array_context.py
alexfikl: Not quite sure what I was thinking there :facepalm:, but made into a clear for loop in f2f5d75.

meshmode/array_context.py
alexfikl: Agreed! Added a `NotImplementedError` for object arrays and scalars in 5d7ee60.

meshmode/array_context.py
alexfikl: Cleaned that up quite a bit in 0c3d018. 

It now has the same signature as `functools.reduce`, so I didn't think a docstring was necessary. I can still add one if you want.

alexfikl: Removed the `rinner/router` stuff completely in 0c3d018.

meshmode/array_context.py
```diff
@@ -160,6 +160,27 @@ def loopy_implemented_elwise_func(*args):
         else:
             raise AttributeError(name)
 
+    def _new_like(self, x, alloc):
+        from numbers import Number
+        from meshmode.dof_array import DOFArray
+
+        if isinstance(x, DOFArray):
+            return x._like_me([alloc(xi) for xi in x])
```
inducer, line 168: I'm not super sure about this; it's a layering violation to me. (In that if array contexts moved to a separate package, suddenly we'd have a circular depdency between meshmode and it.)

Proposed solution: Make a base class `ArrayContainer` here and inherit `DOFArray` from it. (Where the point of an `ArrayContainer` would be to let you iterate over the arrays it contains, together with an opaque index structure, and then rebuild the container with the new arrays.)

meshmode/array_context.py
```diff
@@ -335,22 +360,53 @@ def where(self, criterion, then, else_):
...
+        def reduce_wrapper(ary):
+            return function(ary).get(self._queue)[()]
+
+        from numbers import Number
+        from meshmode.dof_array import DOFArray
+
+        if isinstance(x, Number):
+            return [x]
+        elif isinstance(x, DOFArray):
+            return [reduce_wrapper(xi) for xi in x]
```
inducer, line 377: As above.

meshmode/array_context.py
alexfikl: I didn't feel too bad about it because it was already in `np.linalg.norm`, but agreed that adding more cases makes it a lot more jarring.

I'll go ahead and code up a barebones `ArrayContainer` for this!

inducer: Thanks! I should have caught it then, but I didn't. (Or maybe I introduced it in the first place? Who knows)

alexfikl: I have a bit of an itch to move `thaw`/`freeze`/`obj_or_dof_array_vectorize*` in here from `dof_array` and make them work on the `ArrayContainer`. 

What do you think? Separate PR?

alexfikl: Gave it a go in fd1e2d7. Let me know what you think!

meshmode/array_context.py
```diff
@@ -21,19 +21,206 @@
...
+
+    def __iter__(self):
+        return iter(self._data)
+
+    # }}}
+
+    # {{{
+
+    @property
+    def shape(self):
```
alexfikl, line 104: This is mostly here because `np.linalg.norm` checks that the input is one-dimensional.

inducer:

alexfikl:
@inducer I removed everything from this PR except the `ArrayContainer` class and some helpers. Let me know if you want to continue working on this here or you guys had something else in mind!

alexfikl:
Closed in favor of #175.

PR #142, opened 2021-04-01
Move ElementGroup matrix methods to functions

### LINKED ISSUE #141: Method -> function refactoring
- `Discretization.num_reference_derivative` for sure
- Maybe: `InterpolatoryElementGroup.{diff_matrix,mass_matrices}`.

cc @thomasgibson 

x-ref:
- https://github.com/inducer/meshmode/pull/131#discussion_r602931630
- https://github.com/inducer/meshmode/pull/131#discussion_r602932124

alexfikl:
This does a couple of things
* Makes `num_reference_derivative` a function (part of #141)
* Makes `mass_matrix`, `diff_matrices` and `from_mesh_interp_matrix` functions (part of #141)
* ~~Adds the results to a global cache (helps with #123)~~

Fixes #141

### DIFF
--- a/examples/moving-geometry.py
+++ b/examples/moving-geometry.py
@@ -26,7 +26,7 @@ import pyopencl as cl
 from meshmode.dof_array import thaw
 from meshmode.array_context import PyOpenCLArrayContext
 
-from pytools import memoize_in, memoize_on_first_arg
+from pytools import memoize_in, keyed_memoize_in
 from pytools.obj_array import make_obj_array
 
 import logging
@@ -63,13 +63,17 @@ def reconstruct_discr_from_nodes(actx, discr, x):
             """,
             name="resample_by_mat_prg")
 
-    @memoize_on_first_arg
+    @keyed_memoize_in(actx,
+            (reconstruct_discr_from_nodes, "to_mesh_interp_matrix"),
+            lambda grp: grp.discretization_key())
     def to_mesh_interp_matrix(grp) -> np.ndarray:
         import modepy as mp
-        return mp.resampling_matrix(
+        mat = mp.resampling_matrix(
                 grp.basis_obj().functions,
                 grp.mesh_el_group.unit_nodes,
                 grp.unit_nodes)
+
+        return actx.freeze(actx.from_numpy(mat))
 
     def resample_nodes_to_mesh(grp, igrp, iaxis):
         discr_nodes = x[iaxis][igrp]
@@ -85,7 +89,7 @@ def reconstruct_discr_from_nodes(actx, discr, x):
         return actx.call_loopy(
                 resample_by_mat_prg(),
                 nodes=discr_nodes,
-                resampling_mat=actx.from_numpy(to_mesh_interp_matrix(grp)),
+                resampling_mat=to_mesh_interp_matrix(grp),
                 )["result"]
 
     megs = []
@@ -115,7 +119,7 @@ def advance(actx, dt, t, x, fn):
 def run(actx, *,
         ambient_dim: int = 3,
         resolution: int = None,
-        target_order: int = 8,
+        target_order: int = 4,
         tmax: float = 1.0,
         timestep: float = 1.0e-2,
         group_factory_name: str = "warp_and_blend",
@@ -256,11 +260,11 @@ if __name__ == "__main__":
     actx = PyOpenCLArrayContext(queue)
 
     from pytools import ProcessTimer
-    for _ in range(3):
+    for _ in range(1):
         with ProcessTimer() as p:
             run(actx,
                     ambient_dim=3,
-                    group_factory_name="quadrature",
+                    group_factory_name="warp_and_blend",
                     tmax=1.0,
                     timestep=1.0e-2,
                     visualize=False)
--- a/examples/moving-geometry.py
+++ b/examples/moving-geometry.py
@@ -121,7 +121,7 @@ def advance(actx, dt, t, x, fn):
 def run(actx, *,
         ambient_dim: int = 3,
         resolution: int = None,
-        target_order: int = 4,
+        target_order: int = 8,
         tmax: float = 1.0,
         timestep: float = 1.0e-2,
         group_factory_name: str = "warp_and_blend",
@@ -170,7 +170,7 @@ def run(actx, *,
                 order=mesh_order,
                 unit_nodes=unit_nodes)
     else:
-        nrounds = 3 if resolution is None else resolution
+        nrounds = 4 if resolution is None else resolution
         mesh = gen.generate_icosphere(radius,
                 uniform_refinement_rounds=nrounds,
                 order=mesh_order,
@@ -204,6 +204,20 @@ def run(actx, *,
     def source(t, x):
         discr = reconstruct_discr_from_nodes(actx, discr0, x)
         u = velocity_field(thaw(actx, discr.nodes()))
+
+        # NOTE: these are just here because this was at some point used to
+        # profile some more operators (turned out well!)
+
+        from meshmode.discretization import num_reference_derivative
+        x = thaw(actx, discr.nodes()[0])
+        gradx = sum(
+                num_reference_derivative(discr, (i,), x)
+                for i in range(discr.dim))
+        intx = sum(actx.np.sum(xi * wi) for xi, wi in zip(x, discr.quad_weights()))
+
+        assert gradx is not None
+        assert intx is not None
+
         return u
 
     # }}}
@@ -245,7 +259,7 @@ if __name__ == "__main__":
 
     run(actx,
             ambient_dim=3,
-            group_factory_name="warp_and_blend",
+            group_factory_name="quadrature",
             tmax=1.0,
             timestep=1.0e-2,
             visualize=False)
--- a/examples/moving-geometry.py
+++ b/examples/moving-geometry.py
@@ -63,20 +63,7 @@ def reconstruct_discr_from_nodes(actx, discr, x):
             """,
             name="resample_by_mat_prg")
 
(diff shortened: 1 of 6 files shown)


inducer:
cc @thomasgibson

meshmode/discretization/__init__.py
```diff
@@ -593,6 +568,61 @@ def resample_mesh_nodes(grp, iaxis):
...
+        return vec
+
+    actx = vec.array_context
+    ref_axes = tuple(ref_axes)
+
+    if not all(0 <= ref_axis < discr.dim for ref_axis in ref_axes):
+        raise ValueError("'ref_axes' exceeds discretization dimensions: "
+                f"got {ref_axes} for dimension {discr.dim}")
+
+    @memoize_in(actx, (Discretization, "reference_derivative_prg"))
```
inducer, line 599: I find it useful to use the function symbol (`num_reference_derivative` here) in the cache key, i.e. not the string, but the function object itself. That has a much better chance of being unique than a random string. (Here and elsewhere.)

meshmode/discretization/__init__.py
```diff
@@ -593,6 +568,61 @@ def resample_mesh_nodes(grp, iaxis):
...
+    def prg():
+        return make_loopy_program(
+            "{[iel,idof,j]: 0 <= iel < nelements and 0 <= idof, j < nunit_dofs}",
+            "result[iel,idof] = sum(j, diff_mat[idof, j] * vec[iel, j])",
+            name="diff")
+
+    def get_mat(grp):
```
inducer, line 606: `@keyed_memoize_in` in the array context?

meshmode/discretization/__init__.py
inducer, line 599: Maybe memoize in the array context instead? `Discretization` is pretty global, and it essentially lives forever.

meshmode/discretization/poly_element.py
```diff
@@ -89,38 +94,91 @@
 """
 
 
+# {{{ matrices
+
+@memoize(key=lambda grp: grp.discretization_key())
```
inducer, line 99: `@memoize_first_arg`? Not sure I love these global caches.

inducer:

meshmode/discretization/__init__.py
alexfikl: Agreed! That was just an unfortunate copy pasting mistake.

Should be fixed in ae1f445.

meshmode/discretization/__init__.py
alexfikl: Done in 6ff46ef.

I added a sorted `ref_axes` into the key too. That should be enough to make it unique, right?

meshmode/discretization/__init__.py
alexfikl: Not sure what you mean? This is already memoized in the array context.

meshmode/discretization/poly_element.py
alexfikl: I mostly added this to help with #123. However, on renewed profiling it turns out that running `examples/moving-geometry.py`
* Using the warp and blend nodes goes at about 5s with local caches to 3s with global caches.
* Using the quadrature nodes is about 3s regardless. This is likely because `VioreanuRokhlinSimplexQuadrature` has precomputed nodes and weights.

That's all at about 5120 elements with 230400 degrees of freedom. In 2D it seems to hover at about <1s for 8192 elements and 73728 degrees of freedom. The example takes 100 time steps in both cases.

alexfikl: Used `memoize_on_first_arg` in 50d1aaa.

meshmode/discretization/poly_element.py
```diff
@@ -89,38 +94,91 @@
 """
 
 
+# {{{ matrices
+
+@memoize_on_first_arg
+def mass_matrix(grp: InterpolatoryElementGroupBase) -> np.ndarray:
+    if not isinstance(grp, InterpolatoryElementGroupBase):
+        raise TypeError(f"cannot construct mass matrix on '{type(grp).__name__}'")
```
thomasgibson, line 102:
```suggestion
        raise NoninterpolatoryElementGroupError(f"cannot construct mass matrix on '{type(grp).__name__}'")
```

meshmode/discretization/poly_element.py
```diff
@@ -89,38 +94,91 @@
...
+    assert grp.is_orthonormal_basis()
+    return mp.mass_matrix(
+            grp.basis_obj().functions,
+            grp.unit_nodes)
+
+
+@memoize_on_first_arg
+def inverse_mass_matrix(grp: InterpolatoryElementGroupBase) -> np.ndarray:
+    if not isinstance(grp, InterpolatoryElementGroupBase):
+        raise TypeError(f"cannot construct mass matrix on '{type(grp).__name__}'")
```
thomasgibson, line 113:
```suggestion
        raise NoninterpolatoryElementGroupError(f"cannot construct mass matrix on '{type(grp).__name__}'")
```

meshmode/discretization/poly_element.py
```diff
@@ -89,38 +94,91 @@
...
+
+    return mp.inverse_mass_matrix(
+            grp.basis_obj().functions,
+            grp.unit_nodes)
+
+
+@memoize_on_first_arg
+def diff_matrices(grp: InterpolatoryElementGroupBase) -> Tuple[np.ndarray]:
+    if not isinstance(grp, InterpolatoryElementGroupBase):
+        raise TypeError(f"cannot construct diff matrices on '{type(grp).__name__}'")
```
thomasgibson, line 123:
```suggestion
        raise NoninterpolatoryElementGroupError(f"cannot construct diff matrices on '{type(grp).__name__}'")
```

meshmode/discretization/__init__.py
thomasgibson: Wouldn't this also cause it to rebuild the matrices if a subset of `ref_axes` were used? I guess that's probably a corner case. I think this is fine in the meantime

thomasgibson (commented):
Looks great! Here are some quick comments

meshmode/discretization/poly_element.py
alexfikl: Fixed all of these offline in 6f60f77, since they would have made flake8 angry.

meshmode/discretization/__init__.py
alexfikl: Yes, it definitely would. We could make this function recursive and memoize the whole thing to catch any intermediary matrices that get built, but I'm not sure it's worth it.

From what I've seen, it's mostly just called for the first derivative and some second derivatives in [pytential](https://github.com/inducer/pytential/blob/main/pytential/symbolic/primitives.py#L818-L820).

examples/moving-geometry.py
```diff
@@ -90,14 +85,13 @@ def resample_nodes(grp, igrp, iaxis):
...
-                resampling_mat=to_mesh_interp_matrix(igrp),
-                )
+                resampling_mat=actx.from_numpy(to_mesh_interp_matrix(grp)),
+                )["result"]
 
     megs = []
     for igrp, grp in enumerate(discr.groups):
         nodes = np.stack([
-            actx.to_numpy(resample_nodes(grp, igrp, iaxis))
+            actx.to_numpy(resample_nodes_to_mesh(grp, igrp, iaxis))
```
inducer, line 94: `to_numpy` is expensive, and we should avoid it. Cache the matrix in the array context, keyed on the group's discretization key.

meshmode/discretization/__init__.py
```diff
@@ -593,6 +568,64 @@ def resample_mesh_nodes(grp, iaxis):
...
+    def prg():
+        return make_loopy_program(
+            "{[iel,idof,j]: 0 <= iel < nelements and 0 <= idof, j < nunit_dofs}",
+            "result[iel,idof] = sum(j, diff_mat[idof, j] * vec[iel, j])",
+            name="diff")
+
+    @keyed_memoize_in(actx,
+            (num_reference_derivative, "num_reference_derivative_matrix"),
+            lambda grp: grp.discretization_key() + ref_axes)
+    def get_mat(grp):
```
inducer, line 609: I'd prefer if `ref_axes` were an argument. Then the dependency is explicit, and it doesn't need to be added to the key.

meshmode/discretization/poly_element.py
```diff
@@ -89,38 +94,94 @@
...
+            grp.unit_nodes)
+
+
+@memoize_on_first_arg
+def inverse_mass_matrix(grp: InterpolatoryElementGroupBase) -> np.ndarray:
+    if not isinstance(grp, InterpolatoryElementGroupBase):
+        raise NoninterpolatoryElementGroupError(
+                f"cannot construct mass matrix on '{type(grp).__name__}'")
+
+    return mp.inverse_mass_matrix(
```
inducer, line 117: Is this missing an orthonormality assertion? (Given that `mass_matrix` has one.)

inducer (commented):
Thanks! A few wrinkles, otherwise good.

examples/moving-geometry.py
alexfikl: Done in a2f10ff.

meshmode/discretization/__init__.py
alexfikl: Done in 5d9ae7e.

meshmode/discretization/poly_element.py
alexfikl: Actually removed this completely in 23fb9d4. 

I was looking through some of grudge's matrices and thinking of caching them here too, but gave up on the idea.

meshmode/discretization/__init__.py
inducer: Sorry, I was misreading this.

inducer:
LGTM. Thanks for working on this!

PR #143, opened 2021-04-02
Add tagging functionality to the array context

inducer:
This is intended to make metadata capture available generically, even if it boils down to a no-op for the PyOpenCL array context. @kaushikcfd, could you review? I'm particularly curious what you think of `tag_axis` and the `CommonSubexpression` tag.

cc @MTCam @nchristensen @alexfikl

### DIFF
--- a/meshmode/array_context.py
+++ b/meshmode/array_context.py
@@ -20,14 +20,19 @@ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 THE SOFTWARE.
 """
 
+from typing import Union, Sequence
 from functools import partial
 import numpy as np
 import loopy as lp
 from loopy.version import MOST_RECENT_LANGUAGE_VERSION
 from pytools import memoize_method
+from pytools.tag import Tag
+from abc import ABC, abstractmethod
+
 
 __doc__ = """
 .. autofunction:: make_loopy_program
+.. autoclass:: CommonSubexpression
 .. autoclass:: ArrayContext
 .. autoclass:: PyOpenCLArrayContext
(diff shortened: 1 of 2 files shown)


meshmode/array_context.py
```diff
@@ -175,11 +180,22 @@ def __init__(self, array_context):
         self._array_context = array_context
 
 
-class ArrayContext:
+class CommonSubexpression(Tag):
```
alexfikl, line 183:
```suggestion
class CommonSubexpressionTag(Tag):
```
? Mostly to differentiate it a bit from the same in `pymbolic`, but I don't feel strongly about it.

meshmode/array_context.py
```diff
@@ -213,9 +229,11 @@ def __init__(self):
     def _get_fake_numpy_namespace(self):
         return _BaseFakeNumpyNamespace(self)
 
+    @abstractmethod
     def empty(self, shape, dtype):
         raise NotImplementedError
```
alexfikl, line 234:
```suggestion
        pass
```
? I think abstract methods have their own exception when not implemented?

alexfikl (commented):
I'm still very much in the non-lazy land, so don't have much else to say here. But it looks good!

meshmode/array_context.py
alexfikl: Ah, you just can't instantiate them at all, so I guess this makes no difference.

kaushikcfd (approved):
LGTM.

meshmode/array_context.py
inducer: 51d2f20

inducer:
Thanks to the both of you for taking a look!

inducer:
If you're wondering about the CI failures: https://github.com/illinois-ceesd/mirgecom/issues/302

PR #146, opened 2021-04-04
Generalize array comparison logic

kaushikcfd:
Forcing all array contexts to support `arr1 == arr2` for elementwise comparison is quite restrictive. This PR gets rid of that restriction.


inducer:
Thanks!

PR #154, opened 2021-04-08
Flatten and unflatten by reshaping when possible

alexfikl:
Is this sort of what you hand in mind in https://github.com/inducer/pytential/pull/71#discussion_r609231956? 
It seems to do the trick, i.e. `flatten` and `unflatten` don't show up anywhere near the top of any profiling (for a single group mesh at least).

### DIFF
--- a/meshmode/dof_array.py
+++ b/meshmode/dof_array.py
@@ -499,7 +499,7 @@ def flatten(ary: Union[DOFArray, np.ndarray]) -> Any:
     # NOTE: arrays with one group are common enough and don't require any
     # concatenation to "flatten", but can just be reshaped
     if len(ary) == 1:
-        ary0 = list(ary)[0]
+        ary0, = ary
         if ary0.flags.c_contiguous:
             return ary0.reshape(-1, order="C")
         elif ary0.flags.f_contiguous:
--- a/meshmode/dof_array.py
+++ b/meshmode/dof_array.py
@@ -496,6 +496,11 @@ def flatten(ary: Union[DOFArray, np.ndarray]) -> Any:
     if isinstance(ary, np.ndarray):
         return obj_array_vectorize(flatten, ary)
 
+    # NOTE: arrays with one group are common enough and don't require any
+    # concatenation to "flatten", but can just be reshaped
+    if len(ary) == 1:
+        return list(ary)[0].reshape(-1)
+
     group_sizes = [grp_ary.shape[0] * grp_ary.shape[1] for grp_ary in ary]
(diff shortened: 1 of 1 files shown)


meshmode/dof_array.py
```diff
@@ -496,6 +496,11 @@ def flatten(ary: Union[DOFArray, np.ndarray]) -> Any:
     if isinstance(ary, np.ndarray):
         return obj_array_vectorize(flatten, ary)
 
+    # NOTE: arrays with one group are common enough and don't require any
+    # concatenation to "flatten", but can just be reshaped
+    if len(ary) == 1:
+        return list(ary)[0].reshape(-1)
```
inducer, line 502: To be safe, I think it'd be useful to verify the memory layout using the strides here.

alexfikl: I wanted to do something like that, but wasn't sure how it would interfere with the lazy stuff. Would adding an `actx.np.reshape` and reusing [this](https://github.com/inducer/meshmode/blob/cb6b6b983855c7b7e8aa52968af554005a72a833/meshmode/array_context.py#L415-L426) make more sense?

meshmode/dof_array.py
```diff
@@ -496,6 +496,18 @@ def flatten(ary: Union[DOFArray, np.ndarray]) -> Any:
     if isinstance(ary, np.ndarray):
         return obj_array_vectorize(flatten, ary)
 
+    # NOTE: arrays with one group are common enough and don't require any
+    # concatenation to "flatten", but can just be reshaped
+    if len(ary) == 1:
+        ary0 = list(ary)[0]
```
inducer, line 502:
```suggestion
        ary0, = ary
```

meshmode/dof_array.py
```diff
@@ -496,6 +496,18 @@ def flatten(ary: Union[DOFArray, np.ndarray]) -> Any:
...
+        if ary0.flags.c_contiguous:
+            return ary0.reshape(-1, order="C")
+        elif ary0.flags.f_contiguous:
+            return ary0.reshape(-1, order="F")
+        else:
+            # NOTE: array has unsupported strides, so we need to do a copy
+            pass
```
inducer, line 509: This isn't correct. The existing array can be reshaped if and only if the strides are `(itemsize*nunit_dofs_per_element, itemsize)`. (Note that subclassing array contexts might do weird things like introduce extra dimensions.)

inducer:

alexfikl:
This has been graciously continued by @thomasgibson in #192.

PR #158, opened 2021-04-14
Fix 0d quadrature rules 

inducer:
@alexfikl Could you take a look?

cc @thomasgibson

### DIFF
--- a/meshmode/discretization/poly_element.py
+++ b/meshmode/discretization/poly_element.py
@@ -236,7 +236,7 @@ class InterpolatoryQuadratureSimplexElementGroup(PolynomialSimplexElementGroupBa
     def _quadrature_rule(self):
         dims = self.mesh_el_group.dim
         if dims == 0:
-            return mp.Quadrature(np.empty((0, 1)), np.empty((0, 1)))
+            return mp.Quadrature(np.empty((0, 1)), np.ones((1,)))
         elif dims == 1:
             return mp.LegendreGaussQuadrature(self.order)
         else:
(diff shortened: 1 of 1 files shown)


alexfikl:
From what I remember, I hacked this so that grudge works in for 1D geometries with 0d faces. I'll go ahead and re-run the advection examples in 1D (by default they run in 2D) and see if anything explodes.

alexfikl:
Seems to work just fine, so I'm guessing that was just a weird copy pasting thing. Sorry!

There's also a [test case](https://github.com/inducer/grudge/blob/6ae443d2b4eceac220313fd0737c681a5ead5131/test/test_grudge.py#L645) for the 0d case that seems to be passing, so this is looks good to me!

inducer:
Thanks for taking a look, @alexfikl! Independent of whether the code runs though, does this change make sense to you? I find it strange that the weights would be a two-dimensional (as in rows and columns) array.

thomasgibson (commented):
This looks like a 0-D quadrature rule to me :smile: OOC: `nodes` here are left as `np.empty((0,1))` because they're never used (understandably)? It has the right shape that `modepy.Quadrature` expects, so I think all should be good.

inducer:
> OOC

What's that mean? I googled, and "out of context" was perhaps the best I found... but even that didn't seem to fit.

> `nodes` here are left as `np.empty((0,1))` because they're never used (understandably)?

Suppose you were doing function evaluation on this. Then there is no information content at all in the node position, given that we're, well, integrating, over a point (0d entity). So you take in an array of shape `(0, 1)` (one node in 0d) and return an array of shape `(1,)` (your function value on that one node. That's how that shape makes sense to me. HTH, feel free to follow up with further questions.

PR #161, opened 2021-04-14
Use `ZeroDimensionalQuadrature` for 0-dim reference cells

thomasgibson:
Thanks to @alexfikl for fixing this in modepy! This just carries over that change into meshmode.


inducer:
Thanks!

inducer:
I hesitated for a second there because probably we should've done a version bump on modepy, so that meshmode can depend on the new version.

inducer:
Maybe that's OK this time, but in general, I think we should do that.)

PR #164, opened 2021-04-15
Make `quadrature_rule` a public-facing method

thomasgibson:
Retires `weights` and instead exposes a `modepy.Quadrature` object associated with relevant element groups

### DIFF
--- a/meshmode/discretization/__init__.py
+++ b/meshmode/discretization/__init__.py
@@ -194,11 +194,23 @@ class NodalElementGroupBase(ElementGroupBase):
         """
         return self.unit_nodes.shape[-1]
 
-    @abstractproperty
+    @property
+    @memoize_method
     def unit_nodes(self):
         """Returns a :class:`numpy.ndarray` of shape ``(dim, nunit_dofs)``
         of reference coordinates of interpolation nodes.
+
+        Note: this method dispatches to the nodes of the underlying
+        quadrature rule. This means, for interpolatory element groups,
+        interpolation nodes are collocated with quadrature nodes.
         """
+        result = self.quadrature_rule().nodes
+        if len(result.shape) == 1:
+            result = np.array([result])
+
+        dim2, _ = result.shape
+        assert dim2 == self.mesh_el_group.dim
+        return result
 
     @abstractmethod
     def quadrature_rule(self):
--- a/meshmode/discretization/__init__.py
+++ b/meshmode/discretization/__init__.py
@@ -177,14 +177,14 @@ class NodalElementGroupBase(ElementGroupBase):
     equipped with nodes. Nodes are specific locations defined on the
     reference element (:attr:`~ElementGroupBase.shape`)
     defining a degree of freedom by point evaluation at that location.
-    Such element groups can have an associated quadrature rule to perform
+    Such element groups have an associated quadrature rule to perform
     numerical integration, but are not necessarily usable (unisolvent)
     for interpolation.
 
     Inherits from :class:`ElementGroupBase`.
 
     .. autoattribute:: unit_nodes
-    .. autoattribute:: weights
+    .. automethod:: quadrature_rule
     """
 
     @property
@@ -200,12 +200,23 @@ class NodalElementGroupBase(ElementGroupBase):
         of reference coordinates of interpolation nodes.
         """
 
-    @abstractproperty
+    @abstractmethod
+    def quadrature_rule(self):
+        """Returns a :class:`modepy.Quadrature` object for the
+        element group.
+        """
+
+    @property
     def weights(self):
         """Returns a :class:`numpy.ndarray` of shape ``(nunit_dofs,)``
         containing quadrature weights applicable on the reference
         element.
         """
+        warn("`grp.weights` will be dropped in version 2022.x "
+             "To access the quadrature weights, use "
+             "`grp.quadrature_rule().weights` instead.",
+             DeprecationWarning, stacklevel=2)
+        return self.quadrature_rule().weights
 
 # }}}
 
--- a/meshmode/discretization/__init__.py
+++ b/meshmode/discretization/__init__.py
@@ -224,8 +224,8 @@ class NodalElementGroupBase(ElementGroupBase):
         containing quadrature weights applicable on the reference
         element.
         """
-        warn("`grp.weights` will be dropped in version 2022.x "
-             "To access the quadrature weights, use "
+        warn("`grp.weights` is deprecated and will be dropped "
+             "in version 2022.x. To access the quadrature weights, use "
              "`grp.quadrature_rule().weights` instead.",
              DeprecationWarning, stacklevel=2)
         return self.quadrature_rule().weights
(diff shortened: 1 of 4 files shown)


thomasgibson:
Will be used in https://github.com/inducer/grudge/pull/62

inducer (commented):
Just one question, otherwise LGTM. Thanks for working on this!

meshmode/discretization/__init__.py
```diff
@@ -200,12 +200,23 @@ def unit_nodes(self):
         of reference coordinates of interpolation nodes.
         """
```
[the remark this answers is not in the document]
thomasgibson: I was wondering the same thing. Though I think this implicitly assumes that interpolation nodes and quadrature nodes are collocated in all the element groups. Do we want to do this or make a distinction between "interpolation" and "quadrature" nodes?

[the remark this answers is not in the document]
inducer: I think that's an OK assumption. Each element group is (at most!) going to have one set of nodes. Everything else would be very confusing.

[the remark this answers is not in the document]
thomasgibson: OK. In that case, I think I'll go with making `_nodes` a thing for the mass matrix quadrature group. I prefer to keep `unit_nodes` as is for now

meshmode/discretization/poly_element.py
```diff
@@ -278,31 +262,16 @@ def _quadrature_rule(self):
...
-    @property
     @memoize_method
-    def weights(self):
-        return np.dot(
-                mass_matrix(self),
-                np.ones(len(self.basis_obj().functions)))
+    def quadrature_rule(self):
+        basis_fcts = self.basis_obj().functions
+        nodes = self._nodes
+        mass_matrix = mp.mass_matrix(basis_fcts, nodes)
```
thomasgibson, line 271: @inducer I've opted to just create a mass matrix here (and in other similar places) rather than call `mass_matrix(self)` as above. The reason for this is that one get's a `RecursionError` due to needing `unit_nodes` in advance, but `quadrature_rule` isn't defined because it needs a mass matrix to get the weights, etc.

If this is not ideal, I could just add some `if isinstance(grp, ...)` logic in the `mass_matrix` function. You would need to do this for both `_MassMatrixQuadratureElementGroup` _and_ `TensorProductElementGroupBase`. Not sure which is preferrable

meshmode/discretization/__init__.py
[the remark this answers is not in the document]
thomasgibson: I think https://github.com/inducer/meshmode/pull/164/commits/1942a1b619e911a6e15eb6046eeafbeef41949df presents a good compromise. I've added a method `interp_nodes` used exclusively for the mass matrix element groups. I think this makes mathematical sense because that's precisely what is going into the quadrature rules; these element groups are truly using "interpolation nodes" to derive a quadrature rule from.

meshmode/discretization/poly_element.py
```diff
@@ -267,12 +267,19 @@ class _MassMatrixQuadratureElementGroup(PolynomialSimplexElementGroupBase):
...
                          np.ones(len(basis_fcts)))
         return mp.Quadrature(nodes, weights, exact_to=self.order)
 
+    @property
+    def interp_nodes(self):
+        raise NotImplementedError(
+            "Subclasses of `_MassMatrixQuadratureElementGroup` must "
+            "provide interpolation nodes to be used for quadrature."
+        )
+
```
thomasgibson: This is to prevent pylint from yelling about not having `grp.interp_nodes` defined, since it's being referenced in `grp.quadrature_rule`. I could make this an `@abstractproperty` instead if that's preferrable (`_MassMatrixQuadratureElementGroup` is still technically an abstract class due to inheritance)

meshmode/discretization/__init__.py
```diff
@@ -194,18 +194,41 @@ def nunit_dofs(self):
...
+        element group.
+        """
+
+    @property
     def weights(self):
         """Returns a :class:`numpy.ndarray` of shape ``(nunit_dofs,)``
         containing quadrature weights applicable on the reference
         element.
         """
+        warn("`grp.weights` will be dropped in version 2022.x "
```
inducer, line 227:
```suggestion
        warn("`grp.weights` is deprecated and will be dropped in version 2022.x "
```

meshmode/discretization/poly_element.py
inducer: It's a bit of inelegance, but IMO this is fine. This is memoized, so that cost is only incurred once.

meshmode/discretization/poly_element.py
```diff
@@ -278,31 +262,23 @@ def _quadrature_rule(self):
...
+    def quadrature_rule(self):
+        basis_fcts = self.basis_obj().functions
+        nodes = self.interp_nodes
+        mass_matrix = mp.mass_matrix(basis_fcts, nodes)
+        weights = np.dot(mass_matrix,
+                         np.ones(len(basis_fcts)))
+        return mp.Quadrature(nodes, weights, exact_to=self.order)
 
-class _MassMatrixQuadratureElementGroup(PolynomialSimplexElementGroupBase):
     @property
```
inducer, line 276:
```suggestion
    @abstractproperty
```
?

meshmode/discretization/poly_element.py
```diff
@@ -278,31 +262,23 @@ def _quadrature_rule(self):
...
-    @memoize_method
-    def weights(self):
-        return np.dot(
-                mass_matrix(self),
-                np.ones(len(self.basis_obj().functions)))
+    def interp_nodes(self):
```
inducer, line 277:
```suggestion
    def _interp_nodes(self):
```
(Since this is an interface internal to this class.)

inducer (commented):
Thanks! A few minor wrinkles, otherwise LGTM!

meshmode/discretization/__init__.py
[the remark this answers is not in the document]
inducer: Agreed. Thanks!

meshmode/discretization/poly_element.py
inducer: I'm OK with it as is or as an `@abstractproperty`. If the latter is awkward for some reason, no need to push it through IMO.

meshmode/discretization/__init__.py
thomasgibson: Done in https://github.com/inducer/meshmode/pull/164/commits/204390e2fae48ba49df18a3934fc98c53332e11e (and fixed up similar warnings)

meshmode/discretization/poly_element.py
thomasgibson: Done in https://github.com/inducer/meshmode/pull/164/commits/fe63f45d9df60c65b346a90648500148cf61e961 !

meshmode/discretization/poly_element.py
thomasgibson: Done in https://github.com/inducer/meshmode/pull/164/commits/fe63f45d9df60c65b346a90648500148cf61e961!

meshmode/discretization/poly_element.py
thomasgibson: Done in https://github.com/inducer/meshmode/pull/164/commits/fe63f45d9df60c65b346a90648500148cf61e961!

inducer:
Thx!

PR #166, opened 2021-04-19
Some cleanups for surfaces in meshmode.mesh.generation

alexfikl:
* Added some typing to curves and surfaces.
* Updated docs a bit to match.
* Added ability to disable node consistency checks.
* Updated some usage of `np.random` and `refine_uniformly` for the icosphere.

I also added some `*` (as in `func(arg0, *, kwarg)`) to some functions that may break backwards compatibility here.

### DIFF
--- a/meshmode/mesh/generation.py
+++ b/meshmode/mesh/generation.py
@@ -20,6 +20,7 @@ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 THE SOFTWARE.
 """
 
+from typing import Callable, Optional, Union
 
 import numpy as np
 import numpy.linalg as la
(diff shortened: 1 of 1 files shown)


meshmode/mesh/generation.py
```diff
@@ -187,22 +199,22 @@ class WobblyCircle:
...
+    def __init__(self, coeffs: np.ndarray):
         self.coeffs = coeffs
 
     @staticmethod
-    def random(ncoeffs, seed):
-        rng = np.random.mtrand.RandomState(seed)
-        coeffs = rng.rand(ncoeffs)
+    def random(ncoeffs: int, seed: int):
+        rng = np.random.default_rng(seed)
+        coeffs = rng.random(ncoeffs)
```
alexfikl, line 208: Functional change: using `default_rng` here, as the previous stuff should be deprecated now. It should still generate uniform random numbers in `[0, 1)` as before.

inducer:
LGTM. Thanks!

PR #167, opened 2021-04-21
Fix, test non-equal element counts in X rect mesh

inducer:
cc @anderson2981


meshmode/array_context.py
```diff
@@ -407,9 +407,9 @@ def stack(self, arrays, axis=0):
         import pyopencl.array as cla
         from meshmode.dof_array import obj_or_dof_array_vectorize_n_args
         return obj_or_dof_array_vectorize_n_args(
-            lambda *args: cla.stack(arrays=args, axis=axis,
-                                    queue=self._array_context.queue),
-            *arrays)
+                lambda *args: cla.stack(arrays=args, axis=axis,  # pylint: disable=no-member  # noqa: E501
```
inducer, line 410: @alexfikl Any insight? Did this ever work for anything?

PR #168, opened 2021-04-22
Add empty_like and friends to ArrayContext.np

alexfikl:
Added so that they can distribute nicely over `DOFArrays`.

Should supersede what was in #139 (at some point).

### DIFF
--- a/meshmode/array_context.py
+++ b/meshmode/array_context.py
@@ -390,7 +390,7 @@ class _PyOpenCLFakeNumpyNamespace(_BaseFakeNumpyNamespace):
 
     def ones_like(self, ary):
         def _ones_like(subary):
-            ones = self._array_context.zeros_like(subary)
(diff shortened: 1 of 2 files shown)


meshmode/array_context.py
```diff
@@ -364,6 +388,14 @@ def greater_equal(self, x, y): return self._bop(operator.ge, x, y)  # noqa: E704
     def less(self, x, y): return self._bop(operator.lt, x, y)  # noqa: E704
     def less_equal(self, x, y): return self._bop(operator.le, x, y)  # noqa: E704
 
+    def ones_like(self, ary):
+        def _ones_like(subary):
+            ones = self._array_context.zeros_like(subary)
```
inducer, line 393:
```suggestion
            ones = self._array_context.empty_like(subary)
```

inducer:
Thanks!

PR #169, opened 2021-04-22
Add einsum to the array context

inducer:
- [x] Docs
- [x] A test

cc @thomasgibson

### DIFF
--- a/meshmode/array_context.py
+++ b/meshmode/array_context.py
@@ -64,6 +64,18 @@ def make_loopy_program(domains, statements, kernel_data=None,
             lang_version=MOST_RECENT_LANGUAGE_VERSION)
 
 
+def _loopy_get_default_entrypoint(t_unit):
+    try:
+        # main and "kernel callables" branch
+        return t_unit.default_entrypoint
+    except AttributeError:
+        try:
+            return t_unit.root_kernel
+        except AttributeError:
+            raise TypeError("unable to find default entry point for loopy "
+                    "translation unit")
+
+
 # {{{ ArrayContext
 
 class _BaseFakeNumpyNamespace:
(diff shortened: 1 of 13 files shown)


thomasgibson:
I believe this PR is now finished, just needs https://github.com/inducer/loopy/pull/328

meshmode/array_context.py
```diff
@@ -368,6 +387,41 @@ def tag_axis(self, iaxis, tags: Union[Sequence[Tag], Tag], array):
...
+            This follows the usual :func:`numpy.einsum` convention.
+            Note that the explicit indicator `->` for the precise output
+            form is required.
+        :arg args: a sequence of array-like operands, whose order matches
+            the subscript labels provided by *spec*.
+        :arg arg_names: an optional iterable of string types denoting
+            the names of the *args*. If *None*, default names will be
+            generated.
+        :arg tagged: an optional iterable of :class:`pytools.tag.Tag`
+            objects for specifying kernel transformations.
```
inducer, line 413:
```suggestion
        :arg tagged: an optional sequence of :class:`pytools.tag.Tag`
            objects specifying the tags to be applied to the operation.
```
(It's not this routine's business to determine what the tags are used for.)

meshmode/array_context.py
```diff
@@ -368,6 +387,41 @@ def tag_axis(self, iaxis, tags: Union[Sequence[Tag], Tag], array):
...
+    def _get_einsum_prg(self, spec, arg_names, tagged):
+        return lp.make_einsum(
+            spec,
+            arg_names,
+            options=_DEFAULT_LOOPY_OPTIONS,
+            tags=tagged,
+        )
+
+    def einsum(self, spec, *args, arg_names=None, tagged=()):
+        """Executes a :mod:`loopy` einsum program on the arguments *args*.
```
inducer, line 400:
```suggestion
        """Computes the result of Einstein summation following the
        convention in :func:`numpy.einsum`.
```
(The important bit here is to specify semantics, not implementation details. The docs have no business stating *how* this is to be done ("via a loopy program"? "via a loopy program"... this will be false for the lazy array context.)

inducer (commented):
Two nits about documentation (which I'll apply... but please look at them). LGTM otherwise.

PR #170, opened 2021-04-26
Allow tagging of array axes in `ArrayContext.einsum`

thomasgibson:
An initial quick attempt to allow array axes tagging in `ArrayContext.einsum`.

### DIFF
--- a/meshmode/array_context.py
+++ b/meshmode/array_context.py
@@ -451,9 +451,7 @@ class ArrayContext(ABC):
             if any(not isinstance(tag, str) for tag in dim_tags):
                 raise ValueError("Dimension tags must be strings")
 
-            prg = lp.tag_array_axes(
-                prg, tagged_array_axes.keys(), ",".join(dim_tags)
-            )
+            prg = lp.tag_array_axes(prg, ary_names, ",".join(dim_tags))
 
         return self.call_loopy(
             prg, **{arg_names[i]: arg for i, arg in enumerate(args)}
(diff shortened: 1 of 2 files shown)


meshmode/array_context.py
```diff
@@ -424,13 +425,34 @@ def einsum(self, spec, *args, arg_names=None, tagged=()):
             generated.
         :arg tagged: an optional sequence of :class:`pytools.tag.Tag`
             objects specifying the tags to be applied to the operation.
+        :arg tagged_array_axes: an optional :class:`Dict` whose keys
+            correspond to specific *arg_names* with values consisting
+            of strings that parse to a
+            :class:`~loopy.kernel.array.ArrayDimImplementationTag`.
+            See :func:`~loopy.kernel.array.parse_array_dim_tags` for a
+            description of the allowed string format.
```
inducer, line 433: Anything captured at this level should be
* descriptive ("this axis describes the element number"), not prescriptive ("use this memory layout for this axis"), i.e. a subclass of `pytools.tag.Tag`
* as a result, definitely not a string.

inducer (commented):
Thanks for thinking about this. A comment about the approach below.

meshmode/array_context.py
inducer: (And, to be clear, the translation from "descriptive" to "prescriptive" should be done centrally in program transformation.

As a matter of justification: If the application layer (the bit that describes the math) gets involved in prescribing memory layouts, it takes freedom *away* from the transform layer to do what it wants. In that sense, prescriptive tags here would be a layering violation.

thomasgibson:
Closing this since loopy kernels are already taggable

PR #171, opened 2021-04-28
Put caching in Discretization.nodes() under user control

inducer:
This is supposed to help with https://github.com/inducer/grudge/issues/65. There's a piece in grudge that goes with this that turns off node caching for interior facial nodes, which tend to be unused after the face normals are computed.

cc @nchristensen


thomasgibson (commented):
LGTM. How about a short unit test to make sure caching works as intended?

inducer:
> LGTM. How about a short unit test to make sure caching works as intended?

bd7504a

And thanks for taking a look!

PR #172, opened 2021-04-28
Avoid creating a separate visualization discretization when avoidable

inducer:
Towards https://github.com/inducer/grudge/issues/65.

cc @nchristensen

@alexfikl, what do you think? (I'd request a review, but somehow I can't?)

### DIFF
--- a/meshmode/discretization/visualization.py
+++ b/meshmode/discretization/visualization.py
@@ -1068,18 +1068,11 @@ def make_visualizer(actx, discr, vis_order=None,
     from meshmode.discretization.poly_element import OrderAndTypeBasedGroupFactory
 
     vis_discr = None
-    if element_shrink_factor is None and not force_equidistant:
-        if vis_order is None:
-            vis_discr = discr
-        else:
-            if all(grp.order == vis_order for grp in discr.groups):
-                from warnings import warn
-                warn("Specified vis_order to be same as the order of the existing "
-                        "discretization. To avoid the creation of a separate "
-                        "discretization for visualization, avoid passing vis_order "
-                        "unless needed.", stacklevel=2)
-
-    if vis_discr is None:
+    if (element_shrink_factor is None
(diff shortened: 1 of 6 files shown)


meshmode/discretization/visualization.py
```diff
@@ -1067,22 +1067,35 @@ def make_visualizer(actx, discr, vis_order,
...
     from meshmode.discretization.poly_element import OrderAndTypeBasedGroupFactory
 
-    if force_equidistant:
-        from meshmode.discretization.poly_element import (
+    vis_discr = None
+    if element_shrink_factor is None and not force_equidistant:
+        if vis_order is None:
+            vis_discr = discr
+        else:
+            if all(grp.order == vis_order for grp in discr.groups):
```
alexfikl, line 1075: Does this warrant a warning? 

The visualizer can also change the element group (e.g. from `InterpolatoryQuadrature` to `PolynomialWarpAbdBlend`), so in pytential it makes sense to pass the same order.

alexfikl (approved):
Looks good to me! 

Big +1 on the `IdentityDiscretizationConnection`. I've been using an empty `ChainedDiscretizationConnection` for that, which didn't feel right.

alexfikl:
For the pytential failure, it should just need skipping non-direct connections somewhere around [here](https://github.com/inducer/meshmode/blob/1a61acc202da40fe9582a42bd07991f151d7d1da/meshmode/discretization/connection/chained.py#L212-L214)

meshmode/discretization/visualization.py
inducer: Fair point. How's f3ff137c990ab6198e299a500f9c66832ca4bf0d?

alexfikl: I can't seem to resolve this, so :+1:

inducer:
Thanks for taking a look!

PR #175, opened 2021-05-01
Add a generic ArrayContainer

alexfikl:
This was previously in #139, but opened a new clean PR without all the unrelated comments.

Current functionality:

* Added a barebones `ArrayContainer` class that just defines a serialization protocol for its components.
* Added a subclass `ArrayContainerWithArithmetic` that should allow to easily add new custom classes. `DOFArray` now steals all its arithmetic from here.
* Implemented a `map_array_container` and `multimap_array_container` that recurses through a tree of containers and applies a function to all their components.
* Implemented a `freeze` and `thaw` based on those that seem to work nicely enough.
* On the `DOFArray` side, added some `map_dof_array_container` and friends that stop recursing when they hit `DOFArrays`.

There's an example in `test_array` that shows how to extend a dataclass and automagically have it do arithmetic through all its fields and stuff like that. (Assuming the fields are `DOFArrays` or something else that knows how to do arithmetic).

Some small things that are left to do
- [x] ~~Remove usage of deprecated `dof_array.thaw` and `dof_array.freeze`~~ (will need to port to `arraycontext` anyway)
- [x] ~~Use `mapped_dof_array_container` instead of `obj_array_vectorize` in connections (elsewhere?) so that they vectorize nicely over array containers.~~ (actually used `multimapped_over_dof_arrays`)
- [x] Use `map_dof_array_container` in connections instead of `multimap_dof_array_container`

### DIFF
--- a/meshmode/array_context.py
+++ b/meshmode/array_context.py
@@ -423,6 +423,11 @@ def _get_array_container_ac_arith(ary: ArrayContainerWithArithmetic):
 
 # {{{ dataclass containers
 
+# FIXME: Document what this is good for
+class NumpyObjectArray(np.ndarray):
+    pass
+
+
 class DataclassArrayContainer(ArrayContainer):
     """An :class:`ArrayContainer` that implements serialization and
     deserialization of :func:`~dataclasses.dataclass` fields.
@@ -438,19 +443,27 @@ class DataclassArrayContainerWithArithmetic(
 
 
 @serialize_container.register(DataclassArrayContainer)
-def _serialize_dataclass_ac(ary: DataclassArrayContainer):
-    return ((var.name, getattr(ary, var.name)) for var in fields(ary))
+def _(ary: DataclassArrayContainer):
+    # FIXME: These field lists could be generated statically.
+    z = [(fld.name, getattr(ary, fld.name))
+            for fld in fields(ary)
+            if issubclass(fld.type, (ArrayContainer, NumpyObjectArray))]
+    return z
 
 
-@deserialize_container.register(DataclassArrayContainer)
-def _deserialize_dataclass_ac(template: Any, iterable: Iterable[Tuple[Any, Any]]):
-    return type(template)(**dict(iterable))
+@deserialize_container_class.register(DataclassArrayContainer)
+def _(template: DataclassArrayContainer, iterable: Iterable[Tuple[Any, Any]]):
+    kwargs = dict(iterable)
+    # FIXME: These field lists could be generated statically.
+    for fld in fields(cls):
+        if not issubclass(fld.type, (ArrayContainer, NumpyObjectArray)):
+            kwargs[fld.name] = getattr(template, fld.name)
+    return cls(**kwargs)
 
 # }}}
 
 
 # {{{ ArrayContainer traversal
-
 
 def _zip_containers(arys):
     r"""
--- a/meshmode/array_context.py
+++ b/meshmode/array_context.py
@@ -24,7 +24,7 @@ import operator
 from dataclasses import fields
 from functools import partial, singledispatch, update_wrapper
 from abc import ABC, abstractmethod
-from typing import Any, Callable, Iterable, Optional, Sequence, Tuple, Union
+from typing import Any, Callable, Iterable, Sequence, Tuple, Union
 
 import numpy as np
 
@@ -43,7 +43,6 @@ __doc__ = """
 .. autoclass:: ArrayContainer
 .. autofunction:: is_array_container
 .. autofunction:: serialize_container
-.. autofunction:: deserialize_container_class
 .. autofunction:: deserialize_container
 .. autofunction:: get_container_context
 .. autofunction:: get_container_context_recursively
@@ -56,6 +55,7 @@ __doc__ = """
 .. autofunction:: map_array_container
 .. autofunction:: multimap_array_container
 .. autofunction:: freeze
+.. autofunction:: thaw_impl
 .. autofunction:: thaw
 
 .. autoclass:: ArrayContext
@@ -105,10 +105,6 @@ def _loopy_get_default_entrypoint(t_unit):
 
 # {{{ ArrayContainer
 
-class _ArrayContextNotProvided:
-    pass
-
-
 class ArrayContainer:
     r"""A generic container for the array type supported by the
     :class:`ArrayContext`.
@@ -122,8 +118,7 @@ class ArrayContainer:
     * Serialization functionality is implemented in :func:`serialize_container`
       and :func:`deserialize_container`. This allows enumeration of the
       component arrays in a container and the construction of modified
-      containers from an iterable of those component arrays. Deserialization
-      may be registered using :func:`deserialize_container_class`.
+      containers from an iterable of those component arrays.
     * :func:`get_container_context` retrieves the :class:`ArrayContext` from
       a container, if it has one.
 
@@ -140,7 +135,7 @@ def is_array_container(ary: object):
 
 
 @is_array_container.register(ArrayContainer)
-def _(ary: ArrayContainer):
+def _is_array_container_ac(ary: ArrayContainer):
     return True
 
 
@@ -166,18 +161,8 @@ def serialize_container(ary: ArrayContainer) -> Iterable[Tuple[Any, Any]]:
 
 
 @singledispatch
-def deserialize_container_class(cls: type, template: Any,
-        iterable: Iterable[Tuple[Any, Any]], *,
-        actx: Optional["ArrayContext"] = None):
-    """Serves as the :func:`functools.singledispatch` registration target for
-    container deserialization through :func:`deserialize_container`.
-    """
-    raise NotImplementedError(cls.__name__)
-
-
-def deserialize_container(cls: type, template: Any,
-        iterable: Iterable[Tuple[Any, Any]], *,
-        actx: Optional["ArrayContext"] = _ArrayContextNotProvided):
+def deserialize_container(template: Any,
+        iterable: Iterable[Tuple[Any, Any]]):
     """Deserialize an iterable into an array container.
 
     :param template: an instance of an existing object that
@@ -189,11 +174,7 @@ def deserialize_container(cls: type, template: Any,
         container, if it requires one at all. If not provided, attempt to get
         a context from the *template*.
     """
-    if actx is _ArrayContextNotProvided:
-        actx = get_container_context(template)
-
-    return deserialize_container_class.dispatch(cls)(
-            cls, template, iterable, actx=actx)
+    raise NotImplementedError(type(template).__name__)
 
 
 @singledispatch
@@ -212,23 +193,23 @@ def get_container_context(ary: ArrayContainer):
 # {{{ object arrays as array containers
 
 @is_array_container.register(np.ndarray)
-def _(ary: np.ndarray):
+def _is_array_container_ndarray(ary: np.ndarray):
     return ary.dtype.char == "O"
 
 
 @serialize_container.register(np.ndarray)
-def _(ary: np.ndarray):
+def _serialize_container_ndarray(ary: np.ndarray):
     assert ary.dtype.char == "O"
     return np.ndenumerate(ary)
 
 
-@deserialize_container_class.register(np.ndarray)
-def _(cls: type, template: Any, iterable: Iterable[Tuple[Any, Any]], *,
-        actx: Optional["ArrayContext"] = None):
+@deserialize_container.register(np.ndarray)
+def _deserialize_container_ndarray(
+        template: Any, iterable: Iterable[Tuple[Any, Any]]):
     # disallow subclasses
-    assert cls is np.ndarray
+    assert type(template) is np.ndarray
 
-    result = cls(template.shape, dtype=object)
+    result = type(template)(template.shape, dtype=object)
     for i, subary in iterable:
         result[i] = subary
 
@@ -434,7 +415,7 @@ class ArrayContainerWithArithmetic(ArrayContainer):
 
 
 @get_container_context.register(ArrayContainerWithArithmetic)
-def _(ary: ArrayContainerWithArithmetic):
+def _get_array_container_ac_arith(ary: ArrayContainerWithArithmetic):
     return ary.array_context
 
 # }}}
@@ -457,14 +438,13 @@ class DataclassArrayContainerWithArithmetic(
 
 
 @serialize_container.register(DataclassArrayContainer)
-def _(ary: DataclassArrayContainer):
+def _serialize_dataclass_ac(ary: DataclassArrayContainer):
     return ((var.name, getattr(ary, var.name)) for var in fields(ary))
 
 
-@deserialize_container_class.register(DataclassArrayContainer)
-def _(cls, template: Any, iterable: Iterable[Tuple[Any, Any]], *,
-        actx: Optional["ArrayContext"] = None):
-    return cls(**dict(iterable))
+@deserialize_container.register(DataclassArrayContainer)
+def _deserialize_dataclass_ac(template: Any, iterable: Iterable[Tuple[Any, Any]]):
+    return type(template)(**dict(iterable))
 
 # }}}
 
@@ -492,8 +472,7 @@ def _zip_containers(arys):
         yield key, value
 
 
-def _map_array_container_with_context(f, ary, *,
-        actx=_ArrayContextNotProvided,
+def _map_array_container(f, ary, *,
         leaf_cls=None,
         recursive=False):
     """Helper for :func:`map_array_container`.
@@ -508,26 +487,21 @@ def _map_array_container_with_context(f, ary, *,
     if type(ary) is leaf_cls:  # type(ary) is never None
         return f(ary)
     elif is_array_container(ary):
-        array_context = actx
-        if array_context is _ArrayContextNotProvided:
-            array_context = get_container_context(ary)
-
         if recursive:
-            f = partial(_map_array_container_with_context,
-                    f, actx=actx, leaf_cls=leaf_cls, recursive=True)
+            f = partial(_map_array_container,
+                    f, leaf_cls=leaf_cls, recursive=True)
 
-        return deserialize_container(type(ary), ary, (
+        return deserialize_container(ary, (
                 (key, f(subary)) for key, subary in serialize_container(ary)
-                ), actx=array_context)
+                ))
     else:
         return f(ary)
 
 
 def _multimap_array_container_only_unchecked(f, *args,
-        actx=_ArrayContextNotProvided,
         leaf_cls=None,
         recursive=False):
-    r"""Version of :func:`_multimap_array_container_with_context` that assumes
+    r"""Version of :func:`_multimap_array_container` that assumes
     all *args* are :class:`ArrayContainer`\ s of the same type.
 
     No checks are performed.
@@ -537,21 +511,16 @@ def _multimap_array_container_only_unchecked(f, *args,
             or not is_array_container(template_ary)):
         return f(*args)
 
-    array_context = actx
-    if array_context is _ArrayContextNotProvided:
-        array_context = get_container_context(template_ary)
-
     if recursive:
         f = partial(_multimap_array_container_only_unchecked,
-                f, actx=actx, leaf_cls=leaf_cls, recursive=True)
+                f, leaf_cls=leaf_cls, recursive=True)
 
-    return deserialize_container(type(template_ary), template_ary, (
+    return deserialize_container(template_ary, (
         (key, f(*subarys)) for key, subarys in _zip_containers(args)
-        ), actx=array_context)
+        ))
 
 
-def _multimap_array_container_with_context(f, *args,
-        actx=_ArrayContextNotProvided,
+def _multimap_array_container(f, *args,
         leaf_cls=None,
         recursive=False):
     """Helper for :func:`multimap_array_container`.
@@ -580,15 +549,11 @@ def _multimap_array_container_with_context(f, *args,
 
     if len(container_indices) == len(args):
         return _multimap_array_container_only_unchecked(f, *args,
-                actx=actx, leaf_cls=leaf_cls, recursive=recursive)
-
-    array_context = actx
-    if array_context is _ArrayContextNotProvided:
-        array_context = get_container_context(template_ary)
+                leaf_cls=leaf_cls, recursive=recursive)
 
     if recursive:
-        f = partial(_multimap_array_container_with_context,
-                f, actx=actx, leaf_cls=leaf_cls, recursive=True)
+        f = partial(_multimap_array_container,
+                f, leaf_cls=leaf_cls, recursive=True)
 
     result = []
     new_args = list(args)
@@ -599,9 +564,7 @@ def _multimap_array_container_with_context(f, *args,
 
         result.append((key, f(*new_args)))
 
-    return deserialize_container(
-            type(template_ary), template_ary, tuple(result),
-            actx=array_context)
+    return deserialize_container(template_ary, tuple(result))
 
 
 def array_container_vectorize(f: Callable[[Any], Any], ary):
@@ -615,7 +578,7 @@ def array_container_vectorize(f: Callable[[Any], Any], ary):
     :param ary: a (potentially nested) structure of :class:`ArrayContainer`\ s,
         or an instance of a base array type.
     """
-    return _map_array_container_with_context(f, ary, recursive=False)
+    return _map_array_container(f, ary, recursive=False)
 
 
 def array_container_vectorize_n_args(f: Callable[[Any], Any], *args):
@@ -630,7 +593,7 @@ def array_container_vectorize_n_args(f: Callable[[Any], Any], *args):
     :param args: all :class:`ArrayContainer` arguments must be of the same
         type and with the same structure (same number of components, etc.).
     """
-    return _multimap_array_container_with_context(f, *args, recursive=False)
+    return _multimap_array_container(f, *args, recursive=False)
 
 
 def map_array_container(f: Callable[[Any], Any], ary):
@@ -641,7 +604,7 @@ def map_array_container(f: Callable[[Any], Any], ary):
     :param ary: a (potentially nested) structure of :class:`ArrayContainer`\ s,
         or an instance of a base array type.
     """
-    return _map_array_container_with_context(f, ary, recursive=True)
+    return _map_array_container(f, ary, recursive=True)
 
 
 def mapped_over_array_containers(f: Callable[[Any], Any]):
@@ -659,7 +622,7 @@ def multimap_array_container(f: Callable[[Any], Any], *args):
     :param args: all :class:`ArrayContainer` arguments must be of the same
         type and with the same structure (same number of components, etc.).
     """
-    return _multimap_array_container_with_context(f, *args, recursive=True)
+    return _multimap_array_container(f, *args, recursive=True)
 
 
 def multimapped_over_array_containers(f: Callable[[Any], Any]):
@@ -671,29 +634,49 @@ def multimapped_over_array_containers(f: Callable[[Any], Any]):
     return wrapper
 
 
-def freeze(ary):
+@singledispatch
+def freeze(ary, actx=None):
     r"""Freezes recursively by going through all components of the
     :class:`ArrayContainer` *ary*.
 
     :param ary: a :meth:`~ArrayContext.thaw`\ ed :class:`ArrayContainer`.
+
+    Array container types may use :func:`functools.singledispatch` ``.register`` to
+    register additional implementations.
+
+    See :meth:`ArrayContext.thaw`.
     """
-    def _freeze(subary, actx=None):
-        if is_array_container(subary):
-            if actx is None:
-                actx = get_container_context(subary)
-
-            return _map_array_container_with_context(
-                    partial(_freeze, actx=actx),
-                    subary, actx=None, recursive=False)
+    if is_array_container(ary):
+        return _map_array_container(
+                partial(freeze, actx=actx), ary,
+                recursive=False)
+    else:
+        if actx is None:
+            raise TypeError(
+                    f"cannot freeze arrays of type {type(ary).__name__} "
+                    "when actx is not supplied. Try calling actx.freeze "
+                    "directly or supplying an array context")
         else:
-            return actx.freeze(subary)
+            return actx.freeze(ary)
 
-    if not is_array_container(ary):
-        raise TypeError(
-                f"cannot freeze arrays of type {type(ary).__name__}; "
-                "try calling 'ArrayContext.freeze' directly")
 
-    return _freeze(ary)
+@singledispatch
+def thaw_impl(ary, actx):
+    """Serves as the registration point (using :func:`functools.singledispatch`
+    ``.register`` to register additional implementations for :func:`thaw`.
+
+    .. note::
+
+        This is separate from :func:`thaw` because of argument order. Use of
+        :func:`functools.singledispatch` requires the 'dispatching' argument
+        to come first.
+    """
+    if is_array_container(ary):
+        return deserialize_container(ary,
+                ((key, thaw_impl(subary, actx))
+                    for key, subary in serialize_container(ary)))
+    else:
+        return actx.thaw(ary)
 
 
 def thaw(actx, ary):
@@ -701,14 +684,13 @@ def thaw(actx, ary):
     :class:`ArrayContainer` *ary*.
 
     :param ary: a :meth:`~ArrayContext.freeze`\ ed :class:`ArrayContainer`.
-    """
-    if not is_array_container(ary):
-        raise TypeError(
-                f"cannot thaw arrays of type {type(ary).__name__}; "
-                "try calling 'ArrayContext.thaw' directly")
 
-    return _map_array_container_with_context(
-            actx.thaw, ary, actx=actx, recursive=True)
+    Array container types may use :func:`functools.singledispatch` ``.register``
+    (with :func:`thaw_impl`) to register additional implementations.
+
+    See :meth:`ArrayContext.thaw`.
+    """
+    return thaw_impl(ary, actx)
 
 # }}}
 
--- a/meshmode/array_context.py
+++ b/meshmode/array_context.py
@@ -166,25 +166,32 @@ def serialize_container(ary: ArrayContainer) -> Iterable[Tuple[Any, Any]]:
 
 @singledispatch
 def deserialize_container_class(cls: type,
-        actx: Optional["ArrayContext"],
-        iterable: Iterable[Tuple[Any, Any]]):
+        iterable: Iterable[Tuple[Any, Any]], *,
+        actx: Optional["ArrayContext"] = None,
+        template: Optional[Any] = None):
     """Serves as the :func:`functools.singledispatch` registration target for
-    container serialization.
+    container deserialization through :func:`deserialize_container`.
     """
+
     raise NotImplementedError(cls.__name__)
 
 
 def deserialize_container(cls: type,
-        actx: Optional["ArrayContext"],
-        iterable: Iterable[Tuple[Any, Any]]):
+        iterable: Iterable[Tuple[Any, Any]], *,
+        actx: Optional["ArrayContext"] = None,
+        template: Optional[Any] = None):
     """Deserialize an iterable into an array container.
 
-    :param actx: can be *None* for frozen arrays or if the array does not use
-        an :class:`ArrayContext`.
     :param iterable: an iterable that mirrors the output of
         :meth:`serialize_container`.
+    :param actx: can be *None* for frozen arrays or if the array does not use
+        an :class:`ArrayContext`.
+    :param template: an instance of an existing object of class *cls* that
+        can be used to aid in the deserialization. For a similar choice
+        see :meth:`numpy.ndarray.__array_finalize__`.
     """
-    return deserialize_container_class.dispatch(cls)(cls, actx, iterable)
+    return deserialize_container_class.dispatch(cls)(
+            cls, iterable, actx=actx, template=template)
 
 
 @singledispatch
@@ -215,7 +222,9 @@ def _(ary: Union[list, tuple, np.ndarray]):
 
 
 @deserialize_container_class.register(np.ndarray)
-def _(cls: type, actx: "ArrayContext", iterable):
+def _(cls: type, iterable: Iterable[Tuple[Any, Any]], *,
+        actx: Optional["ArrayContext"] = None,
+        template: Optional[Any] = None):
     # disallow subclasses
     assert cls is np.ndarray
     iterable = list(iterable)
@@ -454,7 +463,9 @@ def _(ary: DataclassArrayContainer):
 
 
 @deserialize_container_class.register(DataclassArrayContainer)
-def _(cls, actx, iterable):
+def _(cls, iterable: Iterable[Tuple[Any, Any]], *,
+        actx: Optional["ArrayContext"] = None,
+        template: Optional[Any] = None):
     return cls(**dict(iterable))
 
 # }}}
@@ -492,9 +503,9 @@ def _map_array_container_with_context(f, ary, *,
             f = partial(_map_array_container_with_context,
                     f, actx=actx, scalar_cls=scalar_cls)
 
-        return deserialize_container(type(ary), ary_actx, (
+        return deserialize_container(type(ary), (
                 (key, f(subary)) for key, subary in serialize_container(ary)
-                ))
+                ), actx=actx, template=ary)
     else:
         return f(ary)
 
@@ -546,7 +557,9 @@ def _multimap_array_container_with_context(f, *args,
 
         result.append((key, f(*new_args)))
 
-    return deserialize_container(type(template_ary), ary_actx, tuple(result))
+    return deserialize_container(
+            type(template_ary), tuple(result),
+            actx=ary_actx, template=template_ary)
 
 
 def array_container_vectorize(f: Callable[[Any], Any], ary):
--- a/meshmode/array_context.py
+++ b/meshmode/array_context.py
@@ -109,8 +109,9 @@ class ArrayContainer:
     r"""A generic container for the array type supported by the
     :class:`ArrayContext`.
 
-    The functionality for the container is implemented through
-    :func:`functools.singledispatch`. The following functions are required
+    The functionality required for the container to operated is supplied via
+    :func:`functools.singledispatch`. Implementations of the following functions need
+    to be registered for a type serving as an :class:`ArrayContainer`:
 
     * :func:`serialize_container` for serialization, which gives the components
       of the array.
@@ -360,9 +361,9 @@ def with_container_arithmetic(
     :arg bcast_number: If *True*, numbers broadcast over the container
         (with the container as the 'outer' structure).
     :arg bcast_obj_array: If *True*, :mod:`numpy` object arrays broadcast over
-        the container.  (with the container as the 'inner' structure).
+        the container.  (with the container as the 'inner' structure)
     :arg bcast_numpy_array: If *True*, any :class:`numpy.ndarray` will broadcast
-        over the container.  (with the container as the 'inner' structure).
+        over the container.  (with the container as the 'inner' structure)
     :arg arithmetic: Implement the conventional arithmetic operators, including
         ``**``, :func:`divmod`, and ``//``. Also includes ``+`` and ``-`` as well as
         :func:`abs`.
@@ -373,7 +374,7 @@ def with_container_arithmetic(
         In that case, if *eq_comparison* is unspecified, it is also set to
         *True*.
 
-    Each operator class also includes the "reverse" operator class if applicable.
+    Each operator class also includes the "reverse" operators if applicable.
 
     .. note::
 
@@ -575,7 +576,7 @@ def dataclass_array_container(cls):
 
     Attributes that are not array containers are allowed. In order to decide
     whether an attribute is an array container, the declared attribute type
-    is checked for whether it is a subclass of :class:`ArrayContainer`.
+    is checked via :func:`is_array_container_type`.
     """
     from dataclasses import is_dataclass
     assert is_dataclass(cls)
@@ -792,6 +793,8 @@ def rec_multimap_array_container(f: Callable[[Any], Any], *args):
 
 def multimapped_over_array_containers(f: Callable[[Any], Any]):
     """Decorator around :func:`rec_multimap_array_container`."""
+    # can't use functools.partial, because its result is insufficiently
+    # function-y to be used as a method definition.
     def wrapper(*args):
         return rec_multimap_array_container(f, *args)
 
--- a/meshmode/array_context.py
+++ b/meshmode/array_context.py
@@ -816,9 +816,7 @@ def freeze(ary, actx=None):
     See :meth:`ArrayContext.thaw`.
     """
     if is_array_container(ary):
-        return _map_array_container_impl(
-                partial(freeze, actx=actx), ary,
-                recursive=False)
+        return map_array_container(partial(freeze, actx=actx), ary)
     else:
         if actx is None:
             raise TypeError(
--- a/meshmode/array_context.py
+++ b/meshmode/array_context.py
@@ -565,7 +565,7 @@ def _multimap_array_container_with_context(f, *args,
     container_indices = [
             i for i, arg in enumerate(args)
             if isinstance(arg, ArrayContainer)
-            or (leaf_cls is not None and type(arg) is not leaf_cls)
+            and (leaf_cls is None or type(arg) is not leaf_cls)
             ]
 
     if not container_indices:
--- a/meshmode/array_context.py
+++ b/meshmode/array_context.py
@@ -509,18 +509,18 @@ def _zip_containers(arys):
 
 def _map_array_container_with_context(f, ary, *,
         actx=_ArrayContextNotProvided,
-        scalar_cls=None,
+        leaf_cls=None,
         recursive=False):
     """Helper for :func:`map_array_container`.
 
     :param actx: :class:`ArrayContext` passed in to :func:`deserialize_container`.
         If not provided, the context of the serialized container is used, if any.
-    :param scalar_cls: class on which we call *f* directly. This is mostly
+    :param leaf_cls: class on which we call *f* directly. This is mostly
         useful in the recursive setting, where it can stop the recursion on
         specific container classes. By default, the recursion is stopped when
         a non-:class:`ArrayContainer` class is encountered.
     """
-    if scalar_cls is not None and type(ary) is scalar_cls:
+    if leaf_cls is not None and type(ary) is leaf_cls:
         return f(ary)
     elif isinstance(ary, ArrayContainer):
         array_context = actx
@@ -529,7 +529,7 @@ def _map_array_container_with_context(f, ary, *,
 
         if recursive:
             f = partial(_map_array_container_with_context,
-                    f, actx=actx, scalar_cls=scalar_cls, recursive=True)
+                    f, actx=actx, leaf_cls=leaf_cls, recursive=True)
 
         return deserialize_container(type(ary), (
                 (key, f(subary)) for key, subary in serialize_container(ary)
@@ -540,7 +540,7 @@ def _map_array_container_with_context(f, ary, *,
 
 def _multimap_array_container_only_unchecked(f, *args,
         actx=_ArrayContextNotProvided,
-        scalar_cls=None,
+        leaf_cls=None,
         recursive=False):
     r"""Version of :func:`_multimap_array_container_with_context` that assumes
     all *args* are :class:`ArrayContainer`\ s of the same type.
@@ -548,7 +548,7 @@ def _multimap_array_container_only_unchecked(f, *args,
     No checks are performed.
     """
     template_ary = args[0]
-    if ((scalar_cls is not None and type(template_ary) is scalar_cls)
+    if ((leaf_cls is not None and type(template_ary) is leaf_cls)
             or not isinstance(template_ary, ArrayContainer)):
         return f(*args)
 
@@ -558,7 +558,7 @@ def _multimap_array_container_only_unchecked(f, *args,
 
     if recursive:
         f = partial(_multimap_array_container_only_unchecked,
-                f, actx=actx, scalar_cls=scalar_cls, recursive=True)
+                f, actx=actx, leaf_cls=leaf_cls, recursive=True)
 
     return deserialize_container(type(template_ary), (
         (key, f(*subarys)) for key, subarys in _zip_containers(args)
@@ -567,13 +567,13 @@ def _multimap_array_container_only_unchecked(f, *args,
 
 def _multimap_array_container_with_context(f, *args,
         actx=_ArrayContextNotProvided,
-        scalar_cls=None,
+        leaf_cls=None,
         recursive=False):
     """Helper for :func:`multimap_array_container`.
 
     :param actx: :class:`ArrayContext` passed in to :func:`deserialize_container`.
         If not provided, the context of the serialized container is used, if any.
-    :param scalar_cls: class on which we call *f* directly. This is mostly
+    :param leaf_cls: class on which we call *f* directly. This is mostly
         useful in the recursive setting, where it can stop the recursion on
         specific container classes. By default, the recursion is stopped when
         a non-:class:`ArrayContainer` class is encountered.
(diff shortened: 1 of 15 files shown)


meshmode/array_context.py
```diff
@@ -76,6 +79,126 @@ def _loopy_get_default_entrypoint(t_unit):
...
+    elif isinstance(ary, np.ndarray) and ary.dtype.char == "O":
+        from pytools import single_valued
+        return single_valued([
+            get_array_container_context(subary) for subary in ary
+            ], equality_pred=operator.is_)
+    else:
+        raise TypeError(f"unsupported type: {type(ary).__name__}")
+
+
+def array_container_vectorize(f: Callable[[Any], Any], ary):
```
alexfikl, line 132: I'm not super fond of this name if it recurses into the tree. `array_container_map`?

thomasgibson: `array_container_map` makes sense to me!

meshmode/dof_array.py
```diff
@@ -170,6 +169,24 @@ def __str__(self):
...
+    @classmethod
+    def from_iterable(cls, actx, iterable):
+        iterable = list(iterable)
+
+        result = [None] * len(iterable)
+        for i, subary in iterable:
+            result[i] = subary
+
+        assert all(subary is not None for subary in result)
+        return DOFArray(actx, tuple(result))
```
thomasgibson, line 186:
```suggestion
    return DOFArray(actx, data=tuple(subary for _, subary in iterable))
```
? (untested but just a thought)

meshmode/dof_array.py
```diff
@@ -170,6 +169,24 @@ def __str__(self):
     def __repr__(self):
         return f"DOFArray({repr(self._data)})"
 
+    # {{{ array container protocol
+
+    def as_iterable(self):
+        return ((i, subary) for i, subary in enumerate(self._data))
```
thomasgibson, line 175: Just a question: Do you envision the use of other identifiers as potential keys, or do you just want to ensure a strict ordering? If it's just ordering, then wouldn't a `List` be sufficient? Unless we want to also care about ordering + immutability, then a `Tuple` is probably better

thomasgibson (commented):
This looks awesome! Just some small remarks

meshmode/dof_array.py
alexfikl: This is meant to be as general as possible so that we can sort of put anything in there and it can be as nested as it needs to. 

For example, in the tests there's a dataclass where the keys are the field names because there's no integer index for them.

alexfikl: The general inspiration is [pytrees in jax](https://jax.readthedocs.io/en/latest/pytrees.html), so maybe that's a better explanation.

meshmode/array_context.py
```diff
@@ -76,6 +79,126 @@ def _loopy_get_default_entrypoint(t_unit):
...
+
+    .. automethod:: as_iterable
+    .. automethod:: from_iterable
+    """
+
+    def __init__(self, actx):
+        self.array_context = actx
+
+    @abstractmethod
+    def as_iterable(self) -> Iterable[Tuple[Any, Any]]:
```
inducer, line 100: `as_arrayctnr_iterable`? `as_array_iterable`? (analogously for `from_...`) I ask because this interface shares its namespace with subclasses, and there may be more than one way to convert the subclass to an iterable.

meshmode/array_context.py
```diff
@@ -76,6 +79,126 @@ def _loopy_get_default_entrypoint(t_unit):
...
+    .. attribute:: array_context
+
+        An :class:`meshmode.array_context.ArrayContext`.
+
+    .. automethod:: as_iterable
+    .. automethod:: from_iterable
+    """
+
+    def __init__(self, actx):
+        self.array_context = actx
```
inducer, line 97: Maybe define that this attribute must be available, rather than forcing it to be stored explicitly? (Don't know... could go both ways on that.)

meshmode/array_context.py
inducer: `map_array_container`? (Or just `map`? We're implicitly qualified by the module name after all.)

meshmode/array_context.py
```diff
@@ -76,6 +79,126 @@ def _loopy_get_default_entrypoint(t_unit):
...
+
+def array_container_vectorize(f: Callable[[Any], Any], ary):
+    r"""Applies *f* recursively over all :class:`ArrayContainer`\ s and object
+    arrays.
+
+    Works similarly to :func:`~pytools.obj_array.obj_array_vectorize`, but
+    recurses into all :class:`ArrayContainer` classes and applies *f* to their
+    components as well.
+
+    :param ary: a tree-like structures of :meth:`~ArrayContext.thaw`\ ed
```
inducer, line 140: Why require `thaw`ed-ness?

meshmode/array_context.py
```diff
@@ -76,6 +79,126 @@ def _loopy_get_default_entrypoint(t_unit):
...
+
+        return ary.from_iterable(None, (
+            (key, freeze(subary, actx=ary.array_context))
+            for key, subary in ary.as_iterable()
+            ))
+    elif isinstance(ary, np.ndarray) and ary.dtype.char == "O":
+        from pytools.obj_array import obj_array_vectorize
+        return obj_array_vectorize(partial(freeze, actx=actx), ary)
+    else:
+        return actx.freeze(ary)
```
inducer, line 175: Why not implement these in terms of `map_array_container`?

meshmode/array_context.py
```diff
@@ -76,6 +79,126 @@ def _loopy_get_default_entrypoint(t_unit):
                     "translation unit")
 
 
+# {{{ ArrayContainer
+
+class ArrayContainer(ABC):
```
inducer, line 84: Should this have a `__subclasshook__` to let it recognize numpy object arrays?

meshmode/array_context.py
```diff
@@ -76,6 +79,126 @@ def _loopy_get_default_entrypoint(t_unit):
...
+    """A generic container for the array type supported by the
+    :attr:`array_context`.
+
+    .. attribute:: array_context
+
+        An :class:`meshmode.array_context.ArrayContext`.
```
inducer, line 90: I wonder how "foreign" of data types we're wanting to allow here (see the `__subclasshook__` comment above). This would be a lot to ask of those.

meshmode/array_context.py
```diff
@@ -76,6 +79,126 @@ def _loopy_get_default_entrypoint(t_unit):
...
+    def __init__(self, actx):
+        self.array_context = actx
+
+    @abstractmethod
+    def as_iterable(self) -> Iterable[Tuple[Any, Any]]:
+        """Serialize the array container into an iterable over its components.
+
+        :returns: an :class:`Iterable` of 2-tuples where the first
+            entry is an identifier for the component and the second entry
+            is an array-like component of the :class:`ArrayContainer`.
```
inducer, line 105: Maybe clarify that keys and ordering of the iterable are entirely under the control of the subclass.

meshmode/array_context.py
inducer, line 100: If we go with accepting all sorts of things (numpy object arrays?) as subclasses of this, then this and `from_iterable` would need to become `singledispatch`, where `from_iterable` would need to become `from_iterable_like`, with a "sample isntance" instance passed as the first argument.

inducer (commented):
Thanks for working on this! Some thoughts (including a few speculative ones) below.

meshmode/array_context.py
alexfikl: Renamed it to `map_array_container` for now, seemed clearer.
60c5756

meshmode/array_context.py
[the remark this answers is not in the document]
alexfikl: Added in 60c5756.

meshmode/array_context.py
alexfikl: Yeah, removed that in 60c5756, not sure what I had in mind at the time.

meshmode/array_context.py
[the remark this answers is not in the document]
alexfikl: Added a blurb in 9e2b850.

meshmode/array_context.py
alexfikl: Yeah, that sounds nicer. At least for the dataclass in the tests, it removes the need for a `__post_init__`.
9e2b850

meshmode/array_context.py
alexfikl: Added a sentence in 561fa46.

meshmode/array_context.py
```diff
@@ -76,6 +79,126 @@ def _loopy_get_default_entrypoint(t_unit):
...
+    @classmethod
+    @abstractmethod
+    def from_iterable(cls,
+            actx: "ArrayContext",
+            iterable: Iterable[Tuple[Any, Any]]):
+        """Convert an iterable into an array container.
+
+        :param iterable: an iterable that mirrors the output of
+            :meth:`as_iterable`.
+        """
```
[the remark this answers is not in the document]
alexfikl: Added in 01e40d2

meshmode/array_context.py
alexfikl: In the case of freeze, because
* I couldn't figure out how to nicely do that check that it's not already frozen
* `map_array_container` uses the `array_context` in the container when calling `from_iterable` instead of `None`.

I'm open to suggestion on how to hack around those?

meshmode/array_context.py
alexfikl: Isn't that what `__iter__` would be for?

Maybe go with a more pickle friendly name: `serialize_container` and `deserialize_container`? Possibly even without the `container` part?

meshmode/dof_array.py
alexfikl: The original code was supposed to do a bit of checking that it's not being passed nonsense. I'll make that a bit more explicit, but I don't think I'd like getting rid of it completely.

This interface isn't particularly meant to be used by the outside world, so I could be convinced to get rid of it?

meshmode/array_context.py
alexfikl: Went with something like this in 9a2ec50.

alexfikl:
@inducer I imagine this still has quite a few rough edges, but it should be in a more usable state now.

meshmode/array_context.py
```diff
@@ -76,6 +92,433 @@ def _loopy_get_default_entrypoint(t_unit):
...
+
+
+def map_array_container(f: Callable[[Any], Any], ary):
+    r"""Applies *f* recursively to an :class:`ArrayContainer`.
+
+    Works similarly to :func:`~pytools.obj_array.obj_array_vectorize`, but
+    recurses into all :class:`ArrayContainer` classes and applies *f* only
+    at the leaf level.
+
+    :param ary: a tree-like (nested) structure of :class:`ArrayContainer`\ s.
```
inducer, line 439:
```suggestion
    :param ary: a (potentially nested) structure of :class:`ArrayContainer`\ s, or
        an instance of a base array type.
```

meshmode/array_context.py
```diff
@@ -76,6 +92,433 @@ def _loopy_get_default_entrypoint(t_unit):
...
+    """
+    actx = get_container_context(ary)
+    return _map_array_container_with_leaf_class(actx, f, ary)
+
+
+def mapped_array_container(f: Callable[[Any], Any]):
```
inducer, line 445:
```suggestion
def mapped_over_array_containers(f: Callable[[Any], Any]):
```
?

meshmode/array_context.py
```diff
@@ -76,6 +92,433 @@ def _loopy_get_default_entrypoint(t_unit):
...
+# 'ArrayContainers' instead of overwriting some `__instancecheck__`.
+
+class ArrayContainer(ABC):
+    """A generic container for the array type supported by the
+    :class:`ArrayContext`.
+
+    Serialization functionality is implemented in :func:`serialize_container`
+    and :func:`deserialize_container` using the :func:`functools.singledispatch`
+    mechanism.
+    """
```
inducer, line 107: Maybe add a nod towards Jax's array trees for a related interface?

meshmode/array_context.py
```diff
@@ -76,6 +92,433 @@ def _loopy_get_default_entrypoint(t_unit):
...
+def thaw(actx, ary):
+    r"""Thaws recursively by going through all components of
+    :class:`ArrayContainer`\ s and object arrays.
+
+    :param ary: a tree-like (nested) structure of :meth:`~ArrayContext.freeze`\ ed
+        :class:`ArrayContainer`\ s.
+    """
+    array_context = get_container_context(ary)
+    if array_context is not None:
+        raise ValueError("ArrayContainer passed to 'thaw' is not frozen")
```
inducer, line 515: I think this ultimately doubles the traversal cost, for an (IMO) fairly negligible safety benefit. (as above, just pointed out for completeness)

meshmode/array_context.py
```diff
@@ -76,6 +92,433 @@ def _loopy_get_default_entrypoint(t_unit):
...
+def freeze(ary):
+    r"""Freezes recursively by going through all components of
+    :class:`ArrayContainer`\ s and object arrays.
+
+    :param ary: a tree-like (nested) structure of :meth:`~ArrayContext.thaw`\ ed
+        :class:`ArrayContainer`\ s.
+    """
+    actx = get_container_context(ary)
+    if actx is None:
+        raise ValueError("ArrayContainer passed to 'freeze' is already frozen")
```
inducer, line 501: I think this ultimately doubles the traversal cost, for an (IMO) fairly negligible safety benefit.

meshmode/array_context.py
```diff
@@ -177,22 +620,20 @@ def loopy_implemented_elwise_func(*args):
...
         from numbers import Number
 
-        if isinstance(ary, DOFArray):
-            return DOFArray(self._array_context, tuple([
-                alloc_like(subary) for subary in ary
-                ]))
-        elif isinstance(ary, np.ndarray) and ary.dtype.char == "O":
+        if isinstance(ary, np.ndarray) and ary.dtype.char == "O":
+            # NOTE: we don't want to match numpy semantics on object arrays,
+            # e.g. `np.zeros_like(x)` returns `array([0, 0, ...], dtype=object)`
```
inducer, line 632: What's wrong with just falling through to `map_array_container` below?

meshmode/array_context.py
```diff
@@ -548,6 +991,7 @@ def norm(self, array, ord=None):
 
         # FIXME: Handling DOFArrays here is not beautiful, but it sure does avoid
         # downstream headaches.
+        # FIXME: Extend this to ArrayContainers somehow?
```
inducer, line 994: Yeah, I think doing this via `serialize_container` is reasonable. All those containers are "mathematically vector-like" IMO.

meshmode/dof_array.py
```diff
@@ -357,149 +282,88 @@ def __setstate__(self, state):
...
+    actx = get_container_context(ary)
+
+    from meshmode.array_context import _map_array_container_with_leaf_class
+    return _map_array_container_with_leaf_class(
+            actx, f, ary, leaf_class=DOFArray)
 
-    def wrapper(*args):
-        return obj_or_dof_array_vectorize_n_args(f, *args)
 
+def mapped_dof_array_container(f):
```
inducer, line 319:
```suggestion
def mapped_over_dof_arrays(f):
```

meshmode/dof_array.py
```diff
@@ -357,149 +282,88 @@ def __setstate__(self, state):
...
+    wrapper = partial(map_dof_array_container, f)
     update_wrapper(wrapper, f)
     return wrapper
 
-# }}}
-
 
-# {{{ thaw / freeze
+def multimap_dof_array_container(f: Callable[[Any], Any], *args):
```
inducer, line 325:
```suggestion
def multimapped_over_dof_arrays(f: Callable[[Any], Any], *args):
```

meshmode/dof_array.py
```diff
@@ -357,149 +282,88 @@ def __setstate__(self, state):
 # }}}
 
 
-# {{{ obj_or_dof vectorization
+# {{{ ArrayContainer implementation
 
-def obj_or_dof_array_vectorize(f, ary):
```
inducer, line 362: Any deprecation period for these? Or are we just ripping them out?

meshmode/dof_array.py
```diff
@@ -357,149 +282,88 @@ def __setstate__(self, state):
...
-    if isinstance(ary, np.ndarray):
-        return obj_array_vectorize(flatten, ary)
 
-    group_sizes = [grp_ary.shape[0] * grp_ary.shape[1] for grp_ary in ary]
-    group_starts = np.cumsum([0] + group_sizes)
-
-    actx = ary.array_context
+    actx = get_container_context(ary)
+    if actx is None:
+        raise ValueError("cannot flatten frozen ArrayContainers")
```
inducer, line 366: This gives me pause a bit, because we ultimately have to traverse the hierarchy twice: once to get the array context, and another time to flatten. I think what we're short here is a "shallow" (non-recursing) interface that says, "hey you: do you know our array context"? That way, we can pick up the the array context as we traverse (and pass it down).

meshmode/array_context.py
```diff
@@ -76,6 +92,433 @@ def _loopy_get_default_entrypoint(t_unit):
...
+    :param actx: can be *None* for frozen arrays or if the array does not use
+        an :class:`ArrayContext`.
+    :param iterable: an iterable that mirrors the output of
+        :meth:`serialize_container`.
+    """
+    return deserialize_container_class.dispatch(cls)(cls, actx, iterable)
+
+
+@singledispatch
+def get_container_context(ary: ArrayContainer):
```
inducer, line 147: I think the basic version of this shouldn't recurse. (But there can be a recursive wrapper for folks who don't mind the expense.) See also below under `flatten`.

test/test_array.py
```diff
@@ -303,6 +303,70 @@ def test_array_context_einsum_array_tripleprod(actx_factory, spec):
...
+        def array_context(self):
+            return self.mass.array_context
+
+    @serialize_container.register(DataclassArrayContainer)
+    def _(ary: DataclassArrayContainer):
+        return ((var.name, getattr(ary, var.name)) for var in fields(ary))
+
+    @deserialize_container_class.register(DataclassArrayContainer)
+    def _(cls, actx, iterable):
+        return DataclassArrayContainer(**dict(iterable))
```
inducer, line 342: Bring a base class like this out into the interface?

inducer (commented):
Thanks for working on this! I like where this is going. Some more thoughts below.

meshmode/array_context.py
alexfikl: How about defining `map_array_container(f, ary=None)` and if it's not given an `ary` return a wrapped thing that can work as a decorator? Sort of a poor man's currying?

meshmode/array_context.py
alexfikl: Basically, for an object array like `x = np.array([obj1, obj2, obj3])`, where the objects are array containers,
* `np.zeros_like(x)` will return `array([0, 0, 0], dtype=object)`
* `actx.np.zeros_like(x)` will recurse into the structure and fill it with zeros

I'm definitely ok with doing the second, since it's mostly likely what you'd want when calling that function, but it doesn't match numpy semantics, so I left it out. It would still recurse when the object array is nested somewhere inside an `ArrayContainer` class though.

Thoughts?

meshmode/array_context.py
alexfikl: Does that work recursively or would it need to flatten the serialization?

meshmode/dof_array.py
alexfikl: They should be nicely deprecated at the end of the file. They're wrapped in a `MovedFunctionDeprecationWrapper` with 2022 for the deadline.

inducer: Sorry for the noise! I thought I had deleted this comment. I wonder which comment I deleted instead. :scream:

meshmode/array_context.py
```diff
@@ -76,6 +92,433 @@ def _loopy_get_default_entrypoint(t_unit):
                     "translation unit")
 
 
+# {{{ ArrayContainer
```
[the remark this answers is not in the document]
inducer: Given your thumbs-up, let's just put an `arraycontext` package on the docket this week, after this merges. Then we can forget about the separate-file business.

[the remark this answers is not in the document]
inducer: Made an empty shell: https://github.com/inducer/arraycontext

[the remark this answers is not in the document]
inducer: Nvm this then.

meshmode/array_context.py
```diff
@@ -76,6 +92,433 @@ def _loopy_get_default_entrypoint(t_unit):
...
+        an :class:`ArrayContext`.
+    :param iterable: an iterable that mirrors the output of
+        :meth:`serialize_container`.
+    """
+    return deserialize_container_class.dispatch(cls)(cls, actx, iterable)
+
+
+@singledispatch
+def get_container_context(ary: ArrayContainer):
+    # FIXME: is singledispatch confused by the empty ABC?
```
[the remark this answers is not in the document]
alexfikl: I think just having an `ArrayContainer(ABC)` with no abstract methods or properties acts pretty much as just saying `ary: object` so it accepts everything.

The only reason this is an ABC right now is because of [ABC.register](https://docs.python.org/3/library/abc.html#abc.ABCMeta.register) which allows easily adding new classes. We'll probably need to implement some sort of easily extensible `__instancecheck__` instead.

meshmode/array_context.py
alexfikl: The check isn't super useful, but this needs an `actx.freeze` to pass on to `map_array_container`, so I see two ways out of it
* pass an `actx` to `freeze` too
* copy `map_array_container` in here like it was before so that it can get an `actx` along the way

Any preference?

meshmode/dof_array.py
alexfikl: This actually doesn't need an array context at this level, my bad. The wrapped `_flatten_dof_array` gets a `DOFArray` from which we can get the context without traversing anything.

meshmode/array_context.py
alexfikl: Added a little blurb and link in 5bfa762.

meshmode/array_context.py
alexfikl: Removed the check and the call to `get_container_context` in `thaw`.

meshmode/dof_array.py
alexfikl: Somewhere in 0dbc6b2.

meshmode/array_context.py
alexfikl: Done in 0dbc6b2. 

It just returns `None`. Would you prefer it raises a `NotImplementedError`?

test/test_array.py
alexfikl: Added in 42df8f3.

meshmode/array_context.py
inducer: Not a huge fan. It's not idiomatic in Python, the name will look weird as a decorator (IMO?) and we'll have to keep paying for the conditional.

meshmode/array_context.py
inducer: `None` is good.

meshmode/array_context.py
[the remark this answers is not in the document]
alexfikl: Gave this a go in 5948d34.

It seems to work, but I'm not sure what corner cases it's missing.

meshmode/array_context.py
```diff
@@ -76,6 +92,435 @@ def _loopy_get_default_entrypoint(t_unit):
...
+        is an array-like component of the :class:`ArrayContainer`.
+        Components can themselves be :class:`ArrayContainer`\ s, allowing
+        for arbitrarily nested structures. The identifiers need to be hashable
+        but are otherwise treated as opaque.
+    """
+    raise NotImplementedError(type(ary).__name__)
+
+
+@singledispatch
+def deserialize_container_class(cls: type,
```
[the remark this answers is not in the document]
alexfikl: The issue is that when you serialize, you have an instance that `singledispatch` can use, but when you deserialize, you just have the class. That's why
* `deserialize_container` actually does the dispatch (by calling `singledispatch.dispatch`)
* `deserialize_container_class` is used for registering new implementations.

I'm open for suggestions on how to do this in a cleaner way?

meshmode/array_context.py
```diff
@@ -76,6 +92,435 @@ def _loopy_get_default_entrypoint(t_unit):
...
+        actx: Optional["ArrayContext"],
+        iterable: Iterable[Tuple[Any, Any]]):
+    raise NotImplementedError(cls.__name__)
+
+
+def deserialize_container(cls: type,
```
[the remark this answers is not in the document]
alexfikl: See https://github.com/inducer/meshmode/pull/175#discussion_r627593553

meshmode/array_context.py
```diff
@@ -20,23 +20,44 @@
...
+.. autoclass:: ArrayContainer
+.. autofunction:: is_array_container
+.. autofunction:: serialize_container
+.. autofunction:: deserialize_container
+.. autofunction:: get_container_context
+.. autofunction:: get_container_context_recursively
+
+.. autoclass:: ArrayContainerWithArithmetic
+.. autoclass:: DataclassContainer
+.. autoclass:: DataclassContainerWithArithmetic
```
inducer, line 52:
```suggestion
.. autoclass:: ArithArrayContainer
.. autoclass:: DataclassArrayContainer
.. autoclass:: DataclassArithArrayContainer
```
?

alexfikl: * For `ArrayContainerWithArithmetic`: I kind of like how boring and obvious the original is. Any pros/cons?
* For the dataclass ones, yeah, that sounds good!

alexfikl: Renamed the dataclasses in 11009fc.

meshmode/dof_array.py
alexfikl: Renamed in a5c4493.

meshmode/array_context.py
alexfikl: Renamed in a5c4493.

meshmode/array_context.py
alexfikl: Resolved this by adding a non-recursive map in ed3c0f2.

meshmode/dof_array.py
alexfikl: Renamed ina a5c4493

alexfikl:
@inducer Added some non-recursive map and fixed up the `__instancecheck__`, so this should be good for another look. There's still a few things in there that I had questions about too.

meshmode/array_context.py
```diff
@@ -76,6 +99,548 @@ def _loopy_get_default_entrypoint(t_unit):
...
+        actx, ary_actx = _update_container_context(ary, actx)
+        if recursive:
+            f = partial(_map_array_container_with_context,
+                    f, actx=actx, scalar_cls=scalar_cls)
+
+        return deserialize_container(type(ary), ary_actx, (
+                (key, f(subary)) for key, subary in serialize_container(ary)
+                ))
+    else:
+        return f(ary)
```
inducer, line 486: Error instead?

meshmode/array_context.py
```diff
@@ -76,6 +99,548 @@ def _loopy_get_default_entrypoint(t_unit):
...
+            ]
+
+    if not container_indices:
+        return f(*args)
+
+    template_ary = args[container_indices[0]]
+    if not all(isinstance(args[i], type(template_ary)) for i in container_indices):
+        raise TypeError(
+                "'ArrayContainer' arguments must be of the same type: "
+                f"{type(template_ary).__name__}")
```
inducer, line 506: This is weirdly asymmetric: Subsequent args can be of subtypes of the first, but not the other way around. Why not force everybody to have exactly the same type?

meshmode/array_context.py
```diff
@@ -76,6 +99,548 @@ def _loopy_get_default_entrypoint(t_unit):
...
+            if isinstance(arg1, np.ndarray):
+                return obj_array_vectorize(
+                        lambda subary: cls.binary_op(op, subary, arg2),
+                        arg1.astype(object, copy=False))
+            else:
+                return obj_array_vectorize(
+                        lambda subary: cls.binary_op(op, arg1, subary),
+                        arg2.astype(object, copy=False))
+        elif isinstance(arg1, ArrayContainer) and isinstance(arg2, ArrayContainer):
+            return multimap_array_container(op, arg1, arg2)
```
inducer, line 303: I fear multimap is fairly slow, what with having to discover who is a container and who isn't. This case is much simpler, and there's a substantial premium for doing this quickly.

meshmode/array_context.py
```diff
@@ -76,6 +99,548 @@ def _loopy_get_default_entrypoint(t_unit):
...
+        return map_array_container(op, arg)
+
+    @classmethod
+    def binary_op(cls, op, arg1, arg2):
+        from numbers import Number
+        from pytools.obj_array import obj_array_vectorize
+
+        if isinstance(arg1, np.ndarray) or isinstance(arg2, np.ndarray):
+            # do a bit of broadcasting
+            if isinstance(arg1, np.ndarray):
```
inducer, line 294: Reuse `isinstance(arg1, np.ndarray)`?

meshmode/array_context.py
```diff
@@ -76,6 +99,548 @@ def _loopy_get_default_entrypoint(t_unit):
...
+    :returns: a tuple ``(actx, ary_actx)``, where the first array context
+        is meant to be passed down the container hierarchy and the second one
+        is meant to be used to deserialize the array container.
+    """
+    if actx is not _ArrayContextNotProvided:
+        return actx, actx
+
+    ary_actx = get_container_context(ary)
+    actx = actx if ary_actx is None else ary_actx
+    return actx, ary_actx
```
inducer, line 467: Why are there two array contexts flying around here? What is the use case for two distinct ones? (Shouldn't that just be banned?)

meshmode/array_context.py
```diff
@@ -76,6 +99,548 @@ def _loopy_get_default_entrypoint(t_unit):
...
+        or an instance of a base array type.
+    """
+    return _map_array_container_with_context(f, ary, recursive=False)
+
+
+def array_container_vectorize_n_args(f: Callable[[Any], Any], *args):
+    r"""Applies *f* to the components of multiple :class:`ArrayContainer`\ s.
+
+    Works similarly to :func:`~pytools.obj_array.obj_array_vectorize_n_args`,
+    but on arbitrar containers. The containers must all have the same type,
```
inducer, line 557:
```suggestion
    but on arbitrary containers. The containers must all have the same type,
```

meshmode/array_context.py
inducer: Ah, good point. Let's stick with the conservative route for now.

meshmode/array_context.py
inducer: No need to flatten. At least the (little-)l^p norms (including l^inf) can be applied recursively. There's a bit of inefficiency because of repeated p-sqrt and p-power, but :shrug:.

meshmode/array_context.py
```diff
@@ -76,6 +99,548 @@ def _loopy_get_default_entrypoint(t_unit):
...
+                (key, f(subary)) for key, subary in serialize_container(ary)
+                ))
+    else:
+        return f(ary)
+
+
+def _multimap_array_container_with_context(f, *args,
+        actx=_ArrayContextNotProvided,
+        scalar_cls=None,
+        recursive=True):
```
inducer, line 492: I'd like recursion to be opt-in, I think.

meshmode/array_context.py
```diff
@@ -76,6 +99,548 @@ def _loopy_get_default_entrypoint(t_unit):
...
+
+    ary_actx = get_container_context(ary)
+    actx = actx if ary_actx is None else ary_actx
+    return actx, ary_actx
+
+
+def _map_array_container_with_context(f, ary,
+        actx=_ArrayContextNotProvided,
+        scalar_cls=None,
+        recursive=True):
```
inducer, line 473: I'd like recursion to be opt-in, I think.

meshmode/array_context.py
```diff
@@ -76,6 +99,548 @@ def _loopy_get_default_entrypoint(t_unit):
...
+    """
+    if actx is not _ArrayContextNotProvided:
+        return actx, actx
+
+    ary_actx = get_container_context(ary)
+    actx = actx if ary_actx is None else ary_actx
+    return actx, ary_actx
+
+
+def _map_array_container_with_context(f, ary,
```
inducer, line 470:
```suggestion
def _map_array_container_with_context(f, ary, *,
```

meshmode/array_context.py
```diff
@@ -76,6 +99,548 @@ def _loopy_get_default_entrypoint(t_unit):
...
+                    f, actx=actx, scalar_cls=scalar_cls)
+
+        return deserialize_container(type(ary), ary_actx, (
+                (key, f(subary)) for key, subary in serialize_container(ary)
+                ))
+    else:
+        return f(ary)
+
+
+def _multimap_array_container_with_context(f, *args,
```
inducer, line 489:
```suggestion
def _multimap_array_container_with_context(f, *args, *,
```

meshmode/array_context.py
```diff
@@ -76,6 +99,548 @@ def _loopy_get_default_entrypoint(t_unit):
...
+    actx = actx if ary_actx is None else ary_actx
+    return actx, ary_actx
+
+
+def _map_array_container_with_context(f, ary,
+        actx=_ArrayContextNotProvided,
+        scalar_cls=None,
+        recursive=True):
+    if scalar_cls is not None and type(ary) is scalar_cls:
+        return f(ary)
```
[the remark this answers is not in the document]
inducer: I just read below that this is the stop-recursion-on thing. As you can tell, I'm not digging the naming.

inducer (commented):
Thanks for working on this! I've got *cough* a few more things. Sorry about the sheer volume here... But I think it's important that we get this right. :)

meshmode/array_context.py
```diff
@@ -76,6 +99,548 @@ def _loopy_get_default_entrypoint(t_unit):
...
+@deserialize_container_class.register(np.ndarray)
+def _(cls: type, actx: "ArrayContext", iterable):
+    assert cls is np.ndarray
+    iterable = list(iterable)
+
+    result = cls(len(iterable), dtype=object)
+    shape = np.full(len(iterable[0][0]), -1)
+
+    for i, (index, subary) in enumerate(iterable):
+        shape = np.maximum(shape, index)
```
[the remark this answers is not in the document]
alexfikl: Would the `template` argument to be enough to clean this up? Although there may not always be one.. Hm..

meshmode/array_context.py
```diff
@@ -76,6 +99,548 @@ def _loopy_get_default_entrypoint(t_unit):
...
+
+
+def get_container_context_recursively(ary: Any):
+    """Walks the :class:`ArrayContainer` hierarchy to find an :class:`ArrayContext`
+    associated with it.
+
+    If different components that have different array contexts are found,
+    an assertion error is raised.
+    """
+    actx = None
```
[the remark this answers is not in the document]
alexfikl: Changed in 3d99b48.

meshmode/array_context.py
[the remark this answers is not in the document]
alexfikl: I called that `leaf_cls` initially, but didn't like that much either. Suggestions? `base_cls`?

meshmode/array_context.py
alexfikl: Fixed in beb7728.

meshmode/array_context.py
alexfikl: It's mostly for `freeze`/`thaw`. E.g. `freeze` needs to pass `None` to `deserialize`, but still needs to carry a context around to actually to the freezing at the leaf level.

I'll go over those again and try to get rid of this. Agreeed that it's not pretty!

meshmode/array_context.py
alexfikl: Updated in 3dcbc79.

alexfikl:
@inducer Looks like someone broke firedrake and it's not me? (fingers crossed)

inducer:
#183, maybe?

meshmode/array_context.py
```diff
@@ -75,6 +100,619 @@ def _loopy_get_default_entrypoint(t_unit):
...
+        for arbitrarily nested structures. The identifiers need to be hashable
+        but are otherwise treated as opaque.
+    """
+    raise NotImplementedError(type(ary).__name__)
+
+
+@singledispatch
+def deserialize_container_class(cls: type, template: Any,
+        iterable: Iterable[Tuple[Any, Any]], *,
+        actx: Optional["ArrayContext"] = None):
```
inducer, line 182: How come this still has an `actx` argument? Can't we just always get the array context from the template?

alexfikl: It still has one for freeze/thaw when you want to replace whatever array context is in there.

alexfikl: By default in `deserialize_container` (without the `_class`) it uses the array container from the template if not provided.

meshmode/array_context.py
```diff
@@ -75,6 +100,619 @@ def _loopy_get_default_entrypoint(t_unit):
...
+
+class ArrayContainerMeta(type):
+    def __instancecheck__(cls, instance):
+        if issubclass(type(instance), cls):
+            return True
+
+        if cls is ArrayContainer:
+            return is_array_container(instance)
+
+        return False
```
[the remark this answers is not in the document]
inducer: https://github.com/alexfikl/meshmode/pull/5

meshmode/array_context.py
inducer: This is too gross, IMO. I had ripped all this array context logic out, and it felt great, only to stumble on the exact problem you point out.

I think a better way out of this is to make `freeze`/`thaw` `singledispatch` as well, so that array types can override their... temperature change. That way the `DOFArray` can do what it needs to without all this logic looking like somebody pulled it out of the toilet.

meshmode/array_context.py
```diff
@@ -75,6 +100,623 @@ def _loopy_get_default_entrypoint(t_unit):
...
+
+
+class ArrayContainer(metaclass=ArrayContainerMeta):
```
inducer, line 123: We should also profile this to try and reduce cost as much as possible. In an unscientific test, `simple-dg` on f809720 took 97s, whereas on 6bab0e7 (`main`), it took 92s or so.

alexfikl: Agreed. 

The `singledispatch` stuff isn't the fastest either, which is not surprising as it does more work to dispatch. I remember benchmarking it once against just one function with bunch of `if isinstance(...)` and it was clearly slower.

meshmode/array_context.py
```diff
@@ -75,6 +100,619 @@ def _loopy_get_default_entrypoint(t_unit):
...
+    """Helper for :func:`map_array_container`.
+
+    :param actx: :class:`ArrayContext` passed in to :func:`deserialize_container`.
+        If not provided, the context of the serialized container is used, if any.
+    :param leaf_cls: class on which we call *f* directly. This is mostly
+        useful in the recursive setting, where it can stop the recursion on
+        specific container classes. By default, the recursion is stopped when
+        a non-:class:`ArrayContainer` class is encountered.
+    """
+    if leaf_cls is not None and type(ary) is leaf_cls:
```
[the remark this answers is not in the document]
alexfikl: Yes! This used to be a `and isinstance(ary, leaf_cls)`, which raises with `None`.

meshmode/array_context.py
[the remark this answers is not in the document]
alexfikl: Fixed by requiring a `template` argument.

meshmode/array_context.py
alexfikl: Hopefully ``singledispatch` will be a bit faster with alexfikl/meshmode#5, since it seems to do a fancier algorithm when ABCs are involved.

meshmode/array_context.py
inducer: FYI I'm working on this.

inducer: https://github.com/alexfikl/meshmode/pull/6

meshmode/array_context.py
```diff
@@ -75,6 +100,618 @@ def _loopy_get_default_entrypoint(t_unit):
...
+        If not provided, the context of the serialized container is used, if any.
+    :param leaf_cls: class on which we call *f* directly. This is mostly
+        useful in the recursive setting, where it can stop the recursion on
+        specific container classes. By default, the recursion is stopped when
+        a non-:class:`ArrayContainer` class is encountered.
+    """
+    container_indices = [
+            i for i, arg in enumerate(args)
+            if is_array_container(arg)
+            and (leaf_cls is None or type(arg) is not leaf_cls)
```
inducer, line 569:
```suggestion
            and (type(arg) is not leaf_cls)
```

alexfikl:
I did some naive timing of the `simple-dg` example and it does seem to be slower. On main it seems to go around 29-30s and on this branch at around 32-33.

I did a quick profile with py-spy but nothing pops out. I'll stare some more and hopefully figure out what's up. I'm mostly inclined to blame all the serializing and deserializing going through `singledispatch`.

meshmode/array_context.py
inducer: All this stuff shows up pretty prominently in profiles for now:
```
$ vmprofshow test.prof flat --percent-cutoff=5 --no-native --include-callees
   133.103% - wrapper:/usr/lib/python3.9/functools.py:872
    99.828% - wrapper:/home/andreas/src/pytools/pytools/__init__.py:2527
    99.828% - main:/home/andreas/src/meshmode/examples/simple-dg.py:461
    99.368% - rk4_step:/home/andreas/src/meshmode/examples/simple-dg.py:419
    94.138% - rhs:/home/andreas/src/meshmode/examples/simple-dg.py:493
    94.138% - wave_operator:/home/andreas/src/meshmode/examples/simple-dg.py:397
    90.287% - rec:/home/andreas/src/meshmode/meshmode/array_context.py:512
    86.207% - _map_binary_op_array_container:/home/andreas/src/meshmode/meshmode/array_context.py:258
    74.195% - <genexpr>:/home/andreas/src/meshmode/meshmode/array_context.py:516
    69.598% - _map_array_container_impl:/home/andreas/src/meshmode/meshmode/array_context.py:504
    65.402% - __call__:/home/andreas/src/meshmode/meshmode/discretization/connection/direct.py:272
    64.655% - _binary_op:/home/andreas/src/meshmode/meshmode/array_context.py:293
    55.230% - map_array_container:/home/andreas/src/meshmode/meshmode/array_context.py:592
    49.080% - _deserialize_dataclass_container:/home/andreas/src/meshmode/meshmode/array_context.py:468
    48.966% - wave_flux:/home/andreas/src/meshmode/examples/simple-dg.py:373
    48.391% - _deserialize_dof_container:/home/andreas/src/meshmode/meshmode/dof_array.py:301
    47.644% - <genexpr>:/home/andreas/src/meshmode/meshmode/array_context.py:262
    45.920% - <genexpr>:/home/andreas/src/meshmode/meshmode/dof_array.py:311
    29.828% - _deserialize_ndarray_container:/home/andreas/src/meshmode/meshmode/array_context.py:206
    26.782% - kernel_runner:/home/andreas/src/pyopencl/pyopencl/array.py:167
    20.402% - call_loopy:/home/andreas/src/meshmode/meshmode/array_context.py:1302
    19.713% - interp:/home/andreas/src/meshmode/examples/simple-dg.py:156
    19.310% - __rmul__:/home/andreas/src/meshmode/meshmode/array_context.py:360
```

inducer: Probably more informative:
```
vmprofshow test.prof flat --percent-cutoff=1 --no-native
    10.575% - __init__:/home/andreas/src/pyopencl/pyopencl/array.py:417
     6.207% - enqueue_knl_multiply:<generated code>:6
     6.092% - enqueue_knl_axpbyz:<generated code>:6
     5.230% - enqueue_knl_axpb:<generated code>:6
     4.770% - _lpy_host_resample_by_picking:<generated code>:7
     3.736% - kernel_runner:/home/andreas/src/pyopencl/pyopencl/array.py:167
     2.701% - _binary_op:/home/andreas/src/meshmode/meshmode/array_context.py:293
     2.529% - invoke_resample_by_picking_loopy_kernel:<generated code>:47
     2.184% - __call__:/home/andreas/src/meshmode/meshmode/discretization/connection/direct.py:272
     1.839% - call_loopy:/home/andreas/src/meshmode/meshmode/array_context.py:1302
     1.782% - arg_to_dtype_set:/home/andreas/src/loopy/loopy/target/execution.py:787
     1.667% - _deserialize_dof_container:/home/andreas/src/meshmode/meshmode/dof_array.py:301
     1.552% - _lpy_host_diff:<generated code>:7
     1.379% - _map_binary_op_array_container:/home/andreas/src/meshmode/meshmode/array_context.py:258
```

alexfikl:
@inducer Added some code generation for the dataclass serialization in 4a7078f. Not quite sure if that's how it's done™, so suggestions are most welcome.

It didn't help much with the profile thought, since the `_binary_op` stuff seems to be the main culprit.

meshmode/array_context.py
alexfikl: Mostly a note to self, but that `_deserialize_dof_container` is misleading and it's just `_binary_op` and `_map_binary_op_array_container` that are doing some actual work.

The iterators passed in to `deserialize` are generators, so the actual work is done when iterating over them. I changed them to lists in 4a7078f and `_deserialize_dof_container` disappeared from the profile.

alexfikl:
@inducer Is there anything else that needs doing here?

The code generation for the arithmetic is a bit clunky, but it seems to be doing its job well enough. Besides that, this looks good to me.

meshmode/array_context.py
```diff
@@ -75,6 +100,766 @@ def _loopy_get_default_entrypoint(t_unit):
...
+        ``**``, :func:`divmod`, and ``//``. Also includes ``+`` and ``-`` as well as
+        :func:`abs`.
+    :arg bitwise: If *True*, implement bitwise and, or, not, and inversion.
+    :arg shift: If *True*, implement bit shifts.
+    :arg eq_comparison: If *True*, implement ``==`` and ``!=``.
+    :arg rel_comparison: If *True*, implement ``<``, ``<=``, ``>``, ``>=``.
+        In that case, if *eq_comparison* is unspecified, it is also set to
+        *True*.
+
+    Each operator class also includes the "reverse" operator class if applicable.
```
inducer, line 376:
```suggestion
    Each operator class also includes the "reverse" operators if applicable.
```

meshmode/array_context.py
```diff
@@ -75,6 +100,766 @@ def _loopy_get_default_entrypoint(t_unit):
...
+
+def dataclass_array_container(cls):
+    """A class decorator that makes the class to which it is applied a
+    :class:`ArrayContainer` by registering appropriate implementations of
+    :func:`serialize_container` and :func:`deserialize_container`.
+    *cls* must be a :func:`~dataclasses.dataclass`.
+
+    Attributes that are not array containers are allowed. In order to decide
+    whether an attribute is an array container, the declared attribute type
+    is checked for whether it is a subclass of :class:`ArrayContainer`.
```
inducer, line 578:
```suggestion
    is checked via :func:`is_array_container_type`.
```

meshmode/array_context.py
```diff
@@ -75,6 +100,766 @@ def _loopy_get_default_entrypoint(t_unit):
...
+                cls, template_instance_name, args):
+            return f"{deserialize_init_code}"
+
+        cls._deserialize_init_arrays_code = classmethod(
+            _deserialize_init_arrays_code_{lower_cls_name})
+        """)
+
+    exec_dict = {"cls": cls}
+    exec(compile(serialize_code, "<generated code>", "exec"), exec_dict)
+    exec_dict["_MODULE_SOURCE_CODE"] = serialize_code
```
inducer, line 641:
```suggestion
    exec_dict = {"cls": cls, "_MODULE_SOURCE_CODE": serialize_code}
    exec(compile(serialize_code, "<generated code>", "exec"), exec_dict)
```

meshmode/array_context.py
```diff
@@ -75,6 +100,766 @@ def _loopy_get_default_entrypoint(t_unit):
...
+
+    Works similarly to :func:`~pytools.obj_array.obj_array_vectorize`, but
+    on arbitrary containers.
+
+    For a recursive version, see :func:`rec_map_array_container`.
+
+    :param ary: a (potentially nested) structure of :class:`ArrayContainer`\ s,
+        or an instance of a base array type.
+    """
+    return _map_array_container_impl(f, ary, recursive=False)
```
inducer, line 746:
```suggestion
    if is_array_container(ary):
        return deserialize_container(ary, [
                (key, f(subary)) for key, subary in serialize_container(ary)
                ])
    else:
        return f(ary)
```

meshmode/array_context.py
```diff
@@ -75,6 +100,766 @@ def _loopy_get_default_entrypoint(t_unit):
...
+        specific container classes. By default, the recursion is stopped when
+        a non-:class:`ArrayContainer` class is encountered.
+    """
+    def is_same_type(arys, template_ary):
+        return all(type(arys[i]) is type(template_ary) for i in container_indices)
+
+    def rec(*_args):
+        template_ary = _args[container_indices[0]]
+        assert is_same_type(_args, template_ary), \
+                "expected type '{type(template_ary).__name__}'"
```
inducer, line 686:
```suggestion
    def rec(*_args):
        template_ary = _args[container_indices[0]]
        assert all(type(arys[i]) is type(template_ary) for i in container_indices[1:]), \
                "expected type '{type(template_ary).__name__}'"        
```

meshmode/array_context.py
```diff
@@ -75,6 +100,766 @@ def _loopy_get_default_entrypoint(t_unit):
...
+                serialize_container(_args[i]) for i in container_indices
+                ]):
+            key = None
+            for i, subary in zip(container_indices, subarys):
+                if key is None:
+                    key = subary[0]
+                else:
+                    assert key == subary[0]
+
+                new_args[i] = subary[1]
```
inducer, line 705:
```suggestion
            for i, (subkey, subary) in zip(container_indices, subarys):
                if key is None:
                    key = subkey
                else:
                    assert key == subkey

                new_args[i] = subary
```

meshmode/array_context.py
```diff
@@ -75,6 +100,766 @@ def _loopy_get_default_entrypoint(t_unit):
...
+
+    :param args: all :class:`ArrayContainer` arguments must be of the same
+        type and with the same structure (same number of components, etc.).
+    """
+    return _multimap_array_container_impl(f, *args, recursive=True)
+
+
+def multimapped_over_array_containers(f: Callable[[Any], Any]):
+    """Decorator around :func:`rec_multimap_array_container`."""
+    def wrapper(*args):
```
inducer, line 795:
```suggestion
    # can't use functools.partial, because its result is insufficiently
    # function-y to be used as a method definition.
    def wrapper(*args):
```

meshmode/array_context.py
```diff
@@ -75,6 +100,766 @@ def _loopy_get_default_entrypoint(t_unit):
...
+
+    Array container types may use :func:`functools.singledispatch` ``.register`` to
+    register additional implementations.
+
+    See :meth:`ArrayContext.thaw`.
+    """
+    if is_array_container(ary):
+        return _map_array_container_impl(
+                partial(freeze, actx=actx), ary,
+                recursive=False)
```
inducer, line 817:
```suggestion
        return map_array_container(partial(freeze, actx=actx), ary)
```

inducer (commented):
Reviewed through `thaw`. Will look at the rest later tonight.

meshmode/array_context.py
```diff
@@ -75,6 +100,769 @@ def _loopy_get_default_entrypoint(t_unit):
...
+    raise NotImplementedError(type(template).__name__)
+
+
+def is_array_container_type(cls: type) -> bool:
+    """
+    :returns: *True* if the type *cls* has a registered implementation of
+        :func:`serialize_container`, or if it is an :class:`ArrayContainer`.
+    """
+    return (
+            cls is ArrayContainer
```
alexfikl, line 187: Should this be `issubclass(cls, ArrayContainer)`?

meshmode/array_context.py
```diff
@@ -75,6 +100,769 @@ def _loopy_get_default_entrypoint(t_unit):
...
+            or (serialize_container.dispatch(cls)
+                is not serialize_container.__wrapped__))
+
+
+def is_array_container(ary: Any) -> bool:
+    """Return *True* if the instance *ary* has a registered implementation of
+    :func:`serialize_container`.
+    """
+    return (serialize_container.dispatch(ary.__class__)
```
alexfikl, line 196: Should this have a `type(ary) is ArrayContainer` too?

meshmode/array_context.py
inducer: According to the [docs](https://github.com/alexfikl/meshmode/blob/836fa20b4c473fdd5d88a686515822cc649bf0e4/meshmode/array_context.py#L142-L143), subclassing `ArrayContainer` doesn't do anything special for you (in particular, it doesn't make you an array container). So this branch is here mostly to serve type annotation, where somebody says "this attribute is an `ArrayContainer`.

meshmode/array_context.py
inducer: See above. The `ArrayContainer` bit being left out was actually kind of intentional. But I agree it's a bit weird. `ArrayContainer` is like a Julia-style ABC, or something.

meshmode/array_context.py
```diff
@@ -528,52 +1317,56 @@ def max(self, a):
...
-        # downstream headaches.
-        from meshmode.dof_array import DOFArray
-        if isinstance(array, DOFArray):
+        if is_array_container(ary):
             import numpy.linalg as la
-            return la.norm(np.array([
-                self.norm(_flatten_grp_array(grp_ary), ord)
-                for grp_ary in array]), ord)
+            return la.norm([
+                self.norm(_flatten_array(subary), ord=ord)
```
inducer, line 1356: I vote that we take norms recursively over containers. How does that sound to you?

meshmode/array_context.py
alexfikl: I was also thinking of this as a shortcut to not call `singledispatch.dispatch` (which does a bit of work). E.g. for `DOFArray`, which subclasses `ArrayContainer`, it would have an earlier exit.

meshmode/discretization/connection/chained.py
```diff
@@ -62,7 +61,6 @@ def __init__(self, connections, from_discr=None):
 
         self.connections = connections
 
-    @obj_array_vectorized_n_args
     def __call__(self, ary):
```
inducer, line 64: Does this no longer map over containers?

meshmode/array_context.py
alexfikl: Agreed with the ABC part. I was mostly just thinking that that check could sidestep a slightly more expensive call to `singledispatch.dispatch` (although [it doesn't seem to do much work](https://github.com/python/cpython/blob/main/Lib/functools.py#L830-L838) when the class is already in the cache).

meshmode/dof_array.py
```diff
@@ -55,7 +58,11 @@
 
 # {{{ DOFArray
 
-class DOFArray:
+@with_container_arithmetic(
+        bcast_obj_array=True,
+        bcast_numpy_array=True,
+        rel_comparison=True)
+class DOFArray(ArrayContainer):
```
inducer, line 65:
```suggestion
class DOFArray:
```

meshmode/array_context.py
inducer: :shrug: Not super worried about `singledispatch.dispatch`. And I'm about to propose not having `DOFArray` inherit from `ArrayContainer`.

meshmode/array_context.py
inducer: :shrug: Still not super worried about `singledispatch.dispatch`. :)

meshmode/array_context.py
inducer: cb214fa, in fact.

meshmode/array_context.py
```diff
@@ -528,52 +1317,56 @@ def max(self, a):
...
-    if grp_ary.flags.f_contiguous:
-        return grp_ary.reshape(-1, order="F")
-    elif grp_ary.flags.c_contiguous:
-        return grp_ary.reshape(-1, order="C")
+        return ary._new_with_changes(
+                data=None, offset=0, shape=(0,), strides=(ary.dtype.itemsize,))
+    if ary.flags.f_contiguous:
+        return ary.reshape(-1, order="F")
+    elif ary.flags.c_contiguous:
+        return ary.reshape(-1, order="C")
```
inducer, line 1338: This norm thing is actually quite inconsistent and strange. CL arrays in a container (*cough* `DOFArray` *cough* but nothing checks for that) get flattened, but if you otherwise hand the norm a 2D array, it goes, "no sir!"

I think the cleanest would be to introduce a separate `DOFArray` norm function, and introduce a deprecated special-case here that we'll keep working for a little to give people time to switch. I'll get on that. Let me know if you don't like it.

meshmode/array_context.py
inducer: 47055a7

meshmode/array_context.py
inducer: 47055a7. Let me know if you agree.

PR #176, opened 2021-05-01
Add comment explaining why einsum lives in actx, not actx.np

inducer:
cc @alexfikl

### DIFF
--- a/meshmode/array_context.py
+++ b/meshmode/array_context.py
@@ -415,11 +415,12 @@ class ArrayContext(ABC):
     # metadata describing what's happening, transform_loopy_program
     # has a very difficult (hopeless?) job to do.
     #
-    # Unfortunately, the existing metadata support (cf .tag()) cannot
-    # help with eager mode execution, because, by definition, when the
+    # Unfortunately, the existing metadata support (cf. .tag()) cannot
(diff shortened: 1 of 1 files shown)


meshmode/array_context.py
```diff
@@ -408,6 +408,18 @@ def _get_einsum_prg(self, spec, arg_names, tagged):
...
+    # Well, optimizing generic einsum for OpenCL/GPU execution
+    # is actually difficult, even in eager mode, and so without added
+    # metadata describing what's happening, transform_loopy_program
+    # has a very difficult (hopeless?) job to do.
+    #
+    # Unfortunately, the existing metadata support (cf .tag()) cannot
+    # help with eager mode execution, because, by definition, when the
+    # result is passed to .tag(), it is already computed.
+    # That's why einsum's interface here needs to be cluttered with
+    # metadata, and that's why it can't live under .np.
```
inducer, line 422:
```suggestion
    # Unfortunately, the existing metadata support (cf. .tag()) cannot
    # help with eager mode execution [1], because, by definition, when the
    # result is passed to .tag(), it is already computed.
    # That's why einsum's interface here needs to be cluttered with
    # metadata, and that's why it can't live under .np.
    # [1] https://github.com/inducer/meshmode/issues/177
```

PR #179, opened 2021-05-05
Enable flake8-bugbear

inducer:


alexfikl:
Does this catch anything a more aggressive pylint wouldn't catch?

inducer:
> Does this catch anything a more aggressive pylint wouldn't catch?

Great question, don't know. Quite possibly no. I just tried it and liked the SNR of the warnings.

PR #194, opened 2021-05-21
Firedrake CI, fix delay-setting code for oct2py in nodal-dg interop

inducer:
This was failing very frequently on gitlab, and I had to go check every time to see if it was an actual failure or just *that again*. It finally annoyed me sufficiently.


inducer:
Also (tries to) fix Firedrake CI: 664aeaa

PR #196, opened 2021-05-24
Add flatten and unflatten to numpy

alexfikl:
This basically just moves the code from pytential
https://github.com/inducer/pytential/blob/main/pytential/utils.py
with `flatten_if_needed` getting folded into `flatten` by just doing nothing if it's passed a non-`DOFArray`.

I have a version of these functions in my code as well and they're pretty useful for talking with the outside world. Doing `map_dof_array_container(actx.to_numpy, flatten(ary))` isn't too long either, but that's two traversals.

There's a version of pytential using this stuff at inducer/pytential#88.

### DIFF
--- a/meshmode/dof_array.py
+++ b/meshmode/dof_array.py
@@ -390,7 +390,7 @@ def _flatten_dof_array(ary: Any, strict: bool = True):
     if actx is None:
         raise ValueError("cannot flatten frozen DOFArrays")
 
-    @memoize_in(actx, (flatten, "flatten_prg"))
+    @memoize_in(actx, (_flatten_dof_array, "flatten_prg"))
     def prg():
         return make_loopy_program(
             "{[iel,idof]: 0<=iel<nelements and 0<=idof<ndofs_per_element}",
@@ -446,7 +446,7 @@ def _unflatten_dof_array(actx: ArrayContext, ary: Any,
         else:
             return ary
 
-    @memoize_in(actx, (unflatten, "unflatten_prg"))
+    @memoize_in(actx, (_unflatten_dof_array, "unflatten_prg"))
     def prg():
         return make_loopy_program(
             "{[iel,idof]: 0<=iel<nelements and 0<=idof<ndofs_per_element}",
--- a/meshmode/dof_array.py
+++ b/meshmode/dof_array.py
@@ -378,7 +378,41 @@ def multimapped_over_dof_arrays(f):
 
 # {{{ flatten / unflatten
 
-def flatten(ary: ArrayContainer) -> ArrayContainer:
+def _flatten_dof_array(ary: Any, strict: bool = True):
+    if not isinstance(ary, DOFArray):
+        if strict:
+            raise TypeError(f"non-DOFArray type '{type(ary).__name__}' cannot "
+                    "be flattened; use 'strict=False' to allow other types")
+        else:
+            return ary
+
+    actx = ary.array_context
+    if actx is None:
+        raise ValueError("cannot flatten frozen DOFArrays")
+
+    @memoize_in(actx, (flatten, "flatten_prg"))
+    def prg():
+        return make_loopy_program(
+            "{[iel,idof]: 0<=iel<nelements and 0<=idof<ndofs_per_element}",
+            """result[grp_start + iel*ndofs_per_element + idof] \
+                = grp_ary[iel, idof]""",
+            name="flatten")
+
+    group_sizes = [grp_ary.shape[0] * grp_ary.shape[1] for grp_ary in ary]
+    group_starts = np.cumsum([0] + group_sizes)
+
+    result = actx.empty(group_starts[-1], dtype=ary.entry_dtype)
+
+    for grp_start, grp_ary in zip(group_starts, ary):
+        actx.call_loopy(prg(),
+                grp_ary=grp_ary,
+                result=result,
+                grp_start=grp_start)
+
+    return result
+
+
+def flatten(ary: ArrayContainer, *, strict: bool = True) -> ArrayContainer:
     r"""Convert all :class:`DOFArray`\ s into a "flat" array of degrees of
     freedom, where the resulting type of the array is given by the
     :attr:`DOFArray.array_context`.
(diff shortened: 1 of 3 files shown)


inducer:
LGTM. Thanks!

alexfikl:
> LGTM. Thanks!

Hold off until inducer/pytential#88 passes too?

meshmode/dof_array.py
```diff
@@ -477,6 +485,43 @@ def _unflatten_dof_array(subary):
...
+def unflatten_from_numpy(
+        actx: ArrayContext, discr, ary: ArrayContainer,
+        ndofs_per_element_per_group: Optional[Iterable[int]] = None
+        ) -> ArrayContainer:
+    r"""Takes "flat" arrays returned by :func:`flatten_to_numpy` and
+    reconstructs the corresponding :class:`DOFArray`\ s using :func:`unflatten`.
+    """
+    def _unflatten_from_numpy(subary):
+        if isinstance(subary, np.ndarray) and subary.dtype.char != "O":
+            subary = actx.from_numpy(subary)
```
inducer, line 511: What's the rule on whether numpy arrays get unflattened or not?

alexfikl: Right now it just goes through to `unflatten`, which fails if `ary.size` is different than `discr.ndofs`.

inducer: OIC, I misread this code. Thanks for explaining. (It seemed like it wouldn't go through `unflatten`, which would have seemed inconsistent to me.) All good here from my end, including good to merge.

alexfikl: It's still not great though, since `flatten` just lets non-`DOFArray` pass through, but `unflatten` chokes if it can't transform to a `DOFArray`.

Not sure what's the best way to handle it. Thoughts?

inducer: Maybe error out in `flatten` on leaves that aren't `DOFArray`s?

alexfikl: Yeah, that's what it's doing on main. pytential has that `flatten_if_needed` thing to handle all cases, mostly because it also has a lot of non-`DOFArray` things flying around (e.g. `PointsTarget`). 

Maybe go the other way and make `unflatten` more forgiving? Motivation being that it now iterates over entire containers and it's a bit restrictive to say "all your leaves must be DOFArrays".

inducer: I could see allowing a flag, on both sides.

alexfikl: Added a `strict=True` flag to both of them in c6c218f. Seems to work well on pytential's side too.

meshmode/dof_array.py
```diff
@@ -378,7 +378,41 @@ def wrapper(*args):
...
+            raise TypeError(f"non-DOFArray type '{type(ary).__name__}' cannot "
+                    "be flattened; use 'strict=False' to allow other types")
+        else:
+            return ary
+
+    actx = ary.array_context
+    if actx is None:
+        raise ValueError("cannot flatten frozen DOFArrays")
+
+    @memoize_in(actx, (flatten, "flatten_prg"))
```
inducer, line 393:
```suggestion
    @memoize_in(actx, (_flatten_dof_array, "flatten_prg"))
```

meshmode/dof_array.py
```diff
@@ -389,55 +423,36 @@ def flatten(ary: ArrayContainer) -> Any:
...
-                    = grp_ary[iel, idof]""",
-                name="flatten")
-
-        group_sizes = [grp_ary.shape[0] * grp_ary.shape[1] for grp_ary in subary]
-        group_starts = np.cumsum([0] + group_sizes)
-
-        result = actx.empty(group_starts[-1], dtype=subary.entry_dtype)
+    :param strict: if *True* only :class:`DOFArray`\ s are allowed as leaves
+        in the container *ary*. If *False* any non-:class:`DOFArray` are
+        left as is.
```
inducer, line 429:
```suggestion
    :param strict: if *True*, only :class:`DOFArray`\ s are allowed as leaves
        in the container *ary*. If *False*, any non-:class:`DOFArray` are
        left as is.
```

meshmode/dof_array.py
```diff
@@ -448,34 +463,97 @@ def prg():
...
+    element group in *discr*.
+
     :arg ndofs_per_element: if given, an iterable of numbers representing
         the number of degrees of freedom per element, overriding the numbers
         provided by the element groups in *discr*. May be used (for example)
         to handle :class:`DOFArray`\ s that have only one DOF per element,
         representing some per-element quantity.
+    :param strict: if *True* only :class:`DOFArray`\ s are allowed as leaves
+        in the container *ary*. If *False* any non-:class:`DOFArray` are
+        left as is.
```
inducer, line 502:
```suggestion
    :param strict: if *True*, only :class:`DOFArray`\ s are allowed as leaves
        in the container *ary*. If *False*, any non-:class:`DOFArray` are
        left as is.
```

meshmode/dof_array.py
```diff
@@ -389,55 +423,36 @@ def flatten(ary: ArrayContainer) -> Any:
...
+            raise ValueError("cannot unflatten array: "
+                    f"has size {ary.size}, expected {group_starts[-1]}; "
+                    "use 'strict=False' to leave the array unchanged")
+        else:
+            return ary
 
-def _unflatten(
-        actx: ArrayContext, group_shapes: Iterable[Tuple[int, int]], ary: Any
-        ) -> DOFArray:
     @memoize_in(actx, (unflatten, "unflatten_prg"))
```
inducer, line 449:
```suggestion
    @memoize_in(actx, (_unflatten_dof_array, "unflatten_prg"))
```

meshmode/dof_array.py
```diff
@@ -448,34 +463,97 @@ def prg():
...
+        ndofs_per_element_per_group: Optional[Iterable[int]] = None, *,
+        strict: bool = True,
+        ) -> ArrayContainer:
+    r"""Convert all "flat" arrays returned by :func:`flatten` back to
     :class:`DOFArray`\ s.
 
+    This function recurses into the :class:`~arraycontext.ArrayContainer` for all
+    :class:`DOFArray`\ s. It assumes that all the :class:`DOFArray`\ s inside the
+    container *ary* have the same number of degrees of freedom per
+    element group in *discr*.
```
inducer, line 493: Not sure I understand this sentence. Could you clarify?

inducer (commented):
Thanks! A few minor nits and one question here, then this is good to go.

meshmode/dof_array.py
alexfikl: For example, if you give it an object array of `[element_centroids, nodes]`, where `element_centroids` has 1 DOF per element and `nodes` has `nunit_nodes` DOFs per element. Then it won't know how to unflatten the container as a whole, because it just gets one `ndofs_per_element_per_group` to customize the number of DOFs per element.

```suggestion
    :class:`DOFArray`\ s. All class:`DOFArray`\ s inside the
    container *ary* must have the same number of degrees of freedom,
    as dictated by `ndofs_per_element_per_group`.
```
?

inducer, line 493:
```suggestion
    :class:`DOFArray`\ s. All class:`DOFArray`\ s inside the container 
    *ary* must agree on the mapping from element group number to number
    of degrees of freedom, as given by `ndofs_per_element_per_group`
    (or via *discr*).
```
Thanks for clarifying, I understand now. The original version sounded like all groups needed to have the same number of DOFs per element. I'll commit my version and merge this. Protests if you think I made it worse. :)

inducer: (Thread continued and resolved at https://github.com/inducer/meshmode/pull/196#discussion_r639932687.)

inducer:
LGTM. Thanks for working on this!

PR #197, opened 2021-05-27
Add wrapper around affine_map for rotation

alexfikl:
* Made `affine_map` also take numbers, to scale uniformly in all directions.
* Added a `rotate_mesh_around_axis`, which just constructs the rotation matrix and calls `affine_map`.


inducer:
Thanks!

PR #200, opened 2021-06-01
Add unflatten_like

alexfikl:
- [x] Should add a small test

### DIFF
--- a/meshmode/dof_array.py
+++ b/meshmode/dof_array.py
@@ -514,11 +514,17 @@ def unflatten(
     return rec_map_dof_array_container(_unflatten, ary)
 
 
-def unflatten_like(actx: ArrayContext,
-        ary: ArrayContainer,
-        prototype: ArrayContainer, strict: bool = True) -> ArrayContainer:
+def unflatten_like(
+        actx: ArrayContext, ary: ArrayContainer, prototype: ArrayContainer, *,
+        strict: bool = True,
+        ) -> ArrayContainer:
     r"""Convert all "flat" arrays returned by :func:`flatten` back to
     :class:`DOFArray`\ s based on a *prototype* container.
+
+    This function allows doing a roundtrip with :func:`flatten` for containers
+    which have :class:`DOFArray`\ s with different numbers of degrees of
+    freedom. This is unlike :func:`unflatten`, where all the :class:`DOFArray`\ s
+    must agree on the number of degrees of freedom per element group.
 
     :param prototype: an array container with the same structure as *ary*,
         whose :class:`DOFArray` leaves are used to get the sizes to
--- a/meshmode/dof_array.py
+++ b/meshmode/dof_array.py
@@ -525,6 +525,9 @@ def unflatten_like(
     which have :class:`DOFArray`\ s with different numbers of degrees of
     freedom. This is unlike :func:`unflatten`, where all the :class:`DOFArray`\ s
     must agree on the number of degrees of freedom per element group.
+    For example, this enables "unflattening" of arrays associated with different
+    :class:`~meshmode.discretization.Discretization`\ s within the same
+    container.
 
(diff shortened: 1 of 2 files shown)


meshmode/dof_array.py
```diff
@@ -513,6 +514,49 @@ def _unflatten(subary):
...
+            group_sizes = [subary.size for subary in _prototype]
+            group_starts = np.cumsum([0] + group_sizes)
+
+            return _unflatten_dof_array(
+                    actx, _ary, group_shapes, group_starts,
+                    strict=True)
+        elif is_array_container(_prototype):
+            return deserialize_container(_prototype, [
+                (key, _unflatten_like(subary, subprototype))
+                for (_, subary), (key, subprototype) in zip(
```
inducer, line 544: Should probably assert (maybe via a `_same_key(a, b)` function) that the keys match?

meshmode/dof_array.py
```diff
@@ -513,6 +514,49 @@ def _unflatten(subary):
     return rec_map_dof_array_container(_unflatten, ary)
 
 
+def unflatten_like(actx: ArrayContext,
+        ary: ArrayContainer,
+        prototype: ArrayContainer, strict: bool = True) -> ArrayContainer:
+    r"""Convert all "flat" arrays returned by :func:`flatten` back to
+    :class:`DOFArray`\ s based on a *prototype* container.
```
inducer, line 521: Explain the use case. I'm guessing the point is not having a discretization handy? Depending on the strength of the use case, I may or may not be OK with including this in the docs...

inducer (commented):
Thanks! Two hopefully quick comments below.

meshmode/dof_array.py
alexfikl: My particular use case is that I'm using some external optimization packages and need to flatten and unflatten between them (which is just dandy). This allows it to basically do `unflatten_like(current_iteration, initial_guess)` without having to carry a discretization around.

The reason I thought about upstreaming it is because it more nicely matches `flatten`, i.e. `flatten` can go through containers with `DOFArray`s that have different numbers of degrees of freedom, but `unflatten` needs them to be homogeneous.

I'm definitely fine with keeping this on my side though.

inducer: That's strong enough IMO to keep it documented. Could you add a summary to the docstring?

meshmode/dof_array.py
alexfikl: Added in 65a5423.

meshmode/dof_array.py
alexfikl: Added a couple of sentences in d2246a0. Let me know what you think.

meshmode/dof_array.py
```diff
@@ -513,6 +514,61 @@ def _unflatten(subary):
...
+        actx: ArrayContext, ary: ArrayContainer, prototype: ArrayContainer, *,
+        strict: bool = True,
+        ) -> ArrayContainer:
+    r"""Convert all "flat" arrays returned by :func:`flatten` back to
+    :class:`DOFArray`\ s based on a *prototype* container.
+
+    This function allows doing a roundtrip with :func:`flatten` for containers
+    which have :class:`DOFArray`\ s with different numbers of degrees of
+    freedom. This is unlike :func:`unflatten`, where all the :class:`DOFArray`\ s
+    must agree on the number of degrees of freedom per element group.
```
inducer, line 527:
```suggestion
    must agree on the number of degrees of freedom per element group.
    For example, this enables "unflattening" of arrays associated with different
    :class:`~meshmode.discretization.Discretization`\ s within the same
    container.
```

meshmode/dof_array.py
inducer: Thanks! Added another bit in ef95ea1. LGTM now.

inducer:
Thanks!

PR #204, opened 2021-06-02
Add mesh boundary gluing (aka periodic)

majosm:


majosm:
So I've managed to get periodicity at least partially working.

It seems to work fine with meshes for which the boundary transformation is just a translation, e.g.:

periodic

But, for more general transformations I'm seeing some issues (reflections, instability):

periodic1

Driver code for these cases is [here](https://github.com/majosm/mirgecom/blob/1ca1901da841fc1bcebef7f86a26aa05446f6a92/examples/wave-periodic.py).

As far as I can tell, the transformations are implemented correctly (the Gauss-Newton stuff in `_make_cross_face_batches` is converging, at least). I *think* what's causing this is the fact that for most of our operators, we need to exchange fluxes that are vectors; so in the general transformation case we would need to transform these vectors accordingly when going from one periodic boundary to the other, right? I'm not entirely sure how we could go about doing that, or if that's something we could even realistically do. Any thoughts @inducer?

inducer:
Oh cool. Yes, you're definitely right that the numerical flux would need to play a role in the case of more general transformations. (As in, if you give a numerical flux on the + side a normal that's orthogonal to the one on the - side, it's not a huge surprise that nothing good will come of that. I'm kind of surprised you got the simulation to survive as long as you did. :) Making this work right will require some design decisions, such as how to expose such an interface to the numerical flux. The easiest thing would be to expose the coordinate transform in a trace pair and separate out the periodic part of the boundary in a separate one. If we *don't* separate out that flux exchange, then we'd be paying to apply a pointless identity transform at every flux exchange point. I'm having a hard time guessing whether that would be a big deal in terms of cost.

At any rate, my suggestion would be to devote the rest of this PR to get this merged as is, and to worry about those design decisions in a separate issue. (I'll create one.)

cc @thomasgibson @lukeolson for input

meshmode/discretization/connection/opposite_face.py
```diff
@@ -41,7 +41,8 @@ def thaw_to_numpy(actx, array):
 def _make_cross_face_batches(actx,
         tgt_bdry_discr, src_bdry_discr,
         i_tgt_grp, i_src_grp,
-        tgt_bdry_element_indices, src_bdry_element_indices):
+        tgt_bdry_element_indices, src_bdry_element_indices,
+        tgt_transforms=None, src_transforms=None):
```
inducer, line 45: Rename these `tgt_aff_transforms` or some such?

meshmode/mesh/__init__.py
```diff
@@ -468,6 +468,18 @@ class FacialAdjacencyGroup(Record):
 
         Zero if ``neighbors[i]`` is negative.
 
+    .. attribute:: transform_mats
```
inducer, line 471: Something something affine.

inducer (commented):
Thanks! Some review bits that came up during our meeting.

inducer:

majosm:
Question that just came up while talking to @thomasgibson: what should happen if someone tries to call `map_mesh` on a mesh that has nontrivial transforms in its facial adjacency?

(I haven't thought about this at all yet; just putting it here so I don't forget.)

majosm:
I think this is ready for a look @inducer. (There's still a test failure, but it appears to be a pylint false positive?)

meshmode/mesh/generation.py
```diff
@@ -69,6 +69,8 @@
 .. autofunction:: generate_box_mesh
 .. autofunction:: generate_regular_rect_mesh
 .. autofunction:: generate_warped_rect_mesh
+.. autofunction:: generate_annular_cylinder_slice_mesh
```
thomasgibson, line 72: Can we get a `generate_periodic_rect_mesh` pretty please? :smiley:

meshmode/discretization/connection/opposite_face.py
```diff
@@ -42,7 +42,8 @@ def thaw_to_numpy(actx, array):
 def _make_cross_face_batches(actx,
         tgt_bdry_discr, src_bdry_discr,
         i_tgt_grp, i_src_grp,
-        tgt_bdry_element_indices, src_bdry_element_indices):
+        tgt_bdry_element_indices, src_bdry_element_indices,
+        tgt_aff_transforms=None, src_aff_transforms=None):
```
inducer, line 46: Add docstring? (Shape of `src_aff_transforms` is how I got there.)

majosm:
Superseded by #253.

PR #210, opened 2021-06-06
Try to reproduce NaN in test_flatten_unflatten

inducer:


inducer:
Hmm. I observed some NaNs in `test_flatten_unflatten` on 6e0dffc6f358c62e6be35d1016a62a1106e593a0. I ran the tests in a loop on this PR, it never recurred. Very strange, but I'll stop digging.

PR #213, opened 2021-06-08
Fix affine map with only translation

alexfikl:
Managed to introduce a nice bug in #197 :(


alexfikl:
Ha! Found that `NaN` again!

inducer:
Thanks!

PR #116, opened 2021-02-12
Pytato ArrayContext

kaushikcfd:

PR #216, opened 2021-06-08
Enable simple-dg to run with lazy eval

matthiasdiener:
Followup of #116 with PytatoArrayContext from https://github.com/inducer/arraycontext/pull/14. 

cc @kaushikcfd 

Needs:
- [x] After https://github.com/inducer/arraycontext/pull/14 is merged, change requirements.txt back
- [x] Test lazy eval simple-dg
- ~~[ ] add `transform_loopy_program`~~ Edit: defer until later


examples/simple-dg.py
```diff
@@ -198,7 +198,7 @@ def zeros(self, actx):
         return self.volume_discr.zeros(actx)
 
     def grad(self, vec):
-        ipder = self.inverse_parametrization_derivative()
+        ipder = thaw(self.inverse_parametrization_derivative(), self._setup_actx)
```
inducer, line 201: That's not correct. `_setup_actx` is only for setup tasks. In this case, you should fish the array context out of `vec`.

examples/simple-dg.py
```diff
@@ -467,7 +474,7 @@ def main():
 
     cl_ctx = cl.create_some_context()
     queue = cl.CommandQueue(cl_ctx)
-    actx = PyOpenCLArrayContext(queue)
+    actx = PytatoArrayContext(queue)
```
inducer, line 477: Ideally it should run both.

meshmode/discretization/__init__.py
```diff
@@ -576,7 +577,10 @@ def prg():
                     result[iel, idof] = \
                         sum(j, resampling_mat[idof, j] * nodes[iel, j])
                     """,
-                name="nodes")
+                name="lp_nodes")
+            prg = lp.set_options(
+                prg, enforce_array_accesses_within_bounds="no_check")
```
inducer, line 582: Could you explain why that's needed?

meshmode/discretization/__init__.py
```diff
@@ -591,8 +595,11 @@ def resample_mesh_nodes(grp, iaxis):
                 return nodes
 
             return actx.call_loopy(
-                    prg(),
+                    lp.fix_parameters(prg(),
+                            ndiscr_nodes=grp.nunit_dofs,
+                            nmesh_nodes=grp.mesh_el_group.nunit_nodes),
```
inducer, line 600: Could you explain why that's needed?

meshmode/discretization/__init__.py
kaushikcfd: Access check mapper need to traverse the call-graph, which it doesn't do yet: https://github.com/inducer/loopy/issues/402

meshmode/discretization/__init__.py
kaushikcfd: Figuring out those argument would need some shape inference in `call_loopy`, that's also not yet implemented. Currently, pytato screams saying it didn't receive all the input arguments. That's the reason `call_loopy` needs to pass all arguments.

examples/simple-dg.py
kaushikcfd: Done, see 7e02ae0.

inducer:
It's super sweet to see this come together. Awesome work, both of you!

PR #217, opened 2021-06-08
Revert connection kernels to pre-#192

thomasgibson:
So the source of the slow-down @MTCam and @anderson2981 observed can basically be traced back to the connections. Reverting the changes applied by #192 in the connection kernels (didn't revert everything) appears to eliminate the large drop in speed.

I know these aren't "lazy compatible" but I think time should be spent figuring out a robust way to do these connections, that doesn't have really bad performance penalties.

### DIFF
--- a/meshmode/discretization/connection/direct.py
+++ b/meshmode/discretization/connection/direct.py
@@ -295,24 +295,40 @@ class DirectDiscretizationConnection(DiscretizationConnection):
 
         actx = ary.array_context
 
-        @memoize_in(actx,
-                (DirectDiscretizationConnection, "resample_by_mat_batch_knl"))
-        def batch_mat_knl():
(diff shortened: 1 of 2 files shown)


inducer (approved):
I'm OK with this. I honestly think that this might be an easy fix: it don't think that `iel_init` (and `idof_init`) are tagged for GPU parallelism, which would certainly slow things down. If you'd like to merge this and then fix things up in peace, that's fine by me.

inducer:
Ah, whatever. I'll just hit automerge. Then we can sort this out in peace, and everyone else can continue working.

inducer:
Also, thanks for figuring this out!

MTCam:
Yes, thanks for your help and patience figuring this out, and for fixing it up so quickly!

PR #220, opened 2021-06-13
Undo revert of connection kernels to pre-#192

inducer:
This reverts #217 and thereby makes meshmode lazy-compatible again. 

In an earlier version of this, I said:

> There should not be a performance hit this time.

Turns out that that statement was a bit optimistic. See below.

- [x] Needs https://github.com/inducer/loopy/pull/410
- [x] ~~Needs https://github.com/inducer/arraycontext/pull/29 (for now, until meshmode gets its own `PyOpenCLArrayContext`)~~
- [x] `requirements.txt` should be pointed back to main

cc @thomasgibson @matthiasdiener

### DIFF
--- a/.ci/install-for-firedrake.sh
+++ b/.ci/install-for-firedrake.sh
@@ -25,9 +25,9 @@ pip install dataclasses
 
 pip install pytest
 
-pip install -r /tmp/myreq.txt
+pip install -r /tmp/myreq.txt 
 
-# Use updated loopy for test
-python -m pip install --force-reinstall git+https://github.com/inducer/loopy.git
+# Context: https://github.com/OP2/PyOP2/pull/605
+python -m pip install --force-reinstall git+https://github.com/inducer/pytools.git
 
 pip install .
--- a/.ci/install-for-firedrake.sh
+++ b/.ci/install-for-firedrake.sh
@@ -25,9 +25,9 @@ pip install dataclasses
 
 pip install pytest
 
-pip install -r /tmp/myreq.txt 
+pip install -r /tmp/myreq.txt
 
-# Context: https://github.com/OP2/PyOP2/pull/605
-python -m pip install --force-reinstall git+https://github.com/inducer/pytools.git
+# Use updated loopy for test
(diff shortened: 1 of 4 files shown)


inducer:
The Firedrake failures should resolve once they pull in main and build a container: https://github.com/firedrakeproject/loopy/issues/16

inducer:
@thomasgibson and I worked on the performance impact of this for a bit just now. As a summary, for some benchmark based on `simple-dg.py` with 512 elements:

* fdf1f1d (`main`): 11.1s
* dc7215d: 13.3s
* 6a0f9cf: 12.5s

So while 6a0f9cf helps, this isn't (yet) a change that comes for free in the eager world. From my perspective, the main cost factors are the fact that we're filling a  whole array for each `InterpolationBatch` and then summing them afterwards:

https://github.com/inducer/meshmode/blob/6a0f9cffcc941ce6c298053e500332810f374f12/meshmode/discretization/connection/direct.py#L450

An option that's available to us is to ask the array context whether it allows in-place modification, and if so, use it. (This attribute would have to be exposed on the array context, and it seems like a legitimate thing to ask of it.) This comes at the cost of essentially maintaining the old and the new connection kernels alongside each other. They're small though, so I think maybe that's a cost worth bearing.

inducer:
@nchristensen Since we just discussed removing the write-side indirect access in our meeting today: Code for this is in 6a0f9cf.

inducer:
8781160 adds back the old codepath with in-place updates, based on https://github.com/inducer/arraycontext/pull/37. This eliminates the runtime cost of this for eager eval.

03d883a and 39544de revert to the current transform, name-of-iname-based transform strategy for the moment. This has a performance impact during setup, as described in this FIXME:
https://github.com/inducer/meshmode/blob/e179d357a283269b19b36af29487004d3873252a/meshmode/discretization/connection/direct.py#L126-L132
In my experiments, the cost increase isn't even very noticeable. That means that this no longer depends on any fancy new arraycontext technology, and it's ready to merge now.

PR #223, opened 2021-06-15
Add relative path support for visualization file names.

MTCam:
This adds relative file path support for visualization file names.

Without this change, specifying relative filenames for the visualization file works OK, except for the filenames listed in the parallel manifest, which had previous retained the relative filenames.  This change massages the filenames listed in the parallel manifest so as to drop the relative pathing.

### DIFF
--- a/meshmode/discretization/visualization.py
+++ b/meshmode/discretization/visualization.py
@@ -641,8 +641,9 @@ class Visualizer:
                 par_manifest_filename=par_manifest_filename,
                 par_file_names=[
                     file_name_pattern.format(rank=rank)
-                    for rank in range(nranks)]
-        )
+                    for rank in range(nranks)
+                    ]
+                )
 
     def write_vtk_file(self, file_name, names_and_fields,
             compressor=None, real_only=False, overwrite=False,
--- a/meshmode/discretization/visualization.py
+++ b/meshmode/discretization/visualization.py
@@ -807,7 +807,7 @@ class Visualizer:
 
             if responsible_for_writing_par_manifest:
                 parfile_relnames = [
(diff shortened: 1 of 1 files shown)


meshmode/discretization/visualization.py
```diff
@@ -639,7 +641,7 @@ def write_parallel_vtk_file(self, mpi_comm, file_name_pattern, names_and_fields,
                 use_high_order=use_high_order,
                 par_manifest_filename=par_manifest_filename,
                 par_file_names=[
-                    file_name_pattern.format(rank=rank)
+                    os.path.basename(file_name_pattern.format(rank=rank))
```
inducer, line 644: AFAICT, this breaks the logic that currently determines which process is responsible for writing the manifest:
https://github.com/inducer/meshmode/blob/9058582b695091f74714acc2636339f53b56c4e9/meshmode/discretization/visualization.py#L771-L774
according to which `file_name` must be one of the `par_file_names`.

A better fix IMO would be to change the `par_file_names` right before inserting them into the manifest:
https://github.com/inducer/meshmode/blob/9058582b695091f74714acc2636339f53b56c4e9/meshmode/discretization/visualization.py#L771-L774
to be relative to the manifest path.

Python has [`os.path.relpath`](https://github.com/inducer/meshmode/blob/9058582b695091f74714acc2636339f53b56c4e9/meshmode/discretization/visualization.py#L771-L774) just for that purpose.

MTCam: Did you intend to post the same exact snippet twice here?

MTCam: Ok, I think this is fixed up how you wanted it now.

inducer: LGTM, thanks!

PR #225, opened 2021-06-16
Test that connections that should be permutations actually *are* permutations

inducer:
This is supposed to be a version of #105 that we can merge right away, plus a number of fixes that arose around that. For review, this PR is probably best viewed commit-by-commit. The commit messages provide a pretty good inventory of what's here.

Diff-wise, the biggest part of this is the (annoying, but IMO required) split of `PolynomialWarpAndBlendGroupFactory` into `PolynomialWarpAndBlend2DGroupFactory` and `PolynomialWarpAndBlend3DGroupFactory`. (The 2D WnB nodes are *not* equivalent to a restriction of the 3D nodes to a face, and we very much would like 2D el groups in 3D discretizations to be just that, so that the boundary extractions can be done by indirect addressing.)

- [ ] Post-merge, this needs a corresponding (and hopefully much smaller) change in grudge.

cc @nchristensen

### DIFF
--- a/meshmode/discretization/connection/direct.py
+++ b/meshmode/discretization/connection/direct.py
@@ -22,6 +22,7 @@ THE SOFTWARE.
 
 
 import numpy as np
+import numpy.linalg as la
 
 import loopy as lp
 from pytools import memoize_in, keyed_memoize_method
@@ -250,28 +251,33 @@ class DirectDiscretizationConnection(DiscretizationConnection):
         :class:`pyopencl.array.Array` containing the index subset.
         """
 
-        mat = actx.to_numpy(actx.thaw(
-                self._resample_matrix(actx, to_group_index, ibatch_index)))
-
-        nrows, ncols = mat.shape
-        result = np.zeros(nrows, dtype=self.to_discr.mesh.element_id_dtype)
+        ibatch = self.groups[to_group_index].batches[ibatch_index]
+        from_grp = self.from_discr.groups[ibatch.from_group_index]
 
         if tol_multiplier is None:
             tol_multiplier = 50
 
-        tol = np.finfo(mat.dtype).eps * tol_multiplier
+        tol = np.finfo(ibatch.result_unit_nodes.dtype).eps * tol_multiplier
 
-        for irow in range(nrows):
-            one_indices, = np.where(np.abs(mat[irow] - 1) < tol)
-            zero_indices, = np.where(np.abs(mat[irow]) < tol)
+        dim, ntgt_nodes = ibatch.result_unit_nodes.shape
+        if dim == 0:
+            assert ntgt_nodes == 1
+            return actx.freeze(actx.from_numpy(np.array([0], dtype=np.int32)))
 
-            if len(one_indices) != 1:
+        dist_vecs = (ibatch.result_unit_nodes.reshape(dim, -1, 1)
+                - from_grp.unit_nodes.reshape(dim, 1, -1))
+        dists = la.norm(dist_vecs, axis=0, ord=2)
+
+        result = np.zeros(ntgt_nodes, dtype=self.to_discr.mesh.element_id_dtype)
+
+        for irow in range(ntgt_nodes):
+            close_indices, = np.where(dists[irow] < tol)
+
+            if len(close_indices) != 1:
                 return None
-            if len(zero_indices) != ncols - 1:
-                return None
 
-            one_index, = one_indices
-            result[irow] = one_index
+            close_index, = close_indices
+            result[irow] = close_index
 
         return actx.freeze(actx.from_numpy(result))
 
--- a/meshmode/discretization/poly_element.py
+++ b/meshmode/discretization/poly_element.py
@@ -53,7 +53,8 @@ Simplicial group types
 
 .. autoclass:: InterpolatoryQuadratureSimplexElementGroup
 .. autoclass:: QuadratureSimplexElementGroup
-.. autoclass:: PolynomialWarpAndBlendElementGroup
+.. autoclass:: PolynomialWarpAndBlend2DRestrictingElementGroup
+.. autoclass:: PolynomialWarpAndBlend3DRestrictingElementGroup
 .. autoclass:: PolynomialRecursiveNodesElementGroup
 .. autoclass:: PolynomialEquidistantSimplexElementGroup
 .. autoclass:: PolynomialGivenNodesElementGroup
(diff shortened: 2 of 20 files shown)


meshmode/discretization/connection/direct.py
```diff
@@ -250,28 +250,36 @@ def _resample_point_pick_indices(self, actx: ArrayContext,
...
-                return None
-            if len(zero_indices) != ncols - 1:
+        dist_vecs = (ibatch.result_unit_nodes.reshape(dim, -1, 1)
+                - from_grp.unit_nodes.reshape(dim, 1, -1))
+        dists = np.sqrt(np.sum(dist_vecs**2, axis=0))
+
+        result = np.zeros(ntgt_nodes, dtype=self.to_discr.mesh.element_id_dtype)
+
+        if tol_multiplier is None:
+            tol_multiplier = 500
```
alexfikl, line 273: There's another `tol_multiplier is None` above that sets it to 50. Which one did you mean to keep?

meshmode/discretization/connection/direct.py
```diff
@@ -250,28 +250,36 @@ def _resample_point_pick_indices(self, actx: ArrayContext,
...
+        if dim == 0:
+            assert ntgt_nodes == 1
+            return actx.freeze(actx.from_numpy(np.array([0], dtype=np.int32)))
 
-            if len(one_indices) != 1:
-                return None
-            if len(zero_indices) != ncols - 1:
+        dist_vecs = (ibatch.result_unit_nodes.reshape(dim, -1, 1)
+                - from_grp.unit_nodes.reshape(dim, 1, -1))
+        dists = np.sqrt(np.sum(dist_vecs**2, axis=0))
```
alexfikl, line 268:
```suggestion
        dists = la.norm(dist_vecs, axis=0, ord=2)
```
? Don't know if this is any faster.

meshmode/discretization/poly_element.py
```diff
@@ -287,6 +289,41 @@ class PolynomialWarpAndBlendElementGroup(_MassMatrixQuadratureElementGroup):
     interpolation. Interpolation nodes edge-clustered for avoidance of Runge
     phenomena. Nodes are present on the boundary of the simplex.
 
+    Uses :func:`modepy.warp_and_blend_nodes`.
+    """
+    def __init__(self, mesh_el_group, order, index):
```
alexfikl, line 294: I'm not sure how I feel about this splitting of the warp and blend classes.

A couple of random thoughts:
* If I'm understanding the issue correctly, this can also be solved by just introducing the 3D version and being careful to pass it to `make_face_restriction`? Is that too much to leave to the user to get right?
* How about a flag in the original `PolynomialWarpAndBlendElementGroup` like `restricted_from_3d=None` (that gets set to `True` or `False` based on dimension)?

meshmode/discretization/poly_element.py
```diff
@@ -304,6 +344,35 @@ def _interp_nodes(self):
         return result
 
 
+class PolynomialWarpAndBlend3DElementGroup(_MassMatrixQuadratureElementGroup):
```
alexfikl, line 347: Feels like this should have `From3DRestriction` or something to that effect in the name? This makes it sound like it only works in 3D.

meshmode/interop/firedrake/connection.py
```diff
@@ -39,9 +39,8 @@
 from meshmode.mesh.processing import get_simplex_element_flip_matrix
 
 from meshmode.discretization.poly_element import (
-    PolynomialWarpAndBlendGroupFactory,
-    PolynomialRecursiveNodesGroupFactory,
-    ElementGroupFactory)
+        default_simplex_group_factory,
+        ElementGroupFactory)
```
alexfikl, line 43:
```suggestion
    default_simplex_group_factory,
    ElementGroupFactory)
```
? Everyone else seems to be indented like this.

meshmode/discretization/poly_element.py
```diff
@@ -304,6 +344,35 @@ def _interp_nodes(self):
...
+    """
+    @property
+    @memoize_method
+    def _interp_nodes(self):
+        dim = self.mesh_el_group.dim
+        if self.order == 0:
+            result = mp.warp_and_blend_nodes(dim, 1)
+            result = np.mean(result, axis=1).reshape(-1, 1)
+        else:
+            result = mp.warp_and_blend_nodes(3, self.order)
```
alexfikl, line 365: How nice are the restrictions of the 3D nodes? In terms of Lebesgue constants and things like that.

alexfikl (approved):
Not quite sure about the split of the warp and blend classes, but functionally it looks good to me! (barring a few nitpicks)

meshmode/discretization/poly_element.py
inducer: Fair point.

* `PolynomialWarpAndBlend3DRestrictingElementGroup`?
* `PolynomialWarpAndBlend2DRestrictingElementGroup`?

alexfikl: Yep, that sounds better!

meshmode/discretization/poly_element.py
inducer: > I'm not sure how I feel about this splitting of the warp and blend classes.

I can see where you're coming from. But to me they simply are two mostly unrelated sets of nodes. If we distinguished them by a flag in `PolynomialWarpAndBlendElementGroup`, it would still be like two functions wedged into one with an `if` statement.

And the original `PolynomialWarpAndBlendElementGroup` is just not worth having: For DG, it'll automatically be inefficient.

> Is that too much to leave to the user to get right?

If you'd like to add a special `if` statement for this into `make_face_restriction`, maybe not. Otherwise, I think it's hard to get right: It's totally legitimate to ask for any old node set you like, and if we don't do *something* like the split I'm proposing, you almost automatically end up with 3D->2D nodes that unnecessarily require interpolation.

> How about a flag in the original PolynomialWarpAndBlendElementGroup like restricted_from_3d=None (that gets set to True or False based on dimension)?

To my mind, the issue is that the factory has no way of telling that a 2D group is descended from a 3D group when it creates it. We could set state in the factory based on the first group created, but find that too ugly/problematic to seriously consider.

meshmode/discretization/poly_element.py
inducer: https://gist.github.com/inducer/bbe9fd6e32d5e67b3d4e446479cfa633

tl;dr: Negligibly different IMO.

meshmode/discretization/poly_element.py
alexfikl: Yeah, I'm pretty sold on the fact that they're basically two different family of nodes and they both have to be there.

With the better naming, I'm happy to close this random rant!

meshmode/discretization/poly_element.py
alexfikl: Nice! :tada:

meshmode/interop/firedrake/connection.py
inducer: Gone in 8a822fe.

meshmode/discretization/poly_element.py
inducer: 8a822fe

meshmode/discretization/connection/direct.py
inducer: ba86136

meshmode/discretization/connection/direct.py
inducer: ba86136

PR #226, opened 2021-06-16
_make_cross_face_batches: refactor into multiple functions, try matching nodes exactly before going to Gauss-Newton

inducer:
While it doesn't actually solve the problem described in https://github.com/inducer/meshmode/pull/105#issuecomment-862003829, this should work around the main impact of #105: The DG examples no longer use matvecs to interpolate onto the boundary. This should help speed up larger-scale DG runs somewhat.

Draft because:
- [x] Should merge after #225, specifically https://github.com/inducer/meshmode/pull/225/commits/3d2f376ca69c1de83e13b9d7aaaa6c416b0ef07b (`test_mesh_multiple_groups`: Test inhomogeneous polynomial degree), because otherwise the Gauss-Newton code path would likely be untested.
- [x] After #225 is pulled into this, should update `test_bdry_restriction_is_permutation` to remove the `xfail`s. All the tests should then be passing: cbbca40

Post-merge:
- [ ] Update #105 as described in https://github.com/inducer/meshmode/pull/105#issuecomment-862703485

@majosm Could you review this? I know it'll likely have a little bit of conflict with #204, but `_make_cross_face_batches` was getting crazy long. Hopefully the conflicts won't be bad: the diff doesn't touch any of the actual functionality, it just shuffles function arguments around.

cc @nchristensen @matthiasdiener

### DIFF
--- a/meshmode/discretization/connection/opposite_face.py
+++ b/meshmode/discretization/connection/opposite_face.py
@@ -23,6 +23,7 @@ THE SOFTWARE.
 
 import numpy as np
 import numpy.linalg as la
+from meshmode.discretization.connection.direct import InterpolationBatch
 
 import logging
 logger = logging.getLogger(__name__)
@@ -43,15 +44,13 @@ def _make_cross_face_batches(actx,
         i_tgt_grp, i_src_grp,
         tgt_bdry_element_indices, src_bdry_element_indices):
 
-    from meshmode.discretization.connection.direct import InterpolationBatch
     if tgt_bdry_discr.dim == 0:
-        yield InterpolationBatch(
+        return [InterpolationBatch(
             from_group_index=i_src_grp,
             from_element_indices=freeze_from_numpy(actx, src_bdry_element_indices),
             to_element_indices=freeze_from_numpy(actx, tgt_bdry_element_indices),
             result_unit_nodes=src_bdry_discr.groups[i_src_grp].unit_nodes,
-            to_element_face=None)
-        return
+            to_element_face=None)]
 
     tgt_bdry_nodes = np.array([
         thaw_to_numpy(actx, ary[i_tgt_grp])[tgt_bdry_element_indices]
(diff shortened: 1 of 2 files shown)


meshmode/discretization/connection/opposite_face.py
```diff
@@ -68,11 +67,74 @@ def _make_cross_face_batches(actx,
...
+    is_close = np.sqrt(np.sum(dist_vecs**2, axis=0)) < tol
+
+    num_close_vertices = np.sum(is_close.astype(np.int32), axis=-1)
+    if not (num_close_vertices == 1).all():
+        return None
+
+    # Success: it's just a permutation
+    idx = np.empty_like(is_close, dtype=np.int32)
+    idx[:] = np.arange(src_grp.nunit_dofs).reshape(1, 1, -1)
+    source_indices = idx[is_close].reshape(nelements, ntgt_unit_nodes)
```
majosm, line 113:
```suggestion
    source_indices = np.where(is_close)[-1].reshape(nelements, ntgt_unit_nodes)
```
(I think?)

meshmode/discretization/connection/opposite_face.py
```diff
@@ -68,11 +67,74 @@ def _make_cross_face_batches(actx,
...
+
+    # check
+    matched_src_bdry_nodes = src_bdry_nodes[
+            :, np.arange(nelements).reshape(-1, 1), source_indices]
+    dist_vecs = tgt_bdry_nodes - matched_src_bdry_nodes
+    is_close = np.sqrt(np.sum(dist_vecs**2, axis=0)) < tol
```
majosm, line 119:
```suggestion
    is_close = np.sum(dist_vecs**2, axis=0) < tol**2
```
?

majosm (approved):
Looks good to me. I threw in a few optional suggestions.

meshmode/discretization/connection/opposite_face.py
```diff
@@ -68,12 +67,75 @@ def _make_cross_face_batches(actx,
...
+
+
+# {{{ _find_src_unit_nodes_by_matching
+
+def _find_src_unit_nodes_by_matching(
+        tgt_bdry_nodes,
+        src_bdry_nodes,
+        src_grp, tol):
+    ambient_dim, nelements, ntgt_unit_nodes = tgt_bdry_nodes.shape
+
```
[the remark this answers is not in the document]
inducer: I like the thinking, but the criterion your propose is actually too restrictive. The comparison thing is also supposed to work in the (ideally, common, see #225) case of face restrictions where the face nodes are a subset of the volume nodes.

meshmode/discretization/connection/opposite_face.py
```diff
@@ -68,12 +67,75 @@ def _make_cross_face_batches(actx,
...
+    dist_vecs = (tgt_bdry_nodes.reshape(ambient_dim, nelements, -1, 1)
+            - src_bdry_nodes.reshape(ambient_dim, nelements, 1, -1))
```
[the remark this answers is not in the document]
inducer: Yeah... it is clearly an O(nunit_nodes^2) algorithm, my hope is just that numpy crushes it. Anything with a Python loop will be way slower. I *am* a bit worried about the size of that temporary. If it's too big,we can always do a Python loop for one of those axes.

meshmode/discretization/connection/opposite_face.py
```diff
@@ -68,12 +67,75 @@ def _make_cross_face_batches(actx,
...
+
+    # shape: (nelements, num_tgt_nodes, num_source_nodes)
+    is_close = np.sqrt(np.sum(dist_vecs**2, axis=0)) < tol
```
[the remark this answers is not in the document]
inducer: 74e33b9. Used `la.norm` after all... more descriptive.

meshmode/discretization/connection/opposite_face.py
inducer: Whoa, thanks! You can probably tell that my solution took me a while to construct. This is much better. 74e33b9

meshmode/discretization/connection/opposite_face.py
inducer: 74e33b9. Used `la.norm` after all... more descriptive.

inducer:
@majosm Thanks for taking a look!

PR #233, opened 2021-06-22
Makes array context routine invocation independent of the ArrayContext type

kaushikcfd:


meshmode/dof_array.py
```diff
@@ -667,7 +667,9 @@ def flat_norm(ary, ord=None) -> float:
         actx = ary.array_context
         return la.norm(
                 [
-                    actx.np.linalg.norm(actx.np.ravel(subary, order="A"), ord=ord)
+                    actx.to_numpy(actx.np.linalg.norm(actx.np.ravel(subary,
```
kaushikcfd, line 670: Looks like callers assume that the result of `flat_norm` is a cl_array, but I think that is a violation of `def flat_norm(ary, ord=None) -> float`. I'm tempted to fix those tests assuming this was a bug in `flat_norm`. Opinions?

alexfikl: I thought `flat_norm` always returned host scalar. Which caller expected a CL array?

Is the above supposed to work? `actx.np.linalg.norm` also returns a number, but `actx.to_numpy` doesn't handle numbers (at least not in the `PyOpenCLArrayContext`).

alexfikl: https://github.com/inducer/arraycontext/blob/14efb7c2f4e48844aa496899eab4e6955ea09164/arraycontext/impl/pyopencl/__init__.py#L135-L136

kaushikcfd: Oops sorry. Yep, you're right. It's the other way around. `actx.np.linalg.norm` for PyOpenCLArrayContext  is a host scalar.
That's slightly worrying because it's doing an implicit freeze. I guess the options are either to make `pycl_actx.np.linalg.norm` return a device scalar or accept a host scalar in to_numpy.

(Thanks for the correction!)

alexfikl: Yeah, I imagine all reductions will have to be made into device scalars at some point (inducer/arraycontext#40), but that's going to be a pretty big breaking change for everyone :(

inducer: In the long run, I think even `flat_norm` should return a device scalar.

Letting `to_numpy` tolerate host scalars is reasonable IMO. It's already in https://github.com/inducer/arraycontext/pull/14:

https://github.com/inducer/arraycontext/blob/b2bd3dfa7cb78cda66e2f092be243dec9d514996/arraycontext/impl/pyopencl/__init__.py#L137-L138

We could just adopt that now. About the compat break: I agree that it stinks, but I don't know what else to do. Sticking to host scalars imposes a large penalty on lazy, and it's also not entirely reasonable for CL IMO... I'd like to make it a smooth transition if we can, but the flag thing from https://github.com/inducer/arraycontext/issues/40 is the best idea I've had so far.)

alexfikl: Gave it a go in inducer/arraycontext#43, which should fix this too.

meshmode/dof_array.py
```diff
@@ -423,9 +423,9 @@ def prg():
     def _flatten(grp_ary):
         # If array has two axes, assume they are elements/dofs. If C-contiguous
         # in those, "flat" and "unflat" memory layout agree.
-        if len(grp_ary.shape) == 2 and grp_ary.flags.c_contiguous:
-            return grp_ary.reshape(-1, order="C")
-        else:
+        try:
```
inducer, line 426: This isn't correct: The precondition should stay. For pytato's sake, it's OK to try accessing `.flags.c_contiguous`, and if it doesn't exist (and if there are only two axes), ravel in C order

kaushikcfd: > This isn't correct: 

Is the concern that `ValueError` isn't precise enough?

inducer: No, it's that array contexts are (supposed to) be free to choose the memory layout of their arrays. @nchristensen makes use of this for example and lays out the group arrays of a `DOFArray` in "column-major" order, which he determined works well for his GPU kernels.

kaushikcfd: Sorry, I'm not quite sure I still understand. I'm not proposing to change the layout ordering of the flattened array in this patch. But, instead just rewriting the flattening to C-order in a different manner.

IIUC, the only reason we had the precondition was to step-over pyopencl's rule that reshapes can't trigger copies.

inducer: > IIUC, the only reason we had the precondition was to step-over pyopencl's rule that reshapes can't trigger copies.

An array context can decide (for example) to lay out group arrays in a DOF-major (rather than the normal element-major) order. In that case, `flatten` necessarily involves a (transpose+)copy. Just reshaping those group arrays will wind up with the wrong thing, because unlike the group array, flattened arrays have a defined (element-major) memory layout.

Does that help?

kaushikcfd: > Just reshaping those group arrays will wind up with the wrong thing, because unlike the group array, flattened arrays have a defined (element-major) memory layout.

But we are reshaping with `order="C"`, wouldn't that ensure that the flattened out group arrays is emitted in an element-major order?

inducer: Fair enough. I think the real issue is that I haven't really figured out for myself how "wild' group arrays are allowed to be (tensor product elements? padded batches of elements?), and how this would interact with flattening (which is, effectively, the only interface with the "outside world"). My complaint didn't have a basis in fact otherwise. Your current approach looks OK to me.

meshmode/dof_array.py
inducer: Look at us: barely five days later, we're well on our way to everyone returning device scalars! Thanks @alexfikl for helping to make that happen!

meshmode/dof_array.py
```diff
@@ -675,7 +675,9 @@ def flat_norm(ary, ord=None) -> float:
         actx = ary.array_context
         return _scalar_list_norm(
                 [
-                    actx.np.linalg.norm(actx.np.ravel(subary, order="A"), ord=ord)
+                    actx.to_numpy(actx.np.linalg.norm(actx.np.ravel(subary,
+                                                                    order="A"),
+                                                      ord=ord))
```
inducer, line 680:
```suggestion
                    actx.np.linalg.norm(actx.np.ravel(subary, order="A"), ord=ord)
```

meshmode/dof_array.py
kaushikcfd: Thanks a lot @alexfikl! :)

meshmode/dof_array.py
kaushikcfd: This turned out to be unnecessary as it was already fixed on `main`.

inducer:
Thanks!

PR #234, opened 2021-06-23
Test, fix pickling of DOFArrays

inducer:


MTCam (approved):
Wow, what?

meshmode/dof_array.py
```diff
@@ -280,7 +280,7 @@ def __setstate__(self, state):
             raise RuntimeError("DOFArray instances can only be unpickled while "
                     "array_context_for_pickling is active.")
 
-        self.array_context = actx
+        self._array_context = actx
```
inducer, line 283: > Wow, what?

@MTCam What had happened is that `self.array_context` became a read-only property in 561fa46ac6ad2f70334bcd5d476b52ab7bd015c7, with the actual data living in `self._array_context`. I'm not sure I remember why it was done (@alexfikl?), but there's certainly nothing wrong with doing it that way. The main issues were that
- that change broke the pickling code
- the pickling code had no tests :scream: (my bad!), which means there wasn't a way for anyone to notice what had happened.

alexfikl: :facepalm: Yeah, that's my bad. If I remember correctly, this came out of early #175 when we still had that `ArrayContainerWithArithmetic` subclass that required an `array_context` property.

MTCam: > (...)
> * the pickling code had no tests 😱 (my bad!), which means there wasn't a way for anyone to notice what had happened.

Thanks for the explanation. It happens!  I definitely share the responsibility for the lack of testing .. at least downstream.

PR #241, opened 2021-06-30
Apply temporary work-around for mirgecom production performance issue.

MTCam:


inducer:
Given #242, maybe this is no longer needed.

PR #243, opened 2021-06-30
PyOpenCLArrayContext: emit valid transformation for scalar and 1D assignees

kaushikcfd:


meshmode/array_context.py
```diff
@@ -90,14 +90,16 @@ def _transform_loopy_inner(t_unit):
         for stmt in default_ep.instructions:
             if isinstance(stmt, lp.MultiAssignmentBase):
                 for assignee in stmt.assignees:
+                    if isinstance(assignee, Variable):
+                        # some scalar assignee kernel => no concurrency in the
+                        # workload => skip
+                        continue
```
inducer, line 96: Maybe check that there are no reduction inames? Otherwise we might inadvertently leave global reduction kernels untransformed.

inducer (commented):
Thanks! One suggestion to avoid potential future problems.

meshmode/array_context.py
kaushikcfd: I thought that, but the user tagged the kernel to be "Elementwise", do we need the extra-check here?

inducer: Fair point, maybe not.

inducer:
Thanks!

PR #251, opened 2021-07-21
Disable __bool__ on DOFArray

alexfikl:
`DOFArray`s are automagically truthy because they implement `__len__`. Hopefully this wasn't used in too many places.

Requires inducer/arraycontext#66 (for the error message to be useful)

### DIFF
--- a/meshmode/dof_array.py
+++ b/meshmode/dof_array.py
@@ -175,7 +175,7 @@ class DOFArray:
 
     def __bool__(self):
         raise ValueError(
-                "The truth value of a DOFArray is ambiguous. "
+                "The truth value of a DOFArray is not well-defined. "
                 "Use actx.np.any(x) or actx.np.all(x)")
 
     def __len__(self):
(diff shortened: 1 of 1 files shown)


alexfikl:
@thomasgibson I went ahead and made `DOFArray` complain when it's used like in inducer/grudge#151. Lets see if this was the case anywhere else!

thomasgibson:
@alexfikl Thanks! This is a very good idea --- so far it looks like things are mostly happy?

meshmode/dof_array.py
```diff
@@ -173,6 +173,11 @@ def __repr__(self):
 
     # {{{ sequence protocol
 
+    def __bool__(self):
+        raise ValueError(
+                "The truth value of a DOFArray is ambiguous. "
```
inducer, line 178:
```suggestion
                "The truth value of a DOFArray is not well-defined. "
```

inducer (approved):
Thanks!

PR #252, opened 2021-07-27
Overhaul facial adjacency + boundary tagging

majosm:
This PR:

1. Restructures `Mesh`'s facial adjacency structure to allow multiple adjacency groups for each pair of mesh element groups. In other words, `List[Dict[FacialAdjacencyGroup]]` becomes `List[List[FacialAdjacencyGroup]]`.
2. Splits `FacialAdjacencyGroup` into `InteriorAdjacencyGroup` and `BoundaryAdjacencyGroup` subclasses (with `InterPartitionAdjacencyGroup` inheriting from `BoundaryAdjacencyGroup`).
3. Splits boundary adjacency for different boundary tags into different groups.
4. Splits inter-partition adjacency for different ranks into different groups.
5. Removes `boundary_tags`/`boundary_tag_bit` from `Mesh`.
6. Adds affine transformations to `InteriorAdjacencyGroup`/`InterPartitionAdjacencyGroup`.

~~This will almost certainly need to be split into multiple PRs~~, but I wanted to throw this up here first for CI + discussion.

(Probably incomprehensible unless read commit-by-commit.)

### DIFF
--- a/meshmode/interop/firedrake/mesh.py
+++ b/meshmode/interop/firedrake/mesh.py
@@ -28,7 +28,6 @@ from modepy import resampling_matrix, simplex_best_available_basis
 
 from meshmode.mesh import (
     BTAG_ALL,
-    BTAG_REALLY_ALL,
     BTAG_INDUCED_BOUNDARY,
     Mesh,
     SimplexElementGroup,
@@ -179,9 +178,7 @@ def _get_firedrake_boundary_tags(fdrake_mesh, tag_induced_boundary=False):
     Return a tuple of bdy tags as requested in
     the construction of a :mod:`meshmode` :class:`Mesh`
 
-    The tags used are :class:`meshmode.mesh.BTAG_ALL`,
-    :class:`meshmode.mesh.BTAG_REALLY_ALL`, and
-    any markers in the mesh topology's exterior facets
+    The tags used are the markers in the mesh topology's exterior facets
     (see :attr:`firedrake.mesh.MeshTopology.exterior_facets.unique_markers`)
 
     :arg fdrake_mesh: A `firedrake.mesh.MeshTopology` or
@@ -191,7 +188,7 @@ def _get_firedrake_boundary_tags(fdrake_mesh, tag_induced_boundary=False):
 
     :return: A tuple of boundary tags
     """
-    bdy_tags = [BTAG_ALL, BTAG_REALLY_ALL]
+    bdy_tags = []
     if tag_induced_boundary:
         bdy_tags.append(BTAG_INDUCED_BOUNDARY)
 
@@ -219,9 +216,7 @@ def _get_firedrake_facial_adjacency_groups(fdrake_mesh_topology,
         Also, in this case boundary faces are tagged
         with :class:`meshmode.mesh.BTAG_INDUCED_BOUNDARY`
         if they are not a boundary face in *fdrake_mesh_topology* but become a
-        boundary because the opposite cell is not in *cells_to_use*.  Boundary
-        faces in *fdrake_mesh_topology* are marked with :class:`BTAG_ALL`. Both
-        are marked with :class:`BTAG_REALLY_ALL`.
+        boundary because the opposite cell is not in *cells_to_use*.
 
     :return: A list of maps to :class:`FacialAdjacencyGroup`s as required
         by a :mod:`meshmode` :class:`Mesh`.
@@ -251,27 +246,10 @@ def _get_firedrake_facial_adjacency_groups(fdrake_mesh_topology,
         (fd_local_facet_nr,) = fd_local_facet_nr  # extract id from set({id})
         fd_loc_fac_nr_to_mm[fd_local_facet_nr] = mm_local_facet_nr
 
-    # build a look-up table from firedrake markers to the appropriate values
-    # in the flags array for the external facial adjacency groups
+    # retrieve boundary tag information
     bdy_tags = _get_firedrake_boundary_tags(
         top, tag_induced_boundary=cells_to_use is not None)
     boundary_tag_to_index = {bdy_tag: i for i, bdy_tag in enumerate(bdy_tags)}
-    marker_to_flag_value = {}
-    from meshmode.mesh import _boundary_tag_bit
-    # for convenience,
-    # None maps to the boundary tag for a boundary facet with no marker
-    marker_to_flag_value[None] = (
-        _boundary_tag_bit(bdy_tags, boundary_tag_to_index, BTAG_REALLY_ALL)
-        | _boundary_tag_bit(bdy_tags, boundary_tag_to_index, BTAG_ALL))
-    # firedrake exterior facets with no marker are assigned the
-    # a dummy marker
-    from firedrake.mesh import unmarked as fd_unmarked
-    marker_to_flag_value[fd_unmarked] = marker_to_flag_value[None]
-    # Now figure out the appropriate tags for each firedrake markers
-    for marker in top.exterior_facets.unique_markers:
-        marker_to_flag_value[marker] = (
-            _boundary_tag_bit(bdy_tags, boundary_tag_to_index, marker)
-            | marker_to_flag_value[None])
 
     # {{{ build the FacialAdjacencyGroup for internal connectivity
 
@@ -321,16 +299,6 @@ def _get_firedrake_facial_adjacency_groups(fdrake_mesh_topology,
                                               newly_created_exterior_facs)
         new_ext_elements = int_elements[newly_created_exterior_facs]
         new_ext_element_faces = int_element_faces[newly_created_exterior_facs]
-        new_ext_flag = (
-            _boundary_tag_bit(bdy_tags,
-                              boundary_tag_to_index,
-                              BTAG_REALLY_ALL)
-            | _boundary_tag_bit(bdy_tags,
-                                boundary_tag_to_index,
-                                BTAG_INDUCED_BOUNDARY))
-        new_ext_flags = np.full(new_ext_elements.shape,
-                                new_ext_flag,
-                                dtype=IntType)
         # Remove any (previously) interior facets that have become exterior
         # facets
         remaining_int_facs = np.logical_not(newly_created_exterior_facs)
@@ -349,51 +317,62 @@ def _get_firedrake_facial_adjacency_groups(fdrake_mesh_topology,
 
     # {{{ build the FacialAdjacencyGroup for boundary faces
 
+    exterior_grps = []
+
     # We can get the elements directly from exterior facets
     ext_elements = top.exterior_facets.facet_cell.flatten()
 
     ext_element_faces = np.array([fd_loc_fac_nr_to_mm[fac_nr] for fac_nr in
                                   top.exterior_facets.local_facet_dat.data],
-                                 dtype=Mesh.face_id_dtype)
+                                  dtype=Mesh.face_id_dtype)
     # If only using some of the cells, throw away unused cells and
     # move to new cell index
-    exterior_facet_markers = top.exterior_facets.markers
+    ext_facet_markers = top.exterior_facets.markers
     if cells_to_use is not None:
         to_keep = np.isin(ext_elements, cells_to_use)
         ext_elements = np.vectorize(cells_to_use_inv.__getitem__)(
             ext_elements[to_keep])
         ext_element_faces = ext_element_faces[to_keep]
-        if exterior_facet_markers is not None:
-            exterior_facet_markers = exterior_facet_markers[to_keep]
+        if ext_facet_markers is not None:
+            ext_facet_markers = ext_facet_markers[to_keep]
 
     # tag the boundary, making sure to record custom tags
     # (firedrake "markers") if present
-    if top.exterior_facets.markers is not None:
-        ext_flags = np.zeros(ext_elements.shape, dtype=IntType)
-        for ifac, marker in enumerate(exterior_facet_markers):
-            ext_flags[ifac] = marker_to_flag_value[marker]
-    else:
-        ext_flags = np.full(ext_elements.shape,
-                            marker_to_flag_value[None],
-                            dtype=IntType)
+    ext_bdry_indices = np.full(ext_elements.shape, -1)
+    if ext_facet_markers is not None:
+        for ifac, marker in enumerate(ext_facet_markers):
+            ext_bdry_indices[ifac] = boundary_tag_to_index[marker]
+
+    for ibdry in range(len(bdy_tags)):
+        indices, = np.where(ext_bdry_indices == ibdry)
+        if len(indices) > 0:
+            exterior_grps.append(BoundaryAdjacencyGroup(
+                igroup=0,
+                boundary_tag=bdy_tags[ibdry],
+                elements=ext_elements[indices],
+                element_faces=ext_element_faces[indices]))
+
+    # Explicitly create BTAG_ALL group so that any unmarked boundaries are also
+    # accounted for
+    exterior_grps.append(BoundaryAdjacencyGroup(
+        igroup=0,
+        boundary_tag=BTAG_ALL,
+        elements=ext_elements,
+        element_faces=ext_element_faces))
 
     # If not using all the cells, some interior facets may have become
     # exterior facets:
     if cells_to_use is not None:
         # Record any newly created exterior facets
-        ext_elements = np.concatenate((ext_elements, new_ext_elements))
-        ext_element_faces = np.concatenate((ext_element_faces,
-                                            new_ext_element_faces))
-        ext_flags = np.concatenate((ext_flags, new_ext_flags))
-
-    exterior_grp = BoundaryAdjacencyGroup(igroup=0,
-                                        elements=ext_elements,
-                                        element_faces=ext_element_faces,
-                                        flags=ext_flags)
+        exterior_grps.append(BoundaryAdjacencyGroup(
+            igroup=0,
+            boundary_tag=BTAG_INDUCED_BOUNDARY,
+            elements=new_ext_elements,
+            element_faces=new_ext_element_faces))
 
     # }}}
 
-    return [{0: interconnectivity_grp, None: exterior_grp}]
+    return [[interconnectivity_grp] + exterior_grps]
 
 # }}}
 
@@ -545,8 +524,7 @@ build_connection_from_firedrake`.
         Note that in this latter case, some faces that are not
         boundaries in *fdrake_mesh* may become boundaries in the
         returned mesh. These "induced" boundaries are marked with
-        :class:`~meshmode.mesh.BTAG_INDUCED_BOUNDARY`
-        instead of :class:`~meshmode.mesh.BTAG_ALL`.
+        :class:`~meshmode.mesh.BTAG_INDUCED_BOUNDARY`.
 
     :arg normals: **Only** used if *fdrake_mesh* is a 1-surface
         embedded in 2-space. In this case,
@@ -600,9 +578,6 @@ build_connection_from_firedrake`.
     fdrake_mesh.init()
 
     # Get all the nodal information we can from the topology
-    bdy_tags = _get_firedrake_boundary_tags(
-        fdrake_mesh, tag_induced_boundary=cells_to_use is not None)
-
     with ProcessLogger(logger, "Retrieving vertex indices and computing "
                        "NodalAdjacency from firedrake mesh"):
         vertex_indices, nodal_adjacency = \
@@ -721,9 +696,6 @@ build_connection_from_firedrake`.
     def flip_local_face_indices(faces, elements):
         faces = np.copy(faces)
         neg_elements = np.full(elements.shape, False)
-        # To handle neighbor case, we only need to flip at elements
-        # who have a neighbor, i.e. where neighbors is not a negative
-        # bitmask of bdy tags
         neg_elements[elements >= 0] = (orient[elements[elements >= 0]] < 0)
         no_zero = np.logical_and(neg_elements, faces == no_zero_face_ndx)
         no_one = np.logical_and(neg_elements, faces == no_one_face_ndx)
@@ -755,12 +727,11 @@ build_connection_from_firedrake`.
                     facial_adjacency_groups[igroup].append(
                         BoundaryAdjacencyGroup(
                             igroup=igroup,
+                            boundary_tag=fagrp.boundary_tag,
                             elements=fagrp.elements,
-                            element_faces=new_element_faces,
-                            flags=fagrp.flags))
+                            element_faces=new_element_faces))
 
     return (Mesh(vertices, [group],
-                 boundary_tags=bdy_tags,
                  nodal_adjacency=nodal_adjacency,
                  facial_adjacency_groups=facial_adjacency_groups),
             orient)
--- a/meshmode/interop/firedrake/mesh.py
+++ b/meshmode/interop/firedrake/mesh.py
@@ -26,8 +26,16 @@ import numpy as np
 
 from modepy import resampling_matrix, simplex_best_available_basis
 
-from meshmode.mesh import (BTAG_ALL, BTAG_REALLY_ALL, BTAG_INDUCED_BOUNDARY,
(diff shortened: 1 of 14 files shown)


majosm:
(Looks like I need to do a little more work on the `BTAG_NONE`/`BTAG_ALL`/`BTAG_REALLY_ALL` handling. Stay tuned. 🙂)

majosm:
Just looking for a review of the high-level overview in the OP to check if this seems reasonable (though feel free to take a glance through the code if you're feeling particularly adventurous. 🙂) After that I'll split this up into smaller pieces.

inducer (commented):
Thanks for working on this! Here's an initial review up to (but not including) 5cda6670d67bee68807606493d822190bc8b0012 ("remove unnecessary intermediate steps in partition_mesh")

meshmode/mesh/processing.py
```diff
@@ -232,43 +232,55 @@ def _create_local_to_local_adjacency_groups(mesh, global_elem_to_part_elem,
...
 
-       The index of the shared face inside the remote element.
+        ``face_id_t [nfaces]``. ``neighbor_faces[i]`` stores the element-relative
+        face index of the remote face.
+
+#     .. attribute:: aff_transform
+
+#         ``(float [ndim, ndim], float [ndim])``. ``aff_transform[0]`` and
+#         ``aff_transform[1]`` store the matrix and vector parts of the affine
+#         transformation that maps the local faces to the remote faces.
```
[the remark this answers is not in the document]
majosm: This gets uncommented in 2538baad833b6ea2c.

meshmode/mesh/processing.py
```diff
@@ -23,7 +23,7 @@
 from dataclasses import dataclass
 from functools import reduce
 from numbers import Real
-from typing import Optional, Union
+from typing import Optional, Union  # Tuple
```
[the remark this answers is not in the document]
majosm: This also gets uncommented in 2538baad833b6ea2c.

meshmode/mesh/__init__.py
```diff
@@ -136,9 +145,32 @@ class BTAG_INDUCED_BOUNDARY(BTAG_NO_BOUNDARY):  # noqa: N801
     # firedrakeproject.org seems to reject connections from Github.
 
 
-SYSTEM_TAGS = {BTAG_NONE, BTAG_ALL, BTAG_REALLY_ALL, BTAG_NO_BOUNDARY,
+SYSTEM_TAGS = {BTAG_NONE, BTAG_ALL, BTAG_REALLY_ALL, BTAG_DEFAULT, BTAG_NO_BOUNDARY,
                    BTAG_PARTITION, BTAG_INDUCED_BOUNDARY}
 
+
+def boundary_tag_in(boundary_tag, boundary_tags):
```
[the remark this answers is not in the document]
majosm: As far as I'm aware, I didn't change anything to prevent a given face from having multiple tags (in the sense of belonging to multiple `FacialAdjacencyGroups`).

However, there isn't currently a `FacialAdjacencyGroup` corresponding to `BTAG_NONE`, `BTAG_ALL`, or `BTAG_REALLY_ALL`, so these have become a little bit less like regular boundary tags and more like meta-tags. This is one thing I haven't fully resolved in this PR (you'll see comments in a couple of places about this, if you haven't already), and I'm hoping to get some input on how to approach it. My current temptation would be to rename `BTAG_[NONE, ALL, REALLY_ALL]` to something else and treat them as special values that represent a subset of boundary tags (as I'm doing here in `boundary_tag_in`, where one can pass in a list of boundary tags or one of these values). I added a new tag `BTAG_DEFAULT` to handle the other prior role of `BTAG_ALL`, which is to be the default boundary tag used when facial adjacency is created without a `face_vertex_indices_to_tags` mapping.

meshmode/mesh/__init__.py
```diff
@@ -136,9 +145,32 @@ class BTAG_INDUCED_BOUNDARY(BTAG_NO_BOUNDARY):  # noqa: N801
...
+    if boundary_tags == BTAG_NONE:
+        return False
+    elif boundary_tags == BTAG_ALL:
```
inducer, line 155: `BTAG_ALL` isn't currently the union of all tagged boundaries. It's everything that's a boundary (tagged or not) except the rank boundaries.

meshmode/mesh/__init__.py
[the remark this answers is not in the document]
inducer: (After thinking about it for a bit, I *think* my bias would be towards actually including those special BTAGs in the data structure.)

meshmode/mesh/__init__.py
```diff
@@ -468,10 +468,32 @@ class FacialAdjacencyGroup(Record):
 
         Zero if ``neighbors[i]`` is negative.
 
+    .. attribute:: aff_transform
+
+        ``(np.float64 [ambient_dim, ambient_dim], np.float64 [ambient_dim])`` or
+        (None, None). ``aff_transform[0]`` and ``aff_transform[1]`` give the
```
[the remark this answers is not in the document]
inducer: Just (re-)discovered: there already is an `AffineMap` class in `meshmode.mesh.tools`.

meshmode/mesh/__init__.py
```diff
@@ -468,10 +468,32 @@ class FacialAdjacencyGroup(Record):
 
         Zero if ``neighbors[i]`` is negative.
 
+    .. attribute:: aff_transform
+
+        ``(np.float64 [ambient_dim, ambient_dim], np.float64 [ambient_dim])`` or
```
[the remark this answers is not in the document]
majosm: Fixed.

meshmode/mesh/__init__.py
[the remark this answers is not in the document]
majosm: Done (using the existing `AffineMap`, extended a little).

meshmode/mesh/__init__.py
```diff
@@ -468,10 +468,32 @@ class FacialAdjacencyGroup(Record):
...
     .. automethod:: __ne__
     """
 
+    def __init__(self, igroup, ineighbor_group, *,
+            elements, element_faces,
+            neighbors, neighbor_faces,
+            aff_transform=None,
+            **kwargs):
+        if aff_transform is None:
+            aff_transform = (None, None)
```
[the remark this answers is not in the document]
majosm: Fixed in tuple-to-dataclass conversion (for the record: they can individually be `None` or not).

meshmode/mesh/__init__.py
```diff
@@ -1297,11 +1352,18 @@ def fagrp_params_str(fagrp):
...
             params = {
                     "igroup": fagrp.igroup,
-                    "ineighbor_group": repr(fagrp.ineighbor_group),
+                    "ineighbor_group": fagrp.ineighbor_group,
                     "elements": _numpy_array_as_python(fagrp.elements),
                     "element_faces": _numpy_array_as_python(fagrp.element_faces),
                     "neighbors": _numpy_array_as_python(fagrp.neighbors),
                     "neighbor_faces": _numpy_array_as_python(fagrp.neighbor_faces),
+                    "aff_transform": (
+                            _numpy_array_as_python(fagrp.aff_transform[0])
```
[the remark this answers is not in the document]
majosm: Fixed in tuple-to-dataclass conversion.

meshmode/mesh/__init__.py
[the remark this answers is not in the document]
majosm: Forgot to mention: I wasn't sure if `tools.py` is meant to be a public module or not; for now I left `AffineMap` in there and added the module docs, but I can move it somewhere else if `tools` isn't supposed to be user-visible.

meshmode/mesh/__init__.py
```diff
@@ -468,10 +468,32 @@ class FacialAdjacencyGroup(Record):
...
 
+    def __init__(self, igroup, ineighbor_group, *,
+            elements, element_faces,
+            neighbors, neighbor_faces,
+            aff_transform=None,
+            **kwargs):
+        if aff_transform is None:
+            aff_transform = (None, None)
+
+        Record.__init__(self,
```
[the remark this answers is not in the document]
majosm: Fixed.

meshmode/mesh/__init__.py
```diff
@@ -554,9 +587,31 @@ class InterPartitionAdjacencyGroup(FacialAdjacencyGroup):
...
+            aff_transform=None,
+            **kwargs):
+        FacialAdjacencyGroup.__init__(self,
+            igroup=igroup, ineighbor_group=ineighbor_group,
+            elements=elements, element_faces=element_faces,
+            neighbors=neighbors, neighbor_faces=neighbor_faces,
+            aff_transform=aff_transform,
+            partition_neighbors=partition_neighbors,
+            neighbor_partitions=neighbor_partitions,
+            **kwargs)
```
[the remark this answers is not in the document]
majosm: This eventually becomes (slightly) nontrivial in a later commit.

meshmode/mesh/__init__.py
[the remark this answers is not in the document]
majosm: I took a stab at using the explicit groups for `BTAG_ALL`, etc. It seems to work OK. (You'll see it when you get to "split boundary adjacency by boundary tag and inter-partition adjacency by rank".)

meshmode/mesh/__init__.py
majosm: Should be fixed, I think.

majosm:
I think this is ready for another look @inducer.

meshmode/mesh/__init__.py
```diff
@@ -1382,40 +1318,139 @@ def boundary_tag_bit(boundary_tag):
...
+
+    return BoundaryAdjacencyGroup(
+        igroup=igrp,
+        boundary_tag=merged_btag,
+        elements=merged_elements,
+        element_faces=merged_element_faces)
+
+
+def _complete_facial_adjacency_groups(
+        facial_adjacency_groups, element_id_dtype, face_id_dtype):
```
inducer, line 1367: Help! I want a docstring. :)

meshmode/mesh/__init__.py
```diff
@@ -1533,42 +1568,50 @@ def check_bc_coverage(mesh, boundary_tags, incomplete_ok=False,
...
-    for fagrp_list in mesh.facial_adjacency_groups:
-        bdry_grps = [
-            grp for grp in fagrp_list
-            if isinstance(grp, BoundaryAdjacencyGroup)]
-        if not bdry_grps:
-            continue
-
-        for bdry_grp in bdry_grps:
-            nb_el_bits = bdry_grp.flags
+        all_bdry_grps = [
```
inducer, line 1556:
```
all_bdry_grp, =
```

meshmode/mesh/tools.py
```diff
@@ -27,6 +27,11 @@
 from pytools.spatial_btree import SpatialBinaryTreeBucket
 from pytools import MovedFunctionDeprecationWrapper
 
+__doc__ = """
+
+.. autoclass:: AffineMap
```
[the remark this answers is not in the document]
majosm: Any preference between 4c62225 and https://github.com/majosm/meshmode/commit/9927877660cfc5a242af9b419b575e5e828b4949? (The doc output looks the same between them from what I can tell.)

meshmode/mesh/processing.py
```diff
@@ -167,6 +166,43 @@ def _filter_mesh_groups(groups, selected_elements, vertex_id_dtype):
...
+        partition numbers.
+    :arg global_elem_to_part_elem: A :class:`numpy.ndarray` mapping from global
+        element index to local partition-wide element index for local elements (and
+        -1 otherwise).
+
+    :returns: A :class:`set` of indices of the neighboring partitions.
+    """
+    connected_parts = set()
+
+    for igrp, facial_adj_list in enumerate(mesh.facial_adjacency_groups):
```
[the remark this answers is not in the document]
majosm: Just to make sure I understand: the `#partitions` comes into this because `partition_mesh` is called for each partition, one at a time, right?

If so, maybe I should put the comment next to `partition_mesh`, since it applies to the whole mesh partitioning process, more or less. (I seem to recall starting to fix this at one point? Need to remember where I put that...)

meshmode/mesh/__init__.py
```diff
@@ -1199,59 +1270,178 @@ def boundary_tag_bit(boundary_tag):
...
+                    element_faces = bdry_element_faces[indices]
+                    grp_list.append(
+                        BoundaryAdjacencyGroup(
+                            igroup=igrp,
+                            boundary_tag=btag,
+                            elements=elements,
+                            element_faces=element_faces))
+
+            is_untagged = ~np.any(belongs_to_bdry, axis=0)
+            if len(is_untagged) > 0:
```
thomasgibson, line 1316: I think this should work just fine too. Personal preference though:
```suggestion
            if is_untagged:
```

meshmode/mesh/__init__.py
```diff
@@ -439,128 +484,218 @@ class FacialAdjacencyGroup(Record):
...
-                and np.array_equal(self.neighbors, other.neighbors)
-                and np.array_equal(self.neighbor_faces, other.neighbor_faces)
-                )
+            super().__eq__(other)
+            and self.ineighbor_group == other.ineighbor_group
+            and np.array_equal(self.elements, other.elements)
+            and np.array_equal(self.element_faces, other.element_faces)
+            and np.array_equal(self.neighbors, other.neighbors)
+            and np.array_equal(self.neighbor_faces, other.neighbor_faces)
+            and self.aff_map == other.aff_map)
```
thomasgibson, line 542: I know this logic was more or less already in `meshmode` before this PR, but I'm just asking for curiosity --- are we not perhaps a bit worried about the cost of comparing two `FacialAdjacencyGroup` objects? I can imagine the arrays being quite large for production runs.

meshmode/mesh/__init__.py
```diff
@@ -430,6 +439,42 @@ class FacialAdjacencyGroup(Record):
     necessarily contiguous like the figure may suggest.)
 
     .. attribute:: igroup
+    """
+
+    def __init__(self, igroup, **kwargs):
```
thomasgibson, line 444: Is it documented anywhere what acceptable keyword args are permitted?

meshmode/mesh/__init__.py
```diff
@@ -785,24 +883,20 @@ def __init__(self, vertices, groups, *, skip_tests=False,
...
+                            assert fagrp.element_faces.dtype == self.face_id_dtype
+                            assert fagrp.element_faces.shape == (nfagrp_elements,)
+                            assert fagrp.neighbors.dtype == self.element_id_dtype
+                            assert fagrp.neighbors.shape == (nfagrp_elements,)
+                            assert fagrp.neighbor_faces.dtype == self.face_id_dtype
+                            assert fagrp.neighbor_faces.shape == (nfagrp_elements,)
+                        else:
+                            nfagrp_elements, = fagrp.elements.shape
+                            assert fagrp.element_faces.dtype == self.face_id_dtype
+                            assert fagrp.element_faces.shape == (nfagrp_elements,)
```
thomasgibson, line 899:
```suggestion
                        nfagrp_elements, = fagrp.elements.shape
                        assert fagrp.element_faces.dtype == self.face_id_dtype
                        assert fagrp.element_faces.shape == (nfagrp_elements,)
                        if isinstance(fagrp, InteriorAdjacencyGroup):
                            assert fagrp.neighbors.dtype == self.element_id_dtype
                            assert fagrp.neighbors.shape == (nfagrp_elements,)
                            assert fagrp.neighbor_faces.dtype == self.face_id_dtype
                            assert fagrp.neighbor_faces.shape == (nfagrp_elements,)
```

meshmode/mesh/__init__.py
```diff
@@ -1199,59 +1270,178 @@ def boundary_tag_bit(boundary_tag):
...
-            element_faces = faces.astype(face_id_dtype)
-            neighbors = np.full(len(elements),
-                -(boundary_tag_bit(BTAG_ALL)
-                    | boundary_tag_bit(BTAG_REALLY_ALL)),
-                dtype=element_id_dtype)
+            bdry_indices = np.where(~face_has_neighbor)
+            bdry_element_faces = bdry_indices[0].astype(face_id_dtype)
+            bdry_elements = bdry_indices[1].astype(element_id_dtype)
+            belongs_to_bdry = np.full(
+                (len(boundary_tags), len(bdry_elements)), False)
```
thomasgibson, line 1291: The main reason I prefer this is that the unpacking of `np.where(~face_has_neighbor)` is implicitly checking here that we unpack the expected number of items
```suggestion
            bdry_element_faces, bdry_elements = np.where(~face_has_neighbor)
            bdry_element_faces = bdry_element_faces.astype(face_id_dtype)
            bdry_elements = bdry_elements.astype(element_id_dtype)
            belongs_to_bdry = np.full(
                (len(boundary_tags), len(bdry_elements)), False)
```

meshmode/mesh/__init__.py
```diff
@@ -1199,59 +1270,178 @@ def boundary_tag_bit(boundary_tag):
...
-                neighbors=neighbors,
-                neighbor_faces=neighbor_faces)
-
-        facial_adjacency_groups.append(grp_map)
+                            btag_idx = boundary_tag_to_index[tag]
+                            belongs_to_bdry[btag_idx, i] = True
+
+            for btag_idx, btag in enumerate(boundary_tags):
+                indices, = np.where(belongs_to_bdry[btag_idx, :])
+                if len(indices) > 0:
```
thomasgibson, line 1305: This should be sufficient too:
```suggestion
                if indices:
```

meshmode/mesh/__init__.py
```diff
@@ -1371,42 +1556,49 @@ def check_bc_coverage(mesh, boundary_tags, incomplete_ok=False,
...
+            return counts
 
-            # Consider non-boundary faces 'seen'
-            seen = seen | ~tag_set
+        all_bdry_counts = get_bdry_counts(all_bdry_grp)
 
-        for btag in boundary_tags:
-            tag_bit = mesh.boundary_tag_bit(btag)
-            tag_set = (nb_el_bits & tag_bit) != 0
+        if len(matching_bdry_grps) > 0:
```
thomasgibson, line 1587:
```suggestion
        if matching_bdry_grps:
```

meshmode/interop/firedrake/mesh.py
```diff
@@ -344,57 +317,62 @@ def _get_firedrake_facial_adjacency_groups(fdrake_mesh_topology,
...
-                                marker_to_neighbor_value[None],
-                                dtype=IntType)
+    ext_bdry_indices = np.full(ext_elements.shape, -1)
+    if ext_facet_markers is not None:
+        for ifac, marker in enumerate(ext_facet_markers):
+            ext_bdry_indices[ifac] = boundary_tag_to_index[marker]
+
+    for ibdry in range(len(bdy_tags)):
+        indices, = np.where(ext_bdry_indices == ibdry)
+        if len(indices) > 0:
```
thomasgibson, line 348:
```suggestion
        if indices:
```

meshmode/mesh/processing.py
```diff
@@ -291,86 +303,72 @@ def _collect_nonlocal_adjacency_data(mesh, part_per_elem, global_elem_to_part_el
...
+                    & (neighbor_parts == i_neighbor_part))
 
-        if pairwise_adj:
-            nonlocal_adj_data[i_part_grp] = _NonLocalAdjacencyData(
-                np.concatenate([adj.elements for adj in pairwise_adj]),
-                np.concatenate([adj.element_faces for adj in pairwise_adj]),
-                np.concatenate([adj.neighbor_parts for adj in pairwise_adj]),
-                np.concatenate([adj.global_neighbors for adj in pairwise_adj]),
-                np.concatenate([adj.neighbor_faces for adj in pairwise_adj]))
+                if len(adj_indices) > 0:
```
thomasgibson, line 345:
```suggestion
                if adj_indices:
```

meshmode/mesh/processing.py
```diff
@@ -383,151 +381,42 @@ def _collect_bdry_data(mesh, global_elem_to_part_elem, part_mesh_groups,
...
+        for bdry_grp in bdry_grps:
+            elem_base = mesh.groups[igrp].element_nr_base
 
-            bdry_data[i_part_grp] = _BoundaryData(elements, element_faces, neighbors)
-
-    return bdry_data
+            adj_indices, = np.where(global_elem_to_part_elem[bdry_grp.elements
+                        + elem_base] >= 0)
 
+            if len(adj_indices) > 0:
```
thomasgibson, line 405:
```suggestion
            if adj_indices:
```

thomasgibson (commented):
Some comments/touch-ups in places after reading through the changes. Thank you for spending the time doing all this!

meshmode/mesh/__init__.py
majosm: Actually, now that I look at it I think it's wrong the way I have it. It should be:

```suggestion
            if np.any(is_untagged):
```

meshmode/mesh/__init__.py
majosm: I don't know of a case where we would want to compare facial adjacency groups inside a running simulation. My guess is that this is just here for testing purposes. (Though I don't think we're using it for that either at the moment...)

meshmode/mesh/__init__.py
majosm: All of 'em. 🙂 This is just a passthrough to the `Record` constructor, which takes in all of the fields for the class being initialized. So what gets passed in comes down to what fields the particular subclass defines.

meshmode/mesh/__init__.py
majosm: I don't think I can do these. When I try, I get:

```python
ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
```

thomasgibson: Oh apologies! Then disregard this and related comments if they don't actually work!

meshmode/mesh/__init__.py
majosm: Would something like https://github.com/majosm/meshmode/commit/779969e5af252547b1cfab2b0c49af27cb2e8fe8 be clearer, maybe?

It looks like `FacialAdjacencyGroup` isn't the only place where this pattern is used, though (I see it being used for `MeshElementGroup` as well, not sure what else), so breaking from convention here may be inadvisable? 🤷‍♂️

meshmode/mesh/__init__.py
inducer: I agree that these comparisons are potentially expensive. IMO, that's fine... if someone asks for them, they'll have to pay for it. AFAIK, they're (currently) conly used to ensure the correctness of the `as_python` functionality.

meshmode/mesh/__init__.py
inducer: This is really the fault of `Record`, which insists that all subclasses pass fields that they need set to its constructor. I'm no longer a fan of this, and I think we should move away from `Record` types altogether. (`dataclass`es are a much saner version of the same thing.) That's out of scope for this though, IMO. (-> #266)

meshmode/mesh/__init__.py
inducer: 42f88c8

meshmode/mesh/tools.py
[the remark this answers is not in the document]
inducer: Not a strong preference. 4c62225 is nice because it's easy to see how `AffineMap` is included in the docs. https://github.com/majosm/meshmode/commit/9927877660cfc5a242af9b419b575e5e828b4949 is perhaps a bit more granular. Overall, I think I'm with 4c62225 aka the way it is now.

meshmode/mesh/processing.py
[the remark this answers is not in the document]
inducer: TBH, I'm not sure I understand the original comment any more. I agree with the rest of what you said though---the only bad thing about this that I can see is that we're doing the partitioning once for each part (and recomputing a bunch of stuff in the process). I'd say nvm on this.

inducer:
Thanks for working on this, and thanks @thomasgibson  for taking a look.

PR #254, opened 2021-08-15
resample_by_picking: use array expressions

kaushikcfd:
Why is this approach better?
For PytatoArrayContext: cleaner to fuse over other possible approach of inferring the semantics of a loopy call.

Draft because:
- [x] Needs https://github.com/inducer/pytato/pull/136
- [x] Needs https://github.com/inducer/arraycontext/pull/74
- [x] Needs https://github.com/inducer/arraycontext/pull/125
---
- [x] Point arraycontext in `requirements.txt` back to `main`.

PR #255, opened 2021-08-17
Add workaround for `compose_index_maps` kernel issue

majosm:
Context: illinois-ceesd/mirgecom#468

Works around https://github.com/pocl/pocl/issues/979

### DIFF
--- a/meshmode/discretization/connection/direct.py
+++ b/meshmode/discretization/connection/direct.py
@@ -148,12 +148,21 @@ class InterpolationBatch:
                 "iel_init": ConcurrentElementInameTag(),
                 "iel": ConcurrentElementInameTag()})
 
-        result = actx.freeze(actx.call_loopy(
-            compose_index_maps_kernel(),
-            from_element_indices=self.from_element_indices,
-            to_element_indices=self.to_element_indices,
-            nelements_result=to_group.nelements,
-        )["global_from_element_indices"])
+        # This produces incorrect results on certain machines (details:
+        # https://github.com/pocl/pocl/issues/979)
+#         result = actx.freeze(actx.call_loopy(
+#             compose_index_maps_kernel(),
+#             from_element_indices=self.from_element_indices,
+#             to_element_indices=self.to_element_indices,
+#             nelements_result=to_group.nelements,
+#         )["global_from_element_indices"])
+
+        # Compute on the CPU instead for now
+        from_element_indices = actx.to_numpy(self.from_element_indices)
+        to_element_indices = actx.to_numpy(self.to_element_indices)
+        numpy_result = np.full(to_group.nelements, -1)
+        numpy_result[to_element_indices] = from_element_indices
+        result = actx.freeze(actx.from_numpy(numpy_result))
 
         self._global_from_element_indices_cache = result
         return result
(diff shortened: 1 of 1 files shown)


majosm:
@inducer Weird, the MPI tests are failing (looks like in #254 as well). I can't remember, did the CI previously use OpenMPI or MPICH? It's seems to be using MPICH now, and it's apparently not liking the (OpenMPI-specific, I think) `--oversubscribe` argument to `mpiexec`.

thomasgibson:
> @inducer Weird, the MPI tests are failing (looks like in #254 as well). I can't remember, did the CI previously use OpenMPI or MPICH? It's seems to be using MPICH now, and it's apparently not liking the (OpenMPI-specific, I think) `--oversubscribe` argument to `mpiexec`.

Yeah I noticed this in Grudge CI as well: https://github.com/inducer/grudge/pull/154/checks?check_run_id=3352770289#step:3:709 --- I'm thinking this is likey due to using MPICH rather than OpenMPI

inducer:
I think @alexfikl addressed the MPI issue in #257.

meshmode/discretization/connection/direct.py
```diff
@@ -148,12 +148,21 @@ def compose_index_maps_kernel():
...
+#             to_element_indices=self.to_element_indices,
+#             nelements_result=to_group.nelements,
+#         )["global_from_element_indices"])
+
+        # Compute on the CPU instead for now
+        from_element_indices = actx.to_numpy(self.from_element_indices)
+        to_element_indices = actx.to_numpy(self.to_element_indices)
+        numpy_result = np.full(to_group.nelements, -1)
+        numpy_result[to_element_indices] = from_element_indices
+        result = actx.freeze(actx.from_numpy(numpy_result))
```
alexfikl, line 165: I'm assuming the loopy version is faster, if for no other reason than it doesn't have to do the `to_numpy` exchange.

Is there any way to add an `if` in here? `if is_lazy` or `if is_bad_machine`?

majosm: It might be possible to add a conditional, but I'm not sure if: 1) this is actually lazy-specific, 2) it's limited to a specific architecture/environment that we could check for. @inducer?

Also, given that it's only computed once (per `InterpolationBatch`) and then cached, my guess is that it doesn't matter a whole lot performance-wise...?

majosm:
Hrm, test failed. Back to draft for now.

meshmode/discretization/connection/direct.py
alexfikl: Ah, fair point if it's just computed once at the beginning of a simulation!

Then I would recommend to just delete all the loopy stuff in here and add a FIXME pointing back to this PR (and the POCL issue) if anyone ever needs to add it back?

majosm:
Oh, it's failing on main too...

inducer:
Yep, my bad, came from this: https://github.com/inducer/gmsh_interop/pull/14. Will have a fix shortly.

inducer:
https://github.com/inducer/gmsh_interop/pull/15 Sorry again. Will also add downstream CI to this, so that this won't recur.

inducer:
It's in. I've rerun the tests here.

inducer:
Thanks!

PR #257, opened 2021-08-18
Only oversubscribe on openmpi

alexfikl:
I'm guessing `mpi4py` started pulling in `mpich` instead of `openmpi` in the CI (although couldn't find any changes in the conda-forge packages?) and `test_partition` started failing because it couldn't find the `--oversubscribe` flag.

This just adds `openmpi` to the conda env.

### DIFF
--- a/.test-conda-env-py3.yml
+++ b/.test-conda-env-py3.yml
@@ -24,6 +24,7 @@ dependencies:
 - pybind11
 
 # for MPI-based tests
+- openmpi
 - mpi4py
 
 # for xdmf/hdf5 visualizer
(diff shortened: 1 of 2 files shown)


majosm:
I think `-x` might also be OpenMPI-specific? (Let's see what happens in CI.)

alexfikl:
> I think `-x` might also be OpenMPI-specific? (Let's see what happens in CI.)

Ah, then yeah, might be better to just pull in openmpi from conda in the first place.

alexfikl:
Hm, it also seems to really not appreciate that I'm importing `from mpi4py import MPI` in the parent.

inducer:
Thx!

PR #259, opened 2021-08-28
Merge pickable interpolation batches in direct connection

inducer:
To do:

- [ ] Show that this has a  benefit

### DIFF
--- a/meshmode/discretization/connection/direct.py
+++ b/meshmode/discretization/connection/direct.py
@@ -21,6 +21,9 @@ THE SOFTWARE.
 """
 
 
+import logging
+from typing import Any
+
 import numpy as np
 import numpy.linalg as la
 
@@ -31,6 +34,10 @@ from pytools import memoize_in, keyed_memoize_method
 from arraycontext import (
         ArrayContext, make_loopy_program,
         is_array_container, map_array_container)
+
+from dataclasses import dataclass
+
+logger = logging.getLogger(__name__)
 
 
 # {{{ interpolation batch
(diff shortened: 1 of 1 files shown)


inducer:
@kaushikcfd As far as I can tell, this works now. I don't know how it'll interact with #254, but I hope we can find a way.

meshmode/discretization/connection/direct.py
```diff
@@ -337,20 +408,119 @@ def __call__(self, ary, _force_no_inplace_updates=False):
...
+        """Returns a list of :class:`_FromGroupPickData`, one per source group
+        from which data ist to be transferred, or *None*, if conditions for
+        this representation are not met.
+        """
+        cgrp = self.groups[i_tgrp]
+        tgrp = self.to_discr.groups[i_tgrp]
+
+        batch_dof_pick_lists = [
+                actx.to_numpy(
+                    self._resample_point_pick_indices(actx, i_tgrp, i_batch))
```
kaushikcfd, line 425: Thanks @inducer, I tried this branch and got an error here as [`_resample_point_pick_indices` could return `None`](https://github.com/inducer/meshmode/blob/04bcfba6a94091c90d488b628a516d0316a1e5f3/meshmode/discretization/connection/direct.py#L306) making `actx.to_numpy` unhappy.

inducer: Good catch! 8d08326512ad83ca497f0c6765dc098df3ee8196 should address that.

PR #261, opened 2021-09-03
Add a cubed sphere

alexfikl:
As suggested, it just does the same thing as the simplex-based icosahedron generator.

### DIFF
--- a/meshmode/mesh/generation.py
+++ b/meshmode/mesh/generation.py
@@ -518,7 +518,7 @@ def generate_sphere(r: float, order: int, *,
         group_cls: Optional[type] = None):
     """
     :param r: radius of the sphere.
-    :param order: order of the (simplex) elements. If *unit_nodes* is also
+    :param order: order of the group elements. If *unit_nodes* is also
         provided, the orders should match.
     :param uniform_refinement_rounds: number of uniform refinement rounds to
         perform after the initial mesh was created.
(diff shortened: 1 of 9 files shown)


alexfikl:
And looks something like this

<p align="center">

</p>

meshmode/mesh/generation.py
```diff
@@ -465,6 +465,29 @@ def generate_icosahedron(
             node_vertex_consistency_tolerance=node_vertex_consistency_tolerance,
             is_conforming=True)
 
+
+def generate_cube(r: float, order: int, *,
```
inducer, line 469: * `generate_cube_surface`?
* Add to docs?

meshmode/mesh/generation.py
```diff
@@ -486,9 +510,20 @@ def generate_icosphere(r: float, order: int, *,
...
+    if issubclass(group_cls, SimplexElementGroup):
+        mesh = generate_icosahedron(r, order,
+                node_vertex_consistency_tolerance=node_vertex_consistency_tolerance,
+                unit_nodes=unit_nodes)
+    elif issubclass(group_cls, TensorProductElementGroup):
+        mesh = generate_cube(r, order,
+                node_vertex_consistency_tolerance=node_vertex_consistency_tolerance,
+                unit_nodes=unit_nodes)
+    else:
+        raise TypeError(f"unsupported 'group_cls': {group_cls}")
```
inducer, line 526: The name `icosphere` doesn't match this: The cubed sphere isn't derived from an icosahedron. :)

Rename as `generate_sphere`? Make deprecate `generate_icosphere` a deprecated wrapper?

inducer (commented):
Thx! Few wrinkles below

meshmode/mesh/generation.py
alexfikl: Done in 09dbdea

meshmode/mesh/generation.py
alexfikl: Done in 88d184a.

inducer (commented):
LGTM, thanks! I'll apply the one suggestion.

PR #263, opened 2021-09-05
Introduce MeshElementGroup-generic group factories

alexfikl:
Explicitly using `OrderAndTypeBasedGroupFactory` everywhere (in inducer/pytential#46) was getting a bit verbose.

This basically introduces a `HeterogeneousOrderBasedGroupFactory` (naming?) that's supposedly a bit more extensible than `OrderAndTypeBasedGroupFactory` and a bunch of subclasses for the common cases that meshmode/grudge/pytential use.

Not quite sure how future proof this is, so suggestions welcome!

### DIFF
--- a/meshmode/discretization/poly_element.py
+++ b/meshmode/discretization/poly_element.py
@@ -30,10 +30,12 @@ from warnings import warn
 import numpy as np
 from pytools import memoize_method, memoize_on_first_arg
 from meshmode.mesh import (
+        MeshElementGroup as _MeshElementGroup,
         SimplexElementGroup as _MeshSimplexElementGroup,
         TensorProductElementGroup as _MeshTensorProductElementGroup)
 from meshmode.discretization import (
         NoninterpolatoryElementGroupError,
+        ElementGroupBase,
         NodalElementGroupBase, ModalElementGroupBase,
         InterpolatoryElementGroupBase)
 
@@ -72,6 +74,7 @@ Group factories
 ^^^^^^^^^^^^^^^
 
 .. autoclass:: ElementGroupFactory
+.. autoclass:: HomogeneousOrderBasedGroupFactory
 .. autoclass:: TypeMappingGroupFactory
 
 Simplicial group factories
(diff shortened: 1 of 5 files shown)


meshmode/discretization/poly_element.py
```diff
@@ -643,24 +653,52 @@ def __call__(self, mesh_el_group, index):
         return self.group_class(mesh_el_group, self.order, index)
 
 
-class OrderAndTypeBasedGroupFactory(ElementGroupFactory):
-    def __init__(self, order, simplex_group_class, tensor_product_group_class):
+class HeterogeneousOrderBasedGroupFactory(ElementGroupFactory):
```
inducer, line 656:
```suggestion
class GroupTypeMapGroupFactory(ElementGroupFactory):
```
?

inducer (commented):
LGTM. All I've got is a moderately unconvincing counterproposal on the name.

meshmode/discretization/poly_element.py
alexfikl: `ElementTypeMappingGroupFactory`? Too many `Group` in that one, sounds funny.. Hm..

inducer: `TypeMappingGroupFactory`?

thomasgibson: Don't mean to ruin the trend (*flees*), but I actually don't totally hate `HeterogeneousOrderBasedGroupFactory` (heterogeneous polynomial for example comes to mind)

alexfikl: :shrug: `TypeMappingGroupFactory` doesn't sound half bad, I'll go ahead and change it to that.

@thomasgibson There's already a `HomogeneousOrderBasedGroupFactory` too! Oh well..

inducer: OK, last chance to convince me. I like it the way it is now, but if either of you prefer a different name, ... this is the time to say so. :)

thomasgibson: So what should we do with `HomogeneousOrderBasedGroupFactory`? The reason I'm slightly leaning towards `HeterogeneousOrderBasedGroupFactory` is because its both consistent and mathematically correct

inducer: My main objection to "heterogeneous" as a name is that while it describes that this can produce discretizations with different-type groups, it doesn't describe *how* it does that (vs "type-mapping" which is quite explicit in that regard). There's less ambiguity in the homogeneous case.

inducer: @thomasgibson, what's your thinking on naming?

thomasgibson: > My main objection to "heterogeneous" as a name is that while it describes that this can produce discretizations with different-type groups, it doesn't describe _how_ it does that (vs "type-mapping" which is quite explicit in that regard). There's less ambiguity in the homogeneous case.

In that case, I'm fine with leaving the name as proposed. So long as it's clearly documented. So my final suggestion to @alexfikl would be to make sure the docs for `TypeMappingGroupFactory` detail its intended use

alexfikl: @thomasgibson Updated the docs for the group factories a bit in 1291014. Let me know if you'd like any other details in there!

thomasgibson (approved):
Thanks! This looks great. Thanks for adding the docs

inducer:
Thanks!

inducer:
(And thanks @thomasgibson for taking a look!)

Issue #258, opened 2021-08-24
Modal / nodal tracking


inducer:
Right now, it's possible to feed modal data to things that expect nodal data and vice versa. Everything is just a bunch of numbers in a `DOFArray`. Especially if @thomasgibson's entropy-stable work manages to stick, this will be a persistent source of user error. We should perhaps have the `DOFArray` remember the [`discretization_key`](https://documen.tician.de/meshmode/discretization.html?highlight=discretization_key#meshmode.discretization.ElementGroupBase.discretization_key) for each group and check those on input to make sure they're as expected.

alexfikl:
Would it work to introduce some subclassses `NodalDOFArray` and `ModalDOFArray`? Then each connection or whatnot can check if it got the right subclass. Some nice side effects
* the arithmetic in `arraycontext` checks `other.__class__ is cls`, so we wouldn't be able to do arithmetic on modal and nodal `DOFArray`s together by construction
* typing annotations?
* easy to inspect by printing the name vs going through the discretization keys.

inducer:
I agree that using type annotations to check this statically would be appealing.

- The main difficulty I see with the approach you propose is that, technically, modal-vs-nodal is a per-group decision, and pretending otherwise is (IMO) going to lead to trouble. The first yuckiness I can foresee is having to check all element groups for modal-/nodal-ness... and then do you raise an error if it's not unanimous?
- Modal vs nodal is also not really the only DOF-changes-meaning problem we might have. Suppose we have two different element groups with identical node counts, but (say) different nodal locations. Right now, nothing prevents passing `DOFArray`s from one to the other, typically dropping to first order as a result. The `discretization-key tracking might stand a chance at catching these issues before they ruin someone's afternoon.

alexfikl:
> The main difficulty I see with the approach you propose is that, technically, modal-vs-nodal is a per-group decision, and pretending otherwise is (IMO) going to lead to trouble. The first yuckiness I can foresee is having to check all element groups for modal-/nodal-ness... and then do you raise an error if it's not unanimous?

What's the usecase for mixing modal and nodal values in the same `DOFArray`?

Also, if it's a per-group decision, what would passing a mixed `DOFArray` to a connection do? Would it only work on the groups it can or error out?

> Suppose we have two different element groups with identical node counts, but (say) different nodal locations. Right now, nothing prevents passing DOFArrays from one to the other, typically dropping to first order as a result.

Oh yeah, that would be very cool to check somehow!

[...]

PR #264, opened 2021-09-13
Check `DOFArray` group array sizes in connections and visualization

inducer:
I observed that it's currently possible to feed a `DOFArray` for a non-matching discretization to a visualizer or a connection. This adds some checks to catch that. That isn't free, but IMO it's well worth the cost. I'm honestly almost expecting to see some failures in CI, which would immediately prove the worth of these changes.

### DIFF
--- a/meshmode/discretization/connection/direct.py
+++ b/meshmode/discretization/connection/direct.py
@@ -325,11 +325,9 @@ class DirectDiscretizationConnection(DiscretizationConnection):
         if is_array_container(ary) and not isinstance(ary, DOFArray):
             return map_array_container(self, ary)
 
-        if not isinstance(ary, DOFArray):
-            raise TypeError("non-array passed to discretization connection")
-
-        if ary.shape != (len(self.from_discr.groups),):
-            raise ValueError("invalid shape of incoming resampling data")
+        if __debug__:
+            from meshmode.dof_array import check_dofarray_against_discr
+            check_dofarray_against_discr(self.from_discr, ary)
 
         if (ary.array_context.permits_inplace_modification
                 and not _force_no_inplace_updates):
--- a/meshmode/discretization/visualization.py
+++ b/meshmode/discretization/visualization.py
@@ -49,7 +49,7 @@ __doc__ = """
 def separate_by_real_and_imag(names_and_fields, real_only):
     """
     :arg names_and_fields: input data array must be already flattened into a
-        single :mod:`numpy` array using :func:`resample_to_numpy`.
+        single :mod:`numpy` array using :func:`_resample_to_numpy`.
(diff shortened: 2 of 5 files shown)


meshmode/discretization/connection/direct.py
```diff
@@ -331,6 +331,12 @@ def __call__(self, ary, _force_no_inplace_updates=False):
         if ary.shape != (len(self.from_discr.groups),):
             raise ValueError("invalid shape of incoming resampling data")
 
+        for i, (grp, grp_ary) in enumerate(zip(self.from_discr.groups, ary)):
```
alexfikl, line 334: Maybe add this under an `if __debug__`? Not quite sure how big of a cost it would be when the number of groups grows a bit, but connections are called a lot.

alexfikl (commented):
Looks good to me!

Is it worth promoting that function in `visualization.py` to `discretization.__init__.py`? I can the projection connection also checking this and we can add the `discretization_key` check in there too once something is done about #258.

meshmode/discretization/visualization.py
```diff
@@ -91,6 +91,19 @@ def _stack_object_array(vec, *, by_group=False):
         ])
 
 
+def _check_dofarray_sizes(discr, dof_ary):
```
alexfikl, line 94: Is it worth promoting this to `discretization.__init__.py`? I can the projection connection also checking this and we can add the `discretization_key` check in there too once something is done about #258.

meshmode/discretization/connection/direct.py
inducer: a568116

meshmode/discretization/visualization.py
inducer: a568116

inducer:
Thanks! Good ideas, I think I got all of them in a568116.

alexfikl (approved):
Yep, looks better now! :+1:

meshmode/dof_array.py
```diff
@@ -728,6 +731,40 @@ def array_context_for_pickling(actx: ArrayContext):
...
+    :arg discr: a :class:`~meshmode.discretization.Discretization`
+        against which *dof_ary* is to be checked.
+    """
+    if not isinstance(dof_ary, DOFArray):
+        raise TypeError("non-array passed to check_dofarray_against_discr")
+
+    if len(dof_ary) != len(discr.groups):
+        raise InconsistentDOFArray(
+                "DOFArray for visualization has unexpected "
+                f"number of groups "
```
alexfikl, line 755:
```suggestion
                "DOFArray has unexpected number of groups "
```
Missed tiny typo

PR #265, opened 2021-09-19
Rewrite more actx.call_loopy as actx.einsum

kaushikcfd:


thomasgibson (approved):
Thanks for doing this! Looks great to me!

examples/moving-geometry.py
```diff
@@ -93,11 +72,10 @@ def resample_nodes_to_mesh(grp, igrp, iaxis):
...
 
-        return actx.call_loopy(
-                resample_by_mat_prg(),
-                nodes=discr_nodes,
-                resampling_mat=to_mesh_interp_matrix(grp),
-                )["result"]
+        return actx.einsum("ij,ej->ei",
+                           to_mesh_interp_matrix(grp),
+                           discr_nodes,
+                           tagged=(FirstAxisIsElementsTag(),))
```
alexfikl, line 78: Just to make sure I understand what's going on here. 

This should end up tagging / splitting all the inames in the same way as before (because of [this](https://github.com/inducer/meshmode/blob/956f7ddaa7805db89b752494203adc11b77481c8/meshmode/array_context.py#L149-L152)), right? i.e. there shouldn't be any change in performance?

kaushikcfd: Yep, there shouldn't be any noticeable changes in performance for the PyOpenCLArrayContext exactly because of the code you pointed out.

PR #275, opened 2021-10-20
Update DOFArray flatten and unflatten

alexfikl:
Following up on https://github.com/inducer/arraycontext/pull/91#discussion_r716273028.

At the moment this just updates the docs in the `dof_array` flattening functions to also mention the ones in `arraycontext`. There's a few of these: `flatten`, `unflatten`, `unflatten_like`, `flatten_to_numpy` and `unflatten_from_numpy`; and all of them just work on `DOFArray`s and leave the rest of the container alone. 

How many of these should be deprecated? At least the `numpy` ones have some nice uses
* in some tests, but these can probably be replaced with some reshaping, e.g.
```python
nodes = thaw(discr.nodes(), actx)
nodes = np.stack(flatten_to_numpy(nodes, actx))
```

* `pytential` uses [`unflatten_from_numpy`](https://github.com/inducer/pytential/blob/f37481d39dfb230306e29e90a868d22c681fb546/pytential/symbolic/matrix.py#L208) in a couple of places. Not sure any of those actually flatten anything that's not a `DOFArray` though.

* `pytential` also uses [`flatten`](https://github.com/inducer/pytential/blob/f37481d39dfb230306e29e90a868d22c681fb546/pytential/qbx/__init__.py#L777) quite a lot to talk to `sumpy`. This could be replaced by `arraycontext.flatten` + reshape too, to make it `(3, nparticles)`, I think, since `sumpy` supports both types of inputs.

### DIFF
--- a/meshmode/discretization/connection/face.py
+++ b/meshmode/discretization/connection/face.py
@@ -114,6 +114,10 @@ def _get_face_vertices(mesh, boundary_tag):
     if boundary_tag not in [FACE_RESTR_INTERIOR, FACE_RESTR_ALL]:
         # {{{ boundary faces
 
+        from meshmode.mesh import mesh_has_boundary
+        if not mesh_has_boundary(mesh, boundary_tag):
+            raise ValueError(f"invalid boundary tag {boundary_tag}.")
+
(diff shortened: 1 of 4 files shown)


inducer:
Yep, I say let's remove them from the docs as well.

PR #283, opened 2021-11-07
Change MeshElementGroups from Record to dataclass

alexfikl:
Gave this a try and it seems to not break the world. Yey!

Mostly works towards #283.

### DIFF
--- a/meshmode/mesh/__init__.py
+++ b/meshmode/mesh/__init__.py
@@ -221,16 +221,24 @@ class MeshElementGroup(ABC):
     # NOTE: the mesh supports not having vertices if no facial or nodal
     # adjacency is required, so we can mark this as optional
     vertex_indices: Optional[np.ndarray]
-    nodes: np.ndarray
 
-    # NOTE: unit_nodes are "optional" because subclasses may generate default
-    # unit nodes from the order and other parameters
-    unit_nodes: Optional[np.ndarray] = None
+    nodes: np.ndarray
+    unit_nodes: np.ndarray
 
     # FIXME: these should be removed!
     # https://github.com/inducer/meshmode/issues/224
     element_nr_base: Optional[int] = None
     node_nr_base: Optional[int] = None
+
+    # TODO: remove when everything has been constructed through the factory
+    _factory_constructed: bool = False
+
+    def __post_init__(self):
+        if not self._factory_constructed:
+            from warnings import warn
+            warn(f"Calling the constructor of '{type(self).__name__}' is "
+                 "deprecated. Use '{type(self).__name__}.make_group' instead",
+                 DeprecationWarning, stacklevel=2)
 
     @property
     def dim(self):
@@ -316,6 +324,11 @@ class MeshElementGroup(ABC):
             coordinates of each vertex.
         """
 
+    @classmethod
+    @abstractmethod
+    def make_group(cls, **kwargs: Any) -> "MeshElementGroup":
+        pass
+
 # }}}
 
 
@@ -333,14 +346,18 @@ class _ModepyElementGroup(MeshElementGroup):
     .. attribute:: _modepy_space
     """
 
+    # TODO: remove once `make_group` is used everywhere
     dim: Optional[int] = None
-    unit_nodes: Optional[np.ndarray] = None
 
     _modepy_shape_cls: ClassVar[Type[mp.Shape]] = mp.Shape
-    _modepy_shape: mp.Shape = field(default=None, init=False, repr=False)
-    _modepy_space: mp.FunctionSpace = field(default=None, init=False, repr=False)
+    _modepy_shape: mp.Shape = field(default=None, repr=False)
+    _modepy_space: mp.FunctionSpace = field(default=None, repr=False)
 
     def __post_init__(self):
+        super().__post_init__()
+        if self._factory_constructed:
+            return
+
         if self.unit_nodes is None:
             if self.dim is None:
                 raise TypeError("either 'dim' or 'unit_nodes' must be provided")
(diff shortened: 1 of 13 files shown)


alexfikl:
As a side note, tried converting `Mesh` to be a dataclass too, but the init for that is way too complex to easily translate it.

One big issue was that some arguments like `skip_tests` are theoretically `InitVar` with default values, but dataclasses didn't like that until [recently](https://github.com/python/cpython/pull/20867) :cry:

meshmode/mesh/__init__.py
```diff
@@ -155,98 +156,78 @@ class BTAG_INDUCED_BOUNDARY(BTAG_NO_BOUNDARY):  # noqa: N801
...
-
-        super().__init__(
-            order=order,
-            vertex_indices=vertex_indices,
-            nodes=nodes,
-            unit_nodes=unit_nodes,
-            element_nr_base=element_nr_base, node_nr_base=node_nr_base,
-            **kwargs)
+    order: int
+    vertex_indices: Optional[np.ndarray]
```
inducer, line 215: How come this is marked `Optional`?

meshmode/mesh/__init__.py
```diff
@@ -293,104 +284,162 @@ def __eq__(self, other):
...
+            of an element in this group.
+        """
+        raise NotImplementedError
+
+    def vertex_unit_coordinates(self) -> np.ndarray:
+        """
+        :returns: an array of shape ``(nfaces, dim)`` with the unit
+            coordinates of each vertex.
+        """
+        raise NotImplementedError
```
inducer, line 304: Make this an ABC?

meshmode/mesh/__init__.py
```diff
@@ -293,104 +284,162 @@ def __eq__(self, other):
...
-            :func:`modepy.edge_clustered_nodes_for_space` are assumed.
-            These must be in unit coordinates as defined in :mod:`modepy`.
-        :arg dim: only used if *unit_nodes* is *None*, to get
-            the default unit nodes.
-
-        Do not supply *element_nr_base* and *node_nr_base*, they will be
-        automatically assigned.
-        """
+    shape_cls: ClassVar[Type[mp.Shape]] = mp.Shape
+    shape_dim: Optional[int] = None
```
inducer, line 314: Why optional?

meshmode/mesh/__init__.py
```diff
@@ -293,104 +284,162 @@ def __eq__(self, other):
...
+
+    @property
+    def _modepy_space(self):
+        return self.space
 
     @property
-    @memoize_method
     def _modepy_faces(self):
-        return mp.faces_for_shape(self._modepy_shape)
+        return self.faces
```
inducer, line 372: I guess I'm on board with un-underscoring these, but then these should be deprecated.

meshmode/mesh/__init__.py
```diff
@@ -412,6 +461,18 @@ class NodalAdjacency(Record):
     .. automethod:: __ne__
     """
 
+    neighbors_starts: np.ndarray
+    neighbors: np.ndarray
+
+    def copy(self, **kwargs: Any) -> "NodalAdjacency":
+        from warnings import warn
+        warn(f"{type(self).__name__}.copy is deprecated and will be removed in 2022."
```
inducer, line 469:
```suggestion
        warn(f"{type(self).__name__}.copy is deprecated and will be removed in July 2022."
```

inducer (commented):
Thanks for working on this! Looks good except for these (hopefully) minor wrinkles.

inducer:

meshmode/mesh/__init__.py
alexfikl: We made vertices optional a while back in here
https://gitlab.tiker.net/inducer/meshmode/-/merge_requests/66

meshmode/mesh/__init__.py
alexfikl: Done in 4b41215.

meshmode/mesh/__init__.py
alexfikl: Un-un-underscored them in 0807521. There was no reason for the name change and un-underscoring in the first place.

meshmode/mesh/__init__.py
alexfikl: Reworked this a bit in 0807521. It seems to work and hopefully make this clearer.

alexfikl:
This should be ready for another look. I still don't like the big `__post_init__` in the modepy element groups, but it's the same as before.

meshmode/mesh/__init__.py
```diff
@@ -293,104 +297,133 @@ def __eq__(self, other):
...
+    dim: Optional[int] = None
+    unit_nodes: Optional[np.ndarray] = None
+
+    _modepy_shape_cls: ClassVar[Type[mp.Shape]] = mp.Shape
+    _modepy_shape: mp.Shape = field(default=None, init=False, repr=False)
+    _modepy_space: mp.FunctionSpace = field(default=None, init=False, repr=False)
 
-        if unit_nodes is None:
-            unit_nodes = mp.edge_clustered_nodes_for_space(space, shape)
+    def __post_init__(self):
```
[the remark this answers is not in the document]
alexfikl: @inducer Hm, not quite sure how to implement that in a vaguely backwards compatible way though. Any suggestions / preferences?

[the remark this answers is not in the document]
inducer: Maybe make a no-op factory function for now and raise a `DeprecationWarning` for construction that doesn't go through the function?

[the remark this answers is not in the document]
alexfikl: Tried it in 2ecce05. Let me know if that's more or less what you had in mind.

inducer (approved):
Thanks! LGTM, queued to go in after #299.

PR #285, opened 2021-11-16
Allow boundaries on interior faces

majosm:
@inducer I'm guessing you probably won't need this for the multi-domain wave demo, but just in case.

### DIFF
--- a/meshmode/mesh/__init__.py
+++ b/meshmode/mesh/__init__.py
@@ -1283,44 +1283,44 @@ def _compute_facial_adjacency_from_vertices(
                 face_id_pairs[0].faces[is_neighbor_adj],
                 face_id_pairs[0].elements[is_neighbor_adj]] = True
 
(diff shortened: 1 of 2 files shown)


meshmode/mesh/__init__.py
```diff
@@ -1283,44 +1283,44 @@ def _compute_facial_adjacency_from_vertices(
...
-                    grp_list.append(
-                        BoundaryAdjacencyGroup(
-                            igroup=igrp,
-                            boundary_tag=btag,
-                            elements=elements,
-                            element_faces=element_faces))
-
-            is_untagged = ~np.any(belongs_to_bdry, axis=0)
-            if np.any(is_untagged):
+        belongs_to_bdry = np.full(
```
inducer, line 1286: The tagged interfaces should also be removed from the "normal" facial adjacency, to avoid accidentally computing fluxes across these boundary interfaces.

PR #286, opened 2021-11-18
Disallow complex fields in matplotlib visualizer

alexfikl:
Not quite sure what it was doing before (just taking the real part somewhere?), but in matplotlib 3.5 they added an extra check that raises on complex inputs.


inducer:
Thanks!

PR #288, opened 2021-12-12
Improve torus parametrization

alexfikl:
This reworks the torus construction to have smoother looking element boundaries (the commit message has a longer explanation).

As far as I can tell, both were "high-order" (we have some tests in pytential that check the curvature), but this is cleaner.

TODO
- [x] Needs inducer/pytential#125.
- [ ] Point `requirements.txt` back to `pytential@main`.


alexfikl:
<p align="center">

</p>

This is how the torus looks now. I checked the "stretch factor" in pytential and it's also smooth looking on this mesh. Not sure what it didn't like about the previous one though..

inducer:
Thanks!

PR #290, opened 2021-12-31
Use space_for_shape in make_group_from_vertices

alexfikl:
It now doesn't assume `QN` and should work with the new tensor product spaces from inducer/modepy#41.


inducer:
Thx!

PR #291, opened 2022-01-03
Use space_for_shape in tessellation

alexfikl:
Missed a few more issues in #290.

### DIFF
--- a/meshmode/discretization/visualization.py
+++ b/meshmode/discretization/visualization.py
@@ -301,7 +301,9 @@ class VTKConnectivity:
 
         if isinstance(grp.mesh_el_group, _ModepyElementGroup):
             shape = grp.mesh_el_group._modepy_shape
-            space = type(grp.mesh_el_group._modepy_space)(grp.dim, grp.order)
+            space = mp.space_for_shape(shape, grp.order)
+            assert type(space) == type(grp.mesh_el_group._modepy_space)  # noqa: E721
+
             node_tuples = mp.node_tuples_for_space(space)
 
             el_connectivity = np.array(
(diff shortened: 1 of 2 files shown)


meshmode/discretization/visualization.py
```diff
@@ -301,7 +301,9 @@ def connectivity_for_element_group(self, grp):
 
         if isinstance(grp.mesh_el_group, _ModepyElementGroup):
             shape = grp.mesh_el_group._modepy_shape
-            space = type(grp.mesh_el_group._modepy_space)(grp.dim, grp.order)
+            space = mp.space_for_shape(shape, grp.order)
+            assert type(space) == type(grp.mesh_el_group._modepy_space)  # noqa: E721
```
alexfikl, line 305: I added an assert here because, at least theoretically, `_modepy_space` could be different than whatever the default is in `space_for_shape`.

The common usage seems to be "given me the same function space, but with a different order". Maybe it's worth adding a `with_order` to `FunctionSpace`?

inducer: Good idea! https://github.com/inducer/modepy/issues/50

meshmode/mesh/refinement/tessellate.py
```diff
@@ -233,15 +235,16 @@ def _(meg: _ModepyElementGroup, el_tess_info, elements):
...
+    space = mp.space_for_shape(shape, 2)
+    assert type(space) == type(meg._modepy_space)       # noqa: E721
 
     ref_vertices = mp.node_tuples_for_space(space)
     ref_vertices_to_index = {rv: i for i, rv in enumerate(ref_vertices)}
 
     from pytools import add_tuples
-    space = type(meg._modepy_space)(meg.dim, 1)
+    space = mp.space_for_shape(shape, 1)
     orig_vertices = tuple([
```
inducer, line 248:
```suggestion
    assert type(space) == type(meg._modepy_space)  # noqa: E721
    orig_vertices = tuple([
```

inducer:
LGTM, thanks!

PR #292, opened 2022-01-24
Add volume groups

majosm:
Taking another stab at region tagging, since #129 is way out of date. Still need to implement the mesh partitioning part.


meshmode/mesh/__init__.py
```diff
@@ -497,6 +498,35 @@ class TensorProductElementGroup(_ModepyElementGroup):
 # }}}
 
 
+# {{{ volume groups
+
+@dataclass(frozen=True)
+class VolumeGroup:
```
inducer, line 504:
```suggestion
class TaggedElementSubgroup:
```
?
Or `ElementSubgroup`?

majosm:
After thinking about this more (and finding some bugs in my implementation) I'm starting to wonder if it would be better to just have `read_gmsh` return the tag information separately. I'm going to take a stab at that and see how it looks.

inducer:
> After thinking about this more (and finding some bugs in my implementation) I'm starting to wonder if it would be better to just have `read_gmsh` return the tag information separately. I'm going to take a stab at that and see how it looks.

I'm liking that idea.

PR #294, opened 2022-02-07
Add support for VTKHDF

alexfikl:
Found out that VTK has support for HDF5 too (not sure since when?)
https://vtk.org/doc/nightly/html/VTKHDFFileFormat.html
and implemented a little writer for it. It should be better than the XDMF writer in the long run, since that seems generally unmaintained.

It has similar benefits with smaller file sizes. From the test in `test_visualization`
```
2.0M visualizer_vtk_simplex_3d_linear.vtu
1.5M visualizer_vtkhdf_simplex_3d_linear.hdf
```

Big plus is that it supports MPI too so there's no need for that `pvtu` stuff with a gazillion files!

EDIT: Tested with Paraview 5.10 and no guarantees for other version.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,6 +16,7 @@ a.out
 *.pvtu
 *.h5
 *.xmf
+*.hdf
 
 .cache
 .pytest_cache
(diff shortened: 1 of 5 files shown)


inducer:
LGTM, thanks!

cc @MTCam @majosm @anderson2981

PR #295, opened 2022-02-09
Pad arrays to 3d in write_vtkhdf_file

alexfikl:
Just realized I didn't really check the 2D case.. and indeed it wasn't working. 

Paraview just crashed in some `vtkPoints` class saying it didn't have enough dimensions. This seems to fix it, but it's not great to store a bunch of zeros. Not sure if there's some better way..


inducer:
I think the other Vtk formats are the same way. 2D typically isn't what's pushing up against scalability limits, so I'm not too worried. Thanks for checking and pushing this fix!

PR #296, opened 2022-02-09
Fix VTKHDF visualization on Lagrange elements

alexfikl:
Found another issue when trying to use the high-order Lagrange elements. Apparently the offsets need to start at 0.


alexfikl:
This seems to have fixed the Gitlab CI as well
https://gitlab.tiker.net/fikl2/meshmode/-/pipelines/262609

inducer:
Thx!

PR #300, opened 2022-02-15
Remove some deprecated calls

alexfikl:


inducer:
LGTM, thanks!

meshmode/discretization/connection/direct.py
```diff
@@ -642,11 +642,16 @@ def knl():
...
-                "...",
+                ...,
             ],
             name="oversample_mat"
         )
 
+        return lp.tag_inames(t_unit, {
+                "iel": ConcurrentElementInameTag(),
+                "idof": ConcurrentDOFInameTag(),
+                "jdof": ConcurrentDOFInameTag()})
```
alexfikl, line 653: @inducer This isn't right. Should it just have `idof` be concurrent?

inducer: I mean, it's "right" in the literal sense, I think. They're both concurrent inames. What made you say otherwise? It might not work well with the current transform stack (which is a separate bug).

inducer: Just the fact that the current transform stack can't handle it? If so, just mark `idof` with `ConcurrentDOFInameTag` and add a FIXME that `jdof` technically should also be tagged that.

alexfikl: Yeah, I meant that it confused the transform at the moment.

I'll add a `FIXME`. Thanks!

alexfikl: 86a02bb

Issue #224, opened 2021-06-16
Element groups shouldn't have to change when joining a mesh/discretization


inducer:
That means that things like `index` and `element_nr_base` and such shouldn't exist. They belong in the mesh or discretization.

We should deprecate and eventually remove these fields.

majosm:
Can this be closed? (Noticed while scanning the issues list for undergrad project ideas. 🙂)

PR #302, opened 2022-02-20
Remove uses of MeshElementGroup.element_nr_base

alexfikl:
* Adds a `Mesh.element_nr_base` and `Mesh.node_nr_base`, which are arrays of size `len(groups)`.
* Uses those everywhere instead of `MeshElementGroup.element_nr_base` and `node_nr_base`, except in `refinement.Refiner` since that's supposed to go away anyway.

`element_nr_base` and `node_nr_base` are still set in `Mesh.__init__` in case anyone uses them, so not quite sure how to deprecate them.

Work towards #224.

### DIFF
--- a/meshmode/mesh/__init__.py
+++ b/meshmode/mesh/__init__.py
@@ -257,6 +257,15 @@ class MeshElementGroup(ABC):
                  f"Use '{type(self).__name__}.make_group' instead",
                  DeprecationWarning, stacklevel=2)
 
+    def __getattribute__(self, name):
+        if name in ("element_nr_base", "node_nr_base"):
+            from warnings import warn
+            warn(f"'{type(self).__name__}.{name}' is deprecated and will be "
+                 f"removed in July 2022. Use 'Mesh.{name}' instead",
+                 DeprecationWarning, stacklevel=2)
+
+        return super().__getattribute__(name)
+
     @property
     def dim(self):
         return self.unit_nodes.shape[0]
@@ -848,6 +857,16 @@ class Mesh(Record):
 
     .. attribute:: nelements
 
+    .. attribute:: base_element_nrs
+
+        An array of size ``(len(groups),)`` of starting element indices for
+        each group in the mesh.
+
+    .. attribute:: base_node_nrs
+
+        An array of size ``(len(groups),)`` of starting node indices for
+        each group in the mesh.
+
     .. attribute:: nodal_adjacency
 
         An instance of :class:`NodalAdjacency`.
@@ -1071,6 +1090,16 @@ class Mesh(Record):
         return sum(grp.nelements for grp in self.groups)
 
     @property
+    @memoize_method
+    def base_element_nrs(self):
+        return np.cumsum([0] + [grp.nelements for grp in self.groups[:-1]])
+
+    @property
+    @memoize_method
+    def base_node_nrs(self):
+        return np.cumsum([0] + [grp.nnodes for grp in self.groups[:-1]])
+
+    @property
     def nodal_adjacency(self):
         from meshmode import DataUnavailable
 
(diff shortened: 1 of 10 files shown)


meshmode/discretization/__init__.py
```diff
@@ -469,6 +469,11 @@ def ambient_dim(self):
     def ndofs(self):
         return sum(grp.ndofs for grp in self.groups)
 
+    @property
+    @memoize_method
+    def dof_nr_base(self):
```
inducer, line 474: Underscore? Since our `DOFArray`s are nested, I'm not sure there's widespread need for this.

meshmode/mesh/__init__.py
```diff
@@ -1070,6 +1070,16 @@ def nvertices(self):
...
+    @property
+    @memoize_method
+    def node_nr_base(self):
+        return np.cumsum([0] + [grp.nnodes for grp in self.groups[:-1]])
+
+    @property
+    @memoize_method
+    def element_nr_base(self):
+        return np.cumsum([0] + [grp.nelements for grp in self.groups[:-1]])
+
```
inducer, line 1082: Document?

inducer, line 1082: Raise a `DeprecationWarning` on accessing `MeshElementGroup.element_nr_base`? (Deadline mid-this-year?)

inducer (commented):
Thanks for working on this! A few maintainability concerns (plus I need to do a detailed read), otherwise good from a first impression.

meshmode/discretization/__init__.py
alexfikl: Removed this completely in the latest version. It was only used in the visualization code a few times, so probably not worth it to put it here.

meshmode/mesh/__init__.py
alexfikl: Added docs in latest version.

alexfikl: Done.

meshmode/mesh/__init__.py
```diff
@@ -848,6 +857,16 @@ class Mesh(Record):
 
     .. attribute:: nelements
 
+    .. attribute:: element_nr_base
```
inducer, line 860:
```suggestion
    .. attribute:: base_element_nrs
```
? (to convey plural and avoid confusion with existing attribute)

meshmode/mesh/__init__.py
```diff
@@ -848,6 +857,16 @@ class Mesh(Record):
...
+
+        An array of size ``(len(groups),)`` of starting element indices for
+        each group in the mesh.
+
+    .. attribute:: node_nr_base
```
inducer, line 865:
```suggestion
    .. attribute:: base_node_nrs
```
? (to convey plural and avoid confusion with existing attribute)

meshmode/mesh/__init__.py
```diff
@@ -257,6 +257,15 @@ def __post_init__(self):
                  f"Use '{type(self).__name__}.make_group' instead",
                  DeprecationWarning, stacklevel=2)
 
+    def __getattribute__(self, name):
```
inducer, line 260: Any way to do this without causing a high cost on every attribute access? (It's tolerable, since `Mesh` isn't *super* high traffic, but it's still better to avoid IMO.)

inducer (commented):
Sorry, got a naming complaint. I know I could've thought of this one earlier.

meshmode/mesh/__init__.py
alexfikl: Couldn't really think of another way to do this since `element_nr_base` is part of the dataclass fields, so we can't make it a property to add the deprecation warning :\

meshmode/mesh/__init__.py
alexfikl: Renamed in latest version.

meshmode/mesh/__init__.py
alexfikl: Done.

meshmode/mesh/__init__.py
inducer: Yeah, me either. Let's leave it for now.

inducer:
Thanks, LGTM!

PR #304, opened 2022-02-23
Less noisy element_nr_base deprecation

alexfikl:
* `join_mesh` accessed `element_nr_base` and complained on every mesh construction.
* `__eq__` also compared `element_nr_base`, which shouldn't be valid anymore.

Also fixed the deprecation warning, which wasn't pointing to the correct attribute on `Mesh`.


inducer:
Thanks!

PR #305, opened 2022-02-27
Deprecate ElementGroupBase.index

alexfikl:
This deprecates the `index` attribute
* Set `index=None` in `ElementGroupBase` and issue a deprecation warning.
* Set `index=None` in `ElementGroupFactory.__call__` too.

The last two commits are independent and just do some random cleanup in `meshmode.discretization.__init__`. I can pluck those out into a separate PR if that's easier.

Second part of #224.

### DIFF
--- a/meshmode/discretization/__init__.py
+++ b/meshmode/discretization/__init__.py
@@ -23,19 +23,18 @@ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 THE SOFTWARE.
 """
 
-from typing import Iterable
+from abc import ABCMeta, abstractproperty, abstractmethod
+from warnings import warn
+from typing import Hashable, Iterable, Optional
+
 import numpy as np
 
-from abc import ABCMeta, abstractproperty, abstractmethod
+import loopy as lp
+from arraycontext import ArrayContext, make_loopy_program
 from pytools import memoize_in, memoize_method, keyed_memoize_in
 from pytools.obj_array import make_obj_array
-from arraycontext import ArrayContext, make_loopy_program
-
-import loopy as lp
 from meshmode.transform_metadata import (
         ConcurrentElementInameTag, ConcurrentDOFInameTag, FirstAxisIsElementsTag)
-
-from warnings import warn
 
 # underscored because it shouldn't be imported from here.
 from meshmode.dof_array import DOFArray as _DOFArray
@@ -87,7 +86,6 @@ class ElementGroupBase(metaclass=ABCMeta):
 
     .. attribute :: mesh_el_group
     .. attribute :: order
-    .. attribute :: index
 
     .. autoattribute:: is_affine
     .. autoattribute:: nelements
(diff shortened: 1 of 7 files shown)


meshmode/discretization/__init__.py
```diff
@@ -101,11 +107,15 @@ class ElementGroupBase(metaclass=ABCMeta):
...
-            :class:`~meshmode.mesh.MeshElementGroup`.
-        """
+    def __init__(self,
+                 mesh_el_group: _MeshElementGroup,
+                 order: int,
+                 index: Optional[int] = None) -> None:
+        if index is not None:
+            warn("Setting 'index' is deprecated and will be removed in 2023. "
+                 "This is now handled internally by the discretization.",
+                 DeprecationWarning, stacklevel=2)
```
alexfikl, line 117: Would it be better to raise the warning only when someone tries to use `grp.index`? Here it's easy to wrap it up in a property and do that.

meshmode/discretization/__init__.py
```diff
@@ -125,7 +135,8 @@ def nelements(self):
         """
         return self.mesh_el_group.nelements
 
-    @abstractproperty
+    @property
+    @abstractmethod
```
inducer, line 139: Why this change? (also below)

meshmode/discretization/__init__.py
```diff
@@ -172,6 +185,20 @@ def discretization_key(self):
         """
         return (type(self), self.dim, self.order)
 
+
+class ElementGroupFactory:
```
inducer, line 189: Make this a protocol (cf. `typing.Protocol`)? (in preparation for typed `meshmode`) The intention from my end was very much that a plain callable can be a element group factory.

meshmode/discretization/__init__.py
```diff
@@ -396,26 +424,31 @@ class Discretization:
...
+                 mesh: _Mesh,
+                 group_factory: ElementGroupFactory,
+                 real_dtype: Optional[np.dtype] = None,
+                 _force_actx_clone: bool = True) -> None:
         """
-        :arg actx: A :class:`ArrayContext` used to perform computation needed
-            during initial set-up of the mesh.
-        :arg mesh: A :class:`meshmode.mesh.Mesh` over which the discretization is
+        :arg actx: an :class:`arraycontext.ArrayContext` used to perform
+            computation needed during initial set-up of the mesh.
```
inducer, line 434:
```suggestion
            computation needed during initial set-up of the discretization.
```

inducer (commented):
Thanks for working on this. Just a few things below.

meshmode/discretization/__init__.py
alexfikl: This found a bug too!
https://github.com/inducer/meshmode/compare/3f463c49f29f1ea436ca9b05cf42e285bcce62b3..9797d4775cde156d2bc3de883adfdd9aba9f778b

meshmode/discretization/__init__.py
inducer: Huh, TIL. Thanks!

inducer:
Thanks!

PR #307, opened 2022-03-15
Allow distributed partition boundaries to be not just rank boundaries

inducer:
To do before merging:

- [x] Decide on #308
- [ ] ~~Decide on #309~~ (we'll merge this separately)

cc @majosm

PR #308, opened 2022-03-16
Recast `partition_mesh` in terms of generic part identifiers

majosm:
Should remove the need for a custom partition boundary tag in grudge.

Targets #307.


majosm:
@inducer Is there a way I can get the CI to run on this? (Not sure why it isn't; maybe because the target branch isn't main?)

meshmode/mesh/processing.py
```diff
@@ -409,98 +430,121 @@ def _create_boundary_groups(mesh, global_elem_to_part_elem, part_mesh_groups,
...
+            self_vertices,
+            self_mesh_groups,
+            facial_adjacency_groups=self_facial_adj_groups,
             is_conforming=mesh.is_conforming)
 
-    return part_mesh, queried_elems
+    return self_mesh
+
+
+def partition_mesh(mesh, part_id_to_elements):
```
inducer, line 535: This is incompatible with the prior interface, right?

majosm: Yeah. Though grudge and mirgecom at least don't use it directly; they use `MPIMeshDistributor`, which (unless I've messed something up) is still backwards compatible.

inducer: OK, fair enough!

majosm:
I think this is ready for a look, apart from the doc failures (which I'm not sure how to fix).

alexfikl:
> I think this is ready for a look, apart from the doc failures (which I'm not sure how to fix).

You need to add `mpi4py` to intersphinx in `docs/conf.py` for that to work. Probably something like
```
"https://mpi4py.readthedocs.io/en/stable": None
```

meshmode/mesh/processing.py
```diff
@@ -183,20 +188,23 @@ def _filter_mesh_groups(mesh, selected_elements, vertex_id_dtype):
...
-    :arg global_elem_to_part_elem: A :class:`numpy.ndarray` mapping from global
-        element index to local partition-wide element index for local elements (and
-        -1 otherwise).
-
-    :returns: A :class:`set` of indices of the neighboring partitions.
+    :arg part_id_to_part_index: A mapping from partition identifiers to indices in
+        the range ``[0, num_parts)``.
+    :arg global_elem_to_part_elem: A :class:`numpy.ndarray` that maps global element
+        indices to partition indices and partition-wide element indices. See
+        :fun:`_compute_global_elem_to_part_elem`` for details.
```
alexfikl, line 200:
```suggestion
        :func:`_compute_global_elem_to_part_elem`` for details.
```

meshmode/mesh/processing.py
```diff
@@ -208,46 +216,52 @@ def _get_connected_partitions(
...
+    Create self-to-self facial adjacency groups for a partitioned mesh.
 
     :arg mesh: A :class:`~meshmode.mesh.Mesh` representing the unpartitioned mesh.
-    :arg global_elem_to_part_elem: A :class:`numpy.ndarray` mapping from global
-        element index to local partition-wide element index for local elements (and
-        -1 otherwise).
-    :arg part_mesh_groups: An array of :class:`~meshmode.mesh.ElementGroup` instances
+    :arg global_elem_to_part_elem: A :class:`numpy.ndarray` that maps global element
+        indices to partition indices and partition-wide element indices. See
+        :fun:`_compute_global_elem_to_part_elem`` for details.
```
alexfikl, line 245:
```suggestion
        :func:`_compute_global_elem_to_part_elem`` for details.
```

meshmode/mesh/processing.py
```diff
@@ -257,77 +271,79 @@ def _create_local_to_local_adjacency_groups(mesh, global_elem_to_part_elem,
...
-        partition numbers.
-    :arg global_elem_to_part_elem: A :class:`numpy.ndarray` mapping from global
-        element index to local partition-wide element index for local elements (and
-        -1 otherwise).
-    :arg part_mesh_groups: An array of `~meshmode.mesh.ElementGroup` instances
+    :arg part_id_to_part_index: A mapping from partition identifiers to indices in
+        the range ``[0, num_parts)``.
+    :arg global_elem_to_part_elem: A :class:`numpy.ndarray` that maps global element
+        indices to partition indices and partition-wide element indices. See
+        :fun:`_compute_global_elem_to_part_elem`` for details.
```
alexfikl, line 324:
```suggestion
        :func:`_compute_global_elem_to_part_elem`` for details.
```

meshmode/mesh/processing.py
```diff
@@ -343,63 +359,68 @@ def _create_nonlocal_adjacency_groups(
...
     Create boundary groups for partitioned mesh.
 
     :arg mesh: A :class:`~meshmode.mesh.Mesh` representing the unpartitioned mesh.
-    :arg global_elem_to_part_elem: A :class:`numpy.ndarray` mapping from global
-        element index to local partition-wide element index for local elements (and
-        -1 otherwise).
-    :arg part_mesh_groups: An array of `~meshmode.mesh.ElementGroup` instances
+    :arg global_elem_to_part_elem: A :class:`numpy.ndarray` that maps global element
+        indices to partition indices and partition-wide element indices. See
+        :fun:`_compute_global_elem_to_part_elem`` for details.
```
alexfikl, line 404:
```suggestion
        :func:`_compute_global_elem_to_part_elem`` for details.
```

meshmode/distributed.py
```diff
@@ -416,39 +412,24 @@ def get_partition_by_pymetis(mesh, num_parts, *, connectivity="facial", **kwargs
...
-        else:
-            raise ValueError("unexpected inter-partition boundary tag type found")
 
+def membership_list_to_sets(membership_list):
     return {
-            _get_neighbor_part_nr(grp.boundary_tag)
-            for fagrp_list in mesh.facial_adjacency_groups
-            for grp in fagrp_list
-            if isinstance(grp, InterPartitionAdjacencyGroup)}
+        entry: set(np.where(membership_list == entry)[0])
```
alexfikl, line 423: Does this need to be a `set`? It seems like there's a lot of places lower down that would like it to be an `ndarray` (might have missed some places where it's better to have it as a set though).

alexfikl (commented):
I'm not familiar with the MPI parts of meshmode, but was looking a bit through this and left some nitpicks!

inducer:
Sorry about the conflicts! Turns out I had some uncommitted work lying around that was needed to get https://github.com/inducer/grudge/pull/236 to go.

majosm:
> I feel like this is headed for BTAG_PARTITION(BTAG_MULTIVOL_PARTITION(...)), which seems not that convenient, even if the inner thing were renamed. Can you explain a bit what your thought would be on how this would look in grudge?

I was picturing that `part_id` in the multivolume case would just be a tuple `(vol, rank)`. Then `BTAG_MULTIVOL_PARTITION(...)` would be replaced by `BTAG_PARTITION((vol, rank))`. Is that not going to work? I haven't looked through the grudge PR closely yet.

inducer:
> I was picturing that `part_id` in the multivolume case would just be a tuple `(vol, rank)`.

The issue I see is that, on many occasions, grudge (at least as currently done in https://github.com/inducer/grudge/pull/236, though I don't see how to avoid it) needs to pick apart the boundary tag to understand its components (to see which ranks/volumes are being connected), so making the parition IDs entirely opaque won't do.

majosm:
> The issue I see is that, on many occasions, grudge (at least as currently done in [inducer/grudge#236](https://github.com/inducer/grudge/pull/236), though I don't see how to avoid it) needs to pick apart the boundary tag to understand its components (to see which ranks/volumes are being connected), so making the parition IDs entirely opaque won't do.

Right, that sounds fine to me. The idea here was to make it opaque to _meshmode_ (and thus avoid the need to postprocess the partitioned mesh in order to relabel the tags). The grudge code can still use the tags however it wants.

inducer:
> stamp out remaining traces of fun

That's no fun! :P

meshmode/distributed.py
majosm: I think you're right. I thought I had a `<element> in <set>` in the code somewhere at one point, but it looks like it's gone. Appears that numpy arrays will work.

majosm:
> Unless we make grudge live with `BTAG_PARTITION(something)`,

That's kind of what I was thinking. I assume we will need a function in grudge that does the mesh partitioning anyway (using `partition_mesh` and `mpi_distribute` under the hood), so I figure at that time it can insert whatever data it needs into the partition ID and pull it out as needed later. All the uses of `<instance of BTAG_MULTIVOL_PARTITION>.<whatever>` would become `<instance of BTAG_PARTITION>.part_id.<whatever>`.

inducer:
Hmm. That's what I figured. The one fly in that ointment is that, currently, `BTAG_MULTIVOL_PARTITION` doesn't just identify the remote partition (as would be the case with your proposed approach), but also the local one (it has `self_volume_tag`). Representation-wise, that is sort of redundant, so I sort of see your point. Let me see how hard it is to get rid of `self_volume_tag` in https://github.com/inducer/grudge/pull/236.

majosm:
> Hmm. That's what I figured. The one fly in that ointment is that, currently, `BTAG_MULTIVOL_PARTITION` doesn't just identify the remote partition (as would be the case with your proposed approach), but also the local one (it has `self_volume_tag`). Representation-wise, that is sort of redundant, so I sort of see your point. Let me see how hard it is to get rid of `self_volume_tag` in [inducer/grudge#236](https://github.com/inducer/grudge/pull/236).

Would it possibly help if the facial adjacency groups had volumes associated with them? We'll need to read in the volume tags in `read_gmsh`, and at least the original plan was to store them inside the mesh (something like #292). We could potentially split the facial adjacency by volume tag the same way we do with boundary tags.

inducer:
> Let me see how hard it is to get rid of `self_volume_tag` in [inducer/grudge#236](https://github.com/inducer/grudge/pull/236).

The initial sense I'm getting is that this is a good direction. I should have that pushed in a bit.

meshmode/distributed.py
```diff
@@ -371,7 +366,8 @@ def complete_some(self):
 # }}}
 
 
-def get_partition_by_pymetis(mesh, num_parts, *, connectivity="facial", **kwargs):
+def get_partition_by_pymetis(
+        mesh, num_parts, *, connectivity="facial", return_sets=False, **kwargs):
```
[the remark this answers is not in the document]
majosm: Opted to get rid of it; wasn't need anymore.

inducer:
Urgh. So, in order to support the changes in https://github.com/inducer/grudge/pull/236 that we discussed (which I still like), I ended up making some changes to #307 that I think duplicate some of what you did here. IOW, I made another mess of conflicts. Sorry!

meshmode/mesh/processing.py
```diff
@@ -409,98 +430,120 @@ def _create_boundary_groups(mesh, global_elem_to_part_elem, part_mesh_groups,
...
-        numbers on *part_mesh* to ones in *mesh*.
+    :returns: A :class:`~meshmode.mesh.Mesh` containing a partition of *mesh*.
 
     .. versionadded:: 2017.1
     """
-    assert len(part_per_element) == mesh.nelements, (
-        "part_per_element must have shape (mesh.nelements,)")
+    nelements = sum(len(elems) for elems in part_id_to_elements.values())
+    assert nelements == mesh.nelements, \
+        "counts of partitioned elements must add up to mesh.nelements"
```
inducer, line 469: `raise` instead of `assert` for input checking. (Yes, I know it was bad before, too.)

meshmode/mesh/processing.py
```diff
@@ -80,25 +80,30 @@ def find_group_indices(groups, meshwide_elems):
 
 # {{{ partition_mesh
 
-def _compute_global_elem_to_part_elem(part_per_element, parts, element_id_dtype):
+def _compute_global_elem_to_part_elem(
```
inducer, line 83: Would you be willing to add types to these functions? https://github.com/Microsoft/pyright is the checker I used while we were working together (mypy is stricter but good too).

meshmode/mesh/processing.py
```diff
@@ -183,20 +188,23 @@ def _filter_mesh_groups(mesh, selected_elements, vertex_id_dtype):
 
 
 def _get_connected_partitions(
```
inducer, line 190:
```suggestion
def _get_connected_parts(
```

meshmode/mesh/processing.py
inducer, line 469: Anything here to ensure `part_id_to_elements` describe disjoint parts? `_compute_global_elem_to_part_elem` might be a natural spot for that.

meshmode/mesh/processing.py
```diff
@@ -257,77 +271,79 @@ def _create_local_to_local_adjacency_groups(mesh, global_elem_to_part_elem,
...
         instances corresponding to the entries in *mesh.facial_adjacency_groups* that
-        have non-local adjacency.
+        have self-to-other adjacency.
     """
-    global_elem_to_neighbor_elem = _compute_global_elem_to_part_elem(
-        part_per_element, connected_parts, mesh.element_id_dtype)
+    self_part_index = part_id_to_part_index[self_part_id]
 
-    nonlocal_adj_groups = [[] for _ in part_mesh_groups]
+    self_to_other_adj_groups = [[] for _ in self_mesh_groups]
```
inducer, line 342: Can you remind me why these functions return lists with only one element populated? (Maybe record the answer in a comment, in each function.)

meshmode/distributed.py
```diff
@@ -188,10 +190,6 @@ def make_remote_group_infos(
 @dataclass(init=True, frozen=True)
 class InterRankBoundaryInfo:
     """
-    .. attribute:: local_btag
-
-        A boundary tag for the local boundary towards the remote partition.
-
     .. attribute:: local_part_id
 
         An opaque, hashable, picklable identifier for the local partition.
```
inducer, line 195:
```suggestion
        An opaque, hashable, picklable identifier for the local part.
```
(also a few lines below) (Sorry, I know this is cleaning up my mess.)

meshmode/distributed.py
```diff
@@ -341,36 +337,40 @@ def complete_some(self):
 
         remote_to_local_bdry_conns = {}
 
-        local_part_id_to_irbi = {
-                irbi.local_part_id: irbi for irbi in self.inter_rank_bdry_info}
-        assert len(local_part_id_to_irbi) == len(self.inter_rank_bdry_info)
+        part_ids_to_irbi = {
+                (irbi.local_part_id, irbi.remote_part_id): irbi
+                for irbi in self.inter_rank_bdry_info}
+        assert len(part_ids_to_irbi) == len(self.inter_rank_bdry_info)
```
inducer, line 343: I guess this is input checking and should probably raise as a result.

meshmode/distributed.py
```diff
@@ -341,36 +337,40 @@ def complete_some(self):
...
 
         for i_src_rank, recvd in zip(
                 source_ranks, data):
```
inducer, line 346:
```suggestion
        for i_src_rank, recvd in zip(source_ranks, data):
```

inducer (commented):
Thanks for working on this! Looks good to me aside from these comments.

inducer:

majosm:
@inducer I ran into a little problem when testing on the isolator. When running with multiple ranks, it's possible some of the volumes won't have any local elements (which was expected). Currently the partitioning filters out empty groups, so the resulting mesh in this case has 0 groups. The problem is that mesh attributes like `dim` and `ambient_dim` look at the groups to figure out what to return, so if there are no groups, they fail. I tried making them return `None` in this case, but (not terribly surprisingly) this doesn't always work.

Do you think this could be fixed by changing the partitioning to retain the (empty) groups? Or could this have pitfalls of its own?

alexfikl:
> Do you think this could be fixed by changing the partitioning to retain the (empty) groups? Or could this have pitfalls of its own?

Do you mean keeping some `np.empty(dim, 0, nnodes)` arrays around in the mesh so that it knows the actual dimensions? That would make sense to me, since setting `nelements = 0` should be consistent.

inducer:
I think there's an indication there that `dim` and `ambient_dim` should be properly attributes of the mesh, but in the meantime I think keeping empty groups around is just fine.

meshmode/mesh/processing.py
```diff
@@ -409,98 +430,120 @@ def _create_boundary_groups(mesh, global_elem_to_part_elem, part_mesh_groups,
...
+            self_vertices,
+            self_mesh_groups,
+            facial_adjacency_groups=self_facial_adj_groups,
             is_conforming=mesh.is_conforming)
 
-    return part_mesh, queried_elems
+    return self_mesh
+
+
+def partition_mesh(mesh, part_id_to_elements):
```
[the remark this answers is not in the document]
majosm: Yeah, I don't know if we ever came to a final conclusion about that. What do you think about the `return_parts` argument I added, as a compromise? This way at least the caller never needs to call `partition_mesh` more than once, which should allow optimizations under the hood without necessarily forcing each rank to compute the whole thing.

inducer:
Grrr, again. Sorry. :)

majosm:
If the CI passes, this should be ready for review now.

Couple of notes:
* There are some FIXMEs relating to relabeling variables and moving code around; I didn't want to clutter this PR up with that stuff, so I'll probably do a follow-up later (maybe after #307 goes in?) to deal with that.
* I think this is squashable now that the mypy stuff is out.

meshmode/mesh/processing.py
```diff
@@ -80,83 +84,80 @@ def find_group_indices(groups, meshwide_elems):
...
+            (len(elements),), dtype=element_id_dtype)
 
     return global_elem_to_part_elem
 
 
-def _filter_mesh_groups(mesh, selected_elements, vertex_id_dtype):
+def _filter_mesh_groups(
+        mesh: Mesh,
+        selected_elements: np.ndarray,
+        vertex_id_dtype: Any) -> Tuple[List, np.ndarray]:
```
inducer, line 120:
```suggestion
        vertex_id_dtype: np.dtype[Any]) -> Tuple[List[ElementGroup], np.ndarray]:
```

meshmode/mesh/processing.py
```diff
@@ -80,83 +84,80 @@ def find_group_indices(groups, meshwide_elems):
 
 # {{{ partition_mesh
 
-def _compute_global_elem_to_part_elem(part_per_element, parts, element_id_dtype):
+def _compute_global_elem_to_part_elem(
+        nelements: int,
+        part_id_to_elements: Dict[PartID, np.ndarray],
+        part_id_to_part_index: Dict[PartID, int],
+        element_id_dtype: Any) -> np.ndarray:
```
inducer, line 91:
```suggestion
        element_id_dtype: np.dtype[Any]) -> np.ndarray:
```

meshmode/mesh/processing.py
```diff
@@ -80,83 +84,80 @@ def find_group_indices(groups, meshwide_elems):
...
         the filtered groups.
-    :returns: A tuple ``(new_groups, group_to_new_group, required_vertex_indices)``,
-        where *new_groups* is made up of groups from *groups* with elements not in
-        *selected_elements* removed (Note: empty groups are omitted),
-        *group_to_new_group* maps groups in *groups* to their corresponding location
-        in *new_groups*, and *required_vertex_indices* contains indices of all
-        vertices required for elements belonging to *new_groups*.
+    :arg vertex_id_dtype: The vertex index data type.
+    :returns: A tuple ``(new_groups, required_vertex_indices)``, where *new_groups*
+        is made up of groups from *groups* containing only elements from
```
inducer, line 129: References a no-longer present argument *groups*.

meshmode/mesh/processing.py
```diff
@@ -173,31 +174,36 @@ def _filter_mesh_groups(mesh, selected_elements, vertex_id_dtype):
...
 
-    return new_groups, group_to_new_group, required_vertex_indices
+    return new_groups, required_vertex_indices
 
 
-def _get_connected_partitions(
-        mesh, part_per_element, global_elem_to_part_elem):
+def _get_connected_parts(
+        mesh: Mesh,
+        part_id_to_part_index: Dict[PartID, int],
```
inducer, line 188:
```suggestion
        part_id_to_part_index: Mapping[PartID, int],
```
since you're not writing it. (Here and elsewhere.)

inducer (approved):
Thanks! Just a few minor things, then this is good to go. (Feel free to squash+merge when done.)

meshmode/mesh/processing.py
majosm: Hmm, looks like I can't do the `dtype[Any]` in 3.8? https://github.com/inducer/meshmode/runs/7123130267?check_suite_focus=true

majosm: @inducer Is it OK to just use `dtype` here? The CI seems to be fine with it.

inducer: I feel like reading somewhere that `dtype[Any]` was preferred, but now I can't find a source for that, and https://numpy.org/devdocs/reference/typing.html doesn't seem to back this up. Just `dtype` is fine I think then, and even if not, it's easy enough to fix.

For reference, another option would be to say
```
from __future__ import annotations
```
at the top of the file to hide the annotations from the interpreter.

PR #309, opened 2022-03-16
Add new helper for distributing data across MPI ranks

majosm:
We'll need something more general than the existing `MPIMeshDistributor` for multi-volume, because it currently does the partitioning on the inside and assumes a one-to-one relationship between ranks and partitions. This PR adds an `mpi_distribute` function which could be used to distribute mesh partitions without such assumptions. (It's more or less just an `MPI_Scatter` on a dict.)

Depends on #308.


inducer:
> assumes a one-to-one relationship between ranks and partitions.

I'm a little confused. As of #307, I believe this is no longer true.

majosm:
> I'm a little confused. As of #307, I believe this is no longer true.

I don't see any changes to `MPIMeshDistributor` in #307, only `MPIBoundaryCommSetupHelper`. As far as I can tell the mesh distributor still interprets partitions as 1 per rank.

inducer:
> > 
> 
> I don't see any changes to `MPIMeshDistributor` in #307, only `MPIBoundaryCommSetupHelper`. As far as I can tell the mesh distributor still interprets partitions as 1 per rank.

Argh, you're completely right. I got the two mixed up in my head. :facepalm: Sorry!

Issue #310, opened 2022-03-18
Tolerance too strict in _test_node_vertex_consistency_resampling


matthiasdiener:
We have seen lots of errors like these:

```
"/Users/mdiener/Work/efuse/meshmode/meshmode/mesh/__init__.py", line 1207, in _test_node_vertex_consistency
    assert _test_node_vertex_consistency_resampling(mesh, mgrp, tol)
  File "/Users/mdiener/Work/efuse/meshmode/meshmode/mesh/__init__.py", line 1195, in _test_node_vertex_consistency_resampling
    assert max_el_vertex_error < tol*size, max_el_vertex_error
AssertionError: 1.144170793882154e-15
```

This is the code:
https://github.com/inducer/meshmode/blob/48abe4bedf5e54e57fca392756423b6f45abff9a/meshmode/mesh/__init__.py#L1178

alexfikl:
What sort of meshes are these? I've only been able to trigger that if there was actually something weird with the mesh.

You can work around it from the outside by setting `node_vertex_consistency_tolerance` in the mesh constructor:
https://github.com/inducer/meshmode/blob/c1ff399316fc15938723433a3308f0710751e11b/meshmode/mesh/__init__.py#L919

matthiasdiener:
We have seen it most recently with this: https://github.com/illinois-ceesd/drivers_y2-isolator/tree/main/data/3D/eightX_injection

alexfikl:
Hm, I don't think the gmsh reader has a way to pass down the tolerance (most things in `mesh.generation` do), so this might need a bit of work.

alexfikl:
@matthiasdiener I just tried that mesh with just
```
from meshmode.mesh.io import read_gmsh
mesh = read_gmsh("isolator.msh")
```
and it didn't throw an error for me, weirdly. What version of `gmsh` are you using?

matthiasdiener:
These errors seem to be dependent on rank numbers as well. I tried https://github.com/illinois-ceesd/drivers_y2-isolator/test_mesh/lazy/oneX with 32 ranks, which does not trigger assertion, but 64 ranks trigger it.

alexfikl:
[...]

PR #311, opened 2022-03-19
Add a way to visualize the mesh node-vertex resampling error

alexfikl:
Used while looking at #310 yesterday and thought it may be useful. What do you think?

Added some types in a second commit, so probably best viewed individually.


inducer:
Eh, can't hurt. Thanks!

PR #314, opened 2022-03-25
`test_element_orientation_via_single_elements` (from @lukeolson)

inducer:
Curiously, there *are* some disagreements here.

cc @lukeolson

### DIFF
--- a/meshmode/mesh/generation.py
+++ b/meshmode/mesh/generation.py
@@ -570,7 +570,7 @@ def generate_sphere(r: float, order: int, *,
 # }}}
 
 
-# {{ generate_surface_of_revolution
+# {{{ generate_surface_of_revolution
 
 def generate_surface_of_revolution(
         get_radius: Callable[[np.ndarray, np.ndarray], np.ndarray],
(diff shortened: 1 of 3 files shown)


lukeolson:
Yep, I see that too.  It fails on this:
```python
import numpy as np
vert = np.array([[1, 0, 0],
                 [0, 1, 0],
                 [0, 0, 0],
                 [0, 0, 1]])
G = np.append(vert, np.ones((4, 1)), axis=1)

print(G)
print(np.linalg.det(G))
```
which gives
```
[[1. 0. 0. 1.]
 [0. 1. 0. 1.]
 [0. 0. 0. 1.]
 [0. 0. 1. 1.]]
-1.0
```

So it appears it should be negative.

inducer:
404f709c08c0df904340762faaf6d18239ce5d7b has the data directly copied from https://github.com/lukeolson/mesh_orientation/blob/460bb2b634e2abb6aa32c3d02e2c732969bf08bf/check.py, and it seems everybody's in agreement now.

@lukeolson, how do you feel about this?

lukeolson:
The test looks good.

inducer:
Thanks for taking a look!

PR #316, opened 2022-03-28
Use elementwise size in node vertex consistency check

### LINKED ISSUE #310: Tolerance too strict in _test_node_vertex_consistency_resampling

majosm:
Fixes #310

### DIFF
--- a/meshmode/mesh/__init__.py
+++ b/meshmode/mesh/__init__.py
@@ -1233,13 +1233,17 @@ def _test_node_vertex_consistency_resampling(mesh, mgrp, tol):
     if tol is None:
         tol = 1e3 * np.finfo(per_element_vertex_errors.dtype).eps
 
-    from meshmode.mesh.processing import find_bounding_box
+    grp_vertices = mesh.vertices[:, mgrp.vertex_indices]
 
-    bbox_min, bbox_max = find_bounding_box(mesh)
-    size = la.norm(bbox_max-bbox_min)
+    per_element_sizes = la.norm(
+        np.max(grp_vertices, axis=-1)
+        - np.min(grp_vertices, axis=-1),
+        axis=0)
 
-    max_el_vertex_error = np.max(per_element_vertex_errors)
-    assert max_el_vertex_error < tol*size, max_el_vertex_error
+    per_element_tols = tol * np.maximum(per_element_sizes, 1)
+
+    assert np.all(per_element_vertex_errors < per_element_tols), \
+        per_element_vertex_errors[per_element_vertex_errors >= per_element_tols][0]
 
     return True
 
--- a/meshmode/mesh/__init__.py
+++ b/meshmode/mesh/__init__.py
@@ -1218,9 +1218,11 @@ def _mesh_group_node_vertex_error(mesh, mgrp):
     return map_vertices - grp_vertices
 
 
-def _test_node_vertex_consistency_resampling(mesh, mgrp, tol):
+def _test_node_vertex_consistency_resampling(mesh, igrp, tol):
     if mesh.vertices is None:
         return True
+
+    mgrp = mesh.groups[igrp]
 
     if mgrp.nelements == 0:
         return True
@@ -1243,8 +1245,15 @@ def _test_node_vertex_consistency_resampling(mesh, mgrp, tol):
 
     per_element_tols = abs_tol + rel_tol * coord_scales
 
-    assert np.all(per_element_vertex_errors < per_element_tols), \
-        per_element_vertex_errors[per_element_vertex_errors >= per_element_tols][0]
+    elements_above_tol, = np.where(per_element_vertex_errors >= per_element_tols)
+    if len(elements_above_tol) > 0:
+        i_grp_elem = elements_above_tol[0]
+        ielem = i_grp_elem + mesh.base_element_nrs[igrp]
+        from meshmode import InconsistentVerticesError
+        raise AssertionError(
+            f"vertex consistency check failed for element {ielem}; "
+            f"{per_element_vertex_errors[i_grp_elem]} >= "
+            f"{per_element_tols[i_grp_elem]}")
 
     return True
 
@@ -1254,9 +1263,9 @@ def _test_node_vertex_consistency(mesh, tol):
     unit vertices.
     """
 
-    for mgrp in mesh.groups:
+    for igrp, mgrp in enumerate(mesh.groups):
         if isinstance(mgrp, _ModepyElementGroup):
-            assert _test_node_vertex_consistency_resampling(mesh, mgrp, tol)
+            assert _test_node_vertex_consistency_resampling(mesh, igrp, tol)
         else:
             from warnings import warn
             warn("not implemented: node-vertex consistency check for '%s'"
--- a/meshmode/mesh/__init__.py
+++ b/meshmode/mesh/__init__.py
@@ -1250,7 +1250,7 @@ def _test_node_vertex_consistency_resampling(mesh, igrp, tol):
         i_grp_elem = elements_above_tol[0]
         ielem = i_grp_elem + mesh.base_element_nrs[igrp]
         from meshmode import InconsistentVerticesError
-        raise AssertionError(
+        raise InconsistentVerticesError(
             f"vertex consistency check failed for element {ielem}; "
             f"{per_element_vertex_errors[i_grp_elem]} >= "
             f"{per_element_tols[i_grp_elem]}")
--- a/meshmode/mesh/__init__.py
+++ b/meshmode/mesh/__init__.py
@@ -1224,26 +1224,23 @@ def _test_node_vertex_consistency_resampling(mesh, igrp, tol):
 
     mgrp = mesh.groups[igrp]
 
-    if mgrp.nelements == 0:
+    if mgrp.dim == 0:
+        return True
+    elif mgrp.nelements == 0:
         return True
 
     per_vertex_errors = _mesh_group_node_vertex_error(mesh, mgrp)
-    per_element_vertex_errors = np.sqrt(
-        np.sum(np.sum((per_vertex_errors)**2, axis=0), axis=-1)
-    )
+    per_element_vertex_errors = np.max(
+        np.max(np.abs(per_vertex_errors), axis=-1), axis=0)
 
     if tol is None:
         tol = 1e3 * np.finfo(per_element_vertex_errors.dtype).eps
-
-    # Assume these are the same for now; can be made separate arguments if necessary
-    abs_tol = tol
-    rel_tol = tol
 
     grp_vertices = mesh.vertices[:, mgrp.vertex_indices]
 
     coord_scales = np.max(np.max(np.abs(grp_vertices), axis=-1), axis=0)
 
-    per_element_tols = abs_tol + rel_tol * coord_scales
+    per_element_tols = tol * coord_scales
 
     elements_above_tol, = np.where(per_element_vertex_errors >= per_element_tols)
     if len(elements_above_tol) > 0:
--- a/meshmode/mesh/__init__.py
+++ b/meshmode/mesh/__init__.py
@@ -1224,9 +1224,7 @@ def _test_node_vertex_consistency_resampling(mesh, igrp, tol):
 
     mgrp = mesh.groups[igrp]
 
-    if mgrp.dim == 0:
-        return True
-    elif mgrp.nelements == 0:
+    if mgrp.nelements == 0:
         return True
 
     per_vertex_errors = _mesh_group_node_vertex_error(mesh, mgrp)
@@ -1240,7 +1238,7 @@ def _test_node_vertex_consistency_resampling(mesh, igrp, tol):
 
     coord_scales = np.max(np.max(np.abs(grp_vertices), axis=-1), axis=0)
 
-    per_element_tols = tol * coord_scales
+    per_element_tols = tol + tol * coord_scales
 
     elements_above_tol, = np.where(per_element_vertex_errors >= per_element_tols)
     if len(elements_above_tol) > 0:
--- a/meshmode/mesh/__init__.py
+++ b/meshmode/mesh/__init__.py
@@ -1259,6 +1259,8 @@ def _test_node_vertex_consistency(mesh, tol):
     """Ensure that order of by-index vertices matches that of mapped
     unit vertices.
     """
+    if not __debug__:
+        return True
 
     for igrp, mgrp in enumerate(mesh.groups):
         if isinstance(mgrp, _ModepyElementGroup):
--- a/meshmode/mesh/__init__.py
+++ b/meshmode/mesh/__init__.py
@@ -1240,6 +1240,8 @@ def _test_node_vertex_consistency_resampling(mesh, mgrp, tol):
         - np.min(grp_vertices, axis=-1),
         axis=0)
 
+    # Only scale by the element size if the elements are sufficiently large,
+    # to prevent the tolerance from dropping below rounding error
     per_element_tols = tol * np.maximum(per_element_sizes, 1)
 
     assert np.all(per_element_vertex_errors < per_element_tols), \
--- a/meshmode/mesh/__init__.py
+++ b/meshmode/mesh/__init__.py
@@ -1230,13 +1230,12 @@ def _test_node_vertex_consistency_resampling(mesh, mgrp, tol):
         np.sum(np.sum((per_vertex_errors)**2, axis=0), axis=-1)
     )
 
-    abs_tol = 1e-12
-    rel_tol = 1e-7
-    if tol is not None:
-        # Rescale for compatibility with tolerances from old version that used
-        # mesh bounding box for scale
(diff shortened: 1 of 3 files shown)


matthiasdiener:
This PR fixes issue #310 for me. In particular, I tried the following configuration, which fails without this PR:

- `drivers_y2-isolator/test_mesh/lazy/oneX` running on 16 nodes (64 MPI ranks).

meshmode/mesh/__init__.py
```diff
@@ -1190,13 +1190,17 @@ def _test_node_vertex_consistency_resampling(mesh, mgrp, tol):
...
+        axis=0)
 
-    bbox_min, bbox_max = find_bounding_box(mesh)
-    size = la.norm(bbox_max-bbox_min)
+    per_element_tols = tol * np.maximum(per_element_sizes, 1)
 
-    max_el_vertex_error = np.max(per_element_vertex_errors)
-    assert max_el_vertex_error < tol*size, max_el_vertex_error
+    assert np.all(per_element_vertex_errors < per_element_tols), \
+        per_element_vertex_errors[per_element_vertex_errors >= per_element_tols][0]
```
inducer, line 1203: Did you intend an `np.where` around the query here?

meshmode/mesh/__init__.py
```diff
@@ -1190,13 +1190,17 @@ def _test_node_vertex_consistency_resampling(mesh, mgrp, tol):
...
+    grp_vertices = mesh.vertices[:, mgrp.vertex_indices]
+
+    per_element_sizes = la.norm(
+        np.max(grp_vertices, axis=-1)
+        - np.min(grp_vertices, axis=-1),
+        axis=0)
 
-    bbox_min, bbox_max = find_bounding_box(mesh)
-    size = la.norm(bbox_max-bbox_min)
+    per_element_tols = tol * np.maximum(per_element_sizes, 1)
```
inducer, line 1200: Why the max with 1? (Add a comment to explain?)

inducer (approved):
Thanks for working on this! LGTM other than those two wrinkles.

meshmode/mesh/__init__.py
majosm: Because of the `[0]`? No, I just didn't want it to print out the whole list, so instead I made it print the first entry that's larger than the tolerance. 🤷‍♂️

meshmode/mesh/__init__.py
majosm: Without the max, as the elements get smaller the tolerance goes below machine precision and then the check always fails. I figure ~1e-13 is a small enough tolerance. I'll add a comment explaining this.

meshmode/mesh/__init__.py
inducer: OIC, fair enough.

meshmode/mesh/__init__.py
inducer: > then the check always fails.

Is this something that you've observed? I'd be really surprised. Machine epsilon is a measure of relative (not absolute) error introduced by rounding, so any deviation from its interpretation as relative error is fishy to me.

majosm: > Is this something that you've observed? I'd be really surprised. Machine epsilon is a measure of relative (not absolute) error introduced by rounding, so any deviation from its interpretation as relative error is fishy to me.

Hmm. Well, I did get a test failure, but I can't say I looked very closely at what was going on numerically. I've never really been able to get away with such tight tolerances in the past, so I just figured that must be it.

Here is the failure, FWIW:

```python
============================================= FAILURES ==============================================
_________________________________________ test_lookup_tree __________________________________________

visualize = False

    def test_lookup_tree(visualize=False):
>       mesh = mgen.make_curve_mesh(mgen.cloverleaf, np.linspace(0, 1, 1000), order=3)

test_mesh.py:675: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
../meshmode/mesh/generation.py:323: in make_curve_mesh
    mesh = Mesh(
../meshmode/mesh/__init__.py:1003: in __init__
    assert _test_node_vertex_consistency(
../meshmode/mesh/__init__.py:1215: in _test_node_vertex_consistency
    assert _test_node_vertex_consistency_resampling(mesh, mgrp, tol)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

mesh = Mesh(element_id_dtype=dtype('int32'), _nodal_adjacency=None, is_conforming=True, groups=[SimplexElementGroup(order=3, ...52,  0.99432003],
       [-0.3       , -0.29365717, -0.2872078 , ..., -0.31838681,
        -0.31236503, -0.30623603]]))
mgrp = SimplexElementGroup(order=3, vertex_indices=array([[  0,   1],
       [  1,   2],
       [  2,   3],
       ...,
     ...-1.       , -0.4472136,  0.4472136,  1.       ]]), element_nr_base=0, node_nr_base=0, _factory_constructed=True, dim=1)
tol = 2.220446049250313e-13

    def _test_node_vertex_consistency_resampling(mesh, mgrp, tol):
        if mesh.vertices is None:
            return True
    
        if mgrp.nelements == 0:
            return True
    
        per_vertex_errors = _mesh_group_node_vertex_error(mesh, mgrp)
        per_element_vertex_errors = np.sqrt(
            np.sum(np.sum((per_vertex_errors)**2, axis=0), axis=-1)
        )
    
        if tol is None:
            tol = 1e3 * np.finfo(per_element_vertex_errors.dtype).eps
    
        grp_vertices = mesh.vertices[:, mgrp.vertex_indices]
    
        per_element_sizes = la.norm(
            np.max(grp_vertices, axis=-1)
            - np.min(grp_vertices, axis=-1),
            axis=0)
    
        per_element_tols = tol * per_element_sizes
    
>       assert np.all(per_element_vertex_errors < per_element_tols), \
            per_element_vertex_errors[per_element_vertex_errors >= per_element_tols][0]
E       AssertionError: 1.5700924586837752e-16

../meshmode/mesh/__init__.py:1202: AssertionError
```

(It passes if I change the `1000` to `500`.)

alexfikl: Would this be better fixed by something like that [np.isclose](https://numpy.org/doc/stable/reference/generated/numpy.isclose.html) does? i.e. Setting
```
per_element_tols = atol + rtol * per_element_sizes
```
where `atol` would just be `np.finfo(per_element_vertex_errors.dtype).eps` if not given.

majosm: Looking a little more closely at one of the entries that fails the assertion:

```python
map_vertices[:, i, :]=array([[-0.49340535, -0.49366492],
       [ 0.49756031,  0.49690906]])
grp_vertices[:, i, :]=array([[-0.49340535, -0.49366492],
       [ 0.49756031,  0.49690906]])
per_vertex_errors[:, i, :]=array([[-1.11022302e-16,  0.00000000e+00],
       [ 1.11022302e-16,  0.00000000e+00]])
per_element_vertex_errors[i]=1.5700924586837752e-16
per_element_sizes[i]=0.0007010691296653309
per_element_tols[i]=1.5566861792167394e-16
```

we can see that the values being subtracted here (`map_vertices` and `grp_vertices`) are both O(1), but the element size is much smaller than that. So I'm not sure if it makes sense to scale the tolerance by the element size for this check. How about scaling by the max absolute coordinate value in `grp_vertices` instead?

inducer: > How about scaling by the max absolute coordinate value in `grp_vertices` instead?

That makes sense to me. That's the correct magnitude to use as far as determining rounding error is concerned.

alexfikl: What about the vertices of the element that's at the origin? Wouldn't that have the same problem? (granted, a bit of an edge case, so feel free to ignore!)

inducer: Hmm, fair. The simplest might be to use `allclose` with 100*eps to do the comparison.

majosm: > What about the vertices of the element that's at the origin? Wouldn't that have the same problem? (granted, a bit of an edge case, so feel free to ignore!)

I _think_ this case should be OK. The max abs. value of the coordinates would be the same order as the vertex values being subtracted, so it would still effectively be computing a relative error. From what @inducer suggested above about machine epsilon, this should still work.

inducer: Upon thinking about this, I think I agree with @majosm: I'm not sure I see what would go wrong if there's an element at the origin. If we use the max over all vertex coordinates, then that should provide a natural magnitude scale with respect to which we should be able to expect nodes and vertices to match to C * eps * scale.

alexfikl: > The max abs. value of the coordinates would be the same order as the vertex values being subtracted

I was mostly worried about these being the same order (i.e. very small) so that `tol * size` will still be smaller than `eps` and make the check fail.

> If we use the max over all vertex coordinates

This would definitely fix my worry!

majosm:
@inducer I reimplemented the check and added some tests for the edge cases that we previously found. These only ensure that the check succeeds though... I'm not sure how to set up a test to find out if the check is actually doing anything (i.e. I would need a case that's expected to fail).

meshmode/mesh/__init__.py
```diff
@@ -1230,16 +1230,22 @@ def _test_node_vertex_consistency_resampling(mesh, mgrp, tol):
         np.sum(np.sum((per_vertex_errors)**2, axis=0), axis=-1)
     )
 
-    if tol is None:
-        tol = 1e3 * np.finfo(per_element_vertex_errors.dtype).eps
+    abs_tol = 1e-12
+    rel_tol = 1e-7
```
inducer, line 1234: It's not safe to assume that mesh data is always in double precision.

meshmode/mesh/__init__.py
```diff
@@ -1230,16 +1230,22 @@ def _test_node_vertex_consistency_resampling(mesh, mgrp, tol):
...
+    if tol is not None:
+        # Rescale for compatibility with tolerances from old version that used
+        # mesh bounding box for scale
+        old_default_tol = 1e3*np.finfo(per_element_vertex_errors.dtype).eps
+        rel_tol *= tol/old_default_tol
```
inducer, line 1239: Should `tol` be deprecated?

inducer:
> (i.e. I would need a case that's expected to fail).

An easy way would be to rotate the vertex indices array.

meshmode/mesh/__init__.py
majosm: Turns out I didn't really need that conversion stuff, so I dropped it. It's now using `tol` for both the absolute and relative tolerances. We can make them separate arguments if it becomes an issue, but for now it seems to work fine.

majosm:
> > (i.e. I would need a case that's expected to fail).
> 
> An easy way would be to rotate the vertex indices array.

Added.

test/test_mesh.py
```diff
@@ -1130,6 +1131,71 @@ def test_mesh_element_group_constructor(shape_cls, group_cls):
...
+# {{{ test_node_vertex_consistency_check
+
+def test_node_vertex_consistency_check(actx_factory):
+    actx = actx_factory()
+
+    # Mesh bounds invariance
+    h = 1e-5
+    for p in range(5):
+        nelems = 2**p
+        size = h*nelems
```
inducer, line 1143: Maybe a larger range? This goes from 1e-5 to 32*1e-5. I was thinking 1e-13 to 1e13.

test/test_mesh.py
```diff
@@ -1130,6 +1131,71 @@ def test_mesh_element_group_constructor(shape_cls, group_cls):
...
+        mesh = mgen.generate_regular_rect_mesh(  # noqa: F841
+            a=(-size/2,)*3, b=(size/2,)*3, nelements_per_axis=(nelems,)*3)
```
inducer, line 1145:
```py
mesh2 = perturb_vertex_coordinates(mesh, well_chosen_tolerance)
with pytest.raises(Stuff):
    perturb_vertex_coordinates(mesh, 2*well_chosen_tolerance)
```

test/test_mesh.py
```diff
@@ -1130,6 +1131,71 @@ def test_mesh_element_group_constructor(shape_cls, group_cls):
...
+    # Translation invariance
+    h = 1e-5
+    nelems = 8
+    size = h*nelems
+    for p in range(5):
+        mesh = mgen.generate_regular_rect_mesh(  # noqa: F841
+            a=(10**p-size/2,)*3, b=(10**p+size/2,)*3,
+            nelements_per_axis=(nelems,)*3)
+
+    # Zero-size elements
```
inducer, line 1156:
```suggestion
    # Zero-D elements
```

test/test_mesh.py
```diff
@@ -1130,6 +1131,71 @@ def test_mesh_element_group_constructor(shape_cls, group_cls):
...
+    from meshmode.discretization import Discretization
+    group_factory = default_simplex_group_factory(1, 1)
+    vol_discr = Discretization(actx, vol_mesh, group_factory)
+    from meshmode.discretization.connection import (
+        FACE_RESTR_ALL,
+        make_face_restriction)
+    faces_mesh = make_face_restriction(  # noqa: F841
+        actx, vol_discr, group_factory, FACE_RESTR_ALL, per_face_groups=False)
+
+    # Zero-size elements at the origin
```
inducer, line 1172:
```suggestion
    # Zero-size D at the origin
```

test/test_mesh.py
```diff
@@ -1130,6 +1131,71 @@ def test_mesh_element_group_constructor(shape_cls, group_cls):
...
+    vol_mesh = mgen.generate_regular_rect_mesh(
+        a=(-size/2,), b=(size/2,),
+        nelements_per_axis=(nelems,))
+    group_factory = default_simplex_group_factory(1, 1)
+    vol_discr = Discretization(actx, vol_mesh, group_factory)
+    faces_mesh = make_face_restriction(  # noqa: F841
+        actx, vol_discr, group_factory, FACE_RESTR_ALL, per_face_groups=False)
+
+    # Expected failing case: element vertex indices rotated
+    with pytest.raises(AssertionError):
```
inducer, line 1185: Maybe make a specific error for this case? (but only test if `__debug__`)

test/test_mesh.py
```diff
@@ -1130,6 +1131,71 @@ def test_mesh_element_group_constructor(shape_cls, group_cls):
...
+        a=(-size/2,), b=(size/2,),
+        nelements_per_axis=(nelems,))
+    from meshmode.discretization import Discretization
+    group_factory = default_simplex_group_factory(1, 1)
+    vol_discr = Discretization(actx, vol_mesh, group_factory)
+    from meshmode.discretization.connection import (
+        FACE_RESTR_ALL,
+        make_face_restriction)
+    faces_mesh = make_face_restriction(  # noqa: F841
+        actx, vol_discr, group_factory, FACE_RESTR_ALL, per_face_groups=False)
```
inducer, line 1170: Add a test for high-aspect ratio?

test/test_mesh.py
majosm: Fixed.

test/test_mesh.py
majosm: Added.

test/test_mesh.py
majosm: Added.

test/test_mesh.py
majosm: Added.

majosm:
Ready for a look again @inducer.

I ran into some difficulty trying to tightly bound the pass/fail in the tests (the `well_chosen_tolerance` thing you suggested) due to the presence of `abs_tol`. But `abs_tol` needs to be nonzero for the zero-D check to work. Question: do we even need to run this check if `dim == 0`? Or can it be skipped (like I'm doing in the current version)?

inducer:
> Question: do we even need to run this check if `dim == 0`? Or can it be skipped (like I'm doing in the current version)?

Yes, we do need to run it. There's still redundant information that could go out of sync.

A few more points:

- How about `machine_epsilon*max(max coordinate, max_element_extent)` as the model of round-off error?
- I'm not super picky about the tolerances being as tight as possible.

Issue #317, opened 2022-04-03
Pickling DOFArrays loses metadata


kaushikcfd:
Consider the snippet below:
```python
import pyopencl as cl
from meshmode.dof_array import array_context_for_pickling, DOFArray
from meshmode.array_context import PytatoPyOpenCLArrayContext
from arraycontext import freeze, thaw
from pytools.tag import Tag
from pickle import loads, dumps

ctx = cl.create_some_context()
cq = cl.CommandQueue(ctx)
actx = PytatoPyOpenCLArrayContext(queue=cq)

class FooTag(Tag):
    pass


state = DOFArray(actx, (actx.zeros((1, ), "float64"),))
state = thaw(freeze(actx.tag(FooTag(), state), actx), actx)

with array_context_for_pickling(actx):
    pkl = dumps((state, ))

with array_context_for_pickling(actx):
    loaded_state, = loads(pkl)

print(state)
print(loaded_state)
```

Prints:
```python
DOFArray((DataWrapper(axes=(Axis(tags=frozenset())), tags={FooTag()}, data=<arraycontext.impl.pyopencl.taggable_cl_array.TaggableCLArray object at 0x7f5db4423b20>, shape=(1), name=None),))
DOFArray((DataWrapper(axes=(Axis(tags=frozenset())), tags={}, data=<pyopencl.array.Array object at 0x7f5db4375cd0>, shape=(1), name=None),))
```

The `loaded_state` is no longer tagged with `FooTag`.

/cc @matthiasdiener

inducer:
#318 is in, which should address this. Closing.

PR #318, opened 2022-04-03
DOFArray: Pickle/unpickle tags

matthiasdiener:
Potential fix for #317

- [x] Add a test
- [x] Test backwards compatibility

### DIFF
--- a/meshmode/dof_array.py
+++ b/meshmode/dof_array.py
@@ -293,8 +293,13 @@ class DOFArray:
 
         self._array_context = actx
 
-        data = state[0]
-        tags = state[1]
+        if isinstance(state, dict):
+            data = state["data"]
+            tags = state["tags"]
+        else:
+            # For backwards compatibility
+            data = state
+            tags = [frozenset() for _ in range(len(state))]
 
         self._data = tuple(
             [actx.tag(tags[idx], actx.from_numpy(ary_i))
--- a/meshmode/dof_array.py
+++ b/meshmode/dof_array.py
@@ -272,7 +272,8 @@ class DOFArray:
             raise RuntimeError("DOFArray instances can only be pickled while "
                     "array_context_for_pickling is active.")
 
-        ary = self
+        # Make sure metadata inference has been done
+        ary = _thaw(freeze(self, self.array_context), self.array_context)
 
         if self.array_context is not actx:
             ary = _thaw(actx, _freeze(self))
--- a/meshmode/dof_array.py
+++ b/meshmode/dof_array.py
@@ -277,7 +277,9 @@ class DOFArray:
         if self.array_context is not actx:
             ary = _thaw(actx, _freeze(self))
 
-        return [actx.to_numpy(ary_i) for ary_i in ary._data]
+        return (
+            [actx.to_numpy(ary_i) for ary_i in ary._data],
+            [ary_i.tags for ary_i in ary._data])
 
     def __setstate__(self, state):
         try:
@@ -290,7 +292,13 @@ class DOFArray:
                     "array_context_for_pickling is active.")
 
         self._array_context = actx
-        self._data = tuple([actx.from_numpy(ary_i) for ary_i in state])
+
+        data = state[0]
+        tags = state[1]
+
+        self._data = tuple(
+            [actx.tag(tags[idx], actx.from_numpy(ary_i))
+             for idx, ary_i in enumerate(data)])
 
     # }}}
 
--- a/meshmode/dof_array.py
+++ b/meshmode/dof_array.py
@@ -309,6 +309,10 @@ class DOFArray:
             axes_tags = state["axes_tags"]
         else:
             # For backwards compatibility
+            from warnings import warn
+            warn("A DOFArray is being unpickled without (tag) metadata. "
+            "Program transformation may fail as a result.")
+
             data = state
             tags = [frozenset() for _ in range(len(data))]
             axes_tags = [[frozenset() for _ in range(leaf_ary.ndim)]
--- a/meshmode/dof_array.py
+++ b/meshmode/dof_array.py
@@ -282,7 +282,7 @@ class DOFArray:
 
         if hasattr(ary._data[0], "tags"):
             d["tags"] = [ary_i.tags for ary_i in ary._data]
-        d["axes"] = [ary_i.axes for ary_i in ary._data]
+        d["axes_tags"] = [ax.tags for ary_i in ary._data for ax in ary_i.axes]
 
         return d
 
@@ -301,14 +301,24 @@ class DOFArray:
         if isinstance(state, dict):
             data = state["data"]
             tags = state["tags"]
+            axes_tags = state["axes_tags"]
         else:
             # For backwards compatibility
             data = state
             tags = [frozenset() for _ in range(len(state))]
+            axes_tags = [frozenset() for _ in range(len(state))]
 
-        self._data = tuple(
-            [actx.tag(tags[idx], actx.from_numpy(ary_i))
-             for idx, ary_i in enumerate(data)])
+        self._data = []
+
+        for idx, ary in enumerate(data):
+            d = actx.tag(tags[idx], actx.from_numpy(ary))
+
+            for ida, ax in enumerate(axes_tags[idx]):
+                d = actx.tag_axis(ida, ax, d)
+
+            self._data.append(d)
+
+        self._data = tuple(self._data)
 
     # }}}
 
--- a/meshmode/dof_array.py
+++ b/meshmode/dof_array.py
@@ -280,12 +280,12 @@ class DOFArray:
         d = {}
         d["data"] = [actx.to_numpy(ary_i) for ary_i in ary._data]
 
-        if hasattr(ary._data[0], "tags"):
+        if len(ary) > 0 and hasattr(ary._data[0], "tags"):
             d["tags"] = [ary_i.tags for ary_i in ary._data]
         else:
             d["tags"] = [frozenset() for _ in range(len(ary._data))]
 
-        if hasattr(ary._data[0], "axes"):
+        if len(ary) > 0 and hasattr(ary._data[0], "axes"):
             d["axes_tags"] = [[ax.tags for ax in ary_i.axes] for ary_i in ary._data]
         else:
             d["axes_tags"] = [[frozenset() for _ in range(leaf_ary.ndim)]
--- a/meshmode/dof_array.py
+++ b/meshmode/dof_array.py
@@ -284,7 +284,11 @@ class DOFArray:
             d["tags"] = [ary_i.tags for ary_i in ary._data]
         else:
             d["tags"] = [frozenset() for _ in range(len(ary._data))]
-        d["axes_tags"] = [ax.tags for ary_i in ary._data for ax in ary_i.axes]
+
+        if hasattr(ary._data[0], "axes"):
+            d["axes_tags"] = [[ax.tags for ax in ary_i.axes] for ary_i in ary._data]
+        else:
+            d["axes_tags"] = [[] for _ in range(len(ary._data))]
 
         return d
 
--- a/meshmode/dof_array.py
+++ b/meshmode/dof_array.py
@@ -280,10 +280,7 @@ class DOFArray:
         d = {}
         d["data"] = [actx.to_numpy(ary_i) for ary_i in ary._data]
 
-        if len(ary) > 0 and hasattr(ary._data[0], "tags"):
-            d["tags"] = [ary_i.tags for ary_i in ary._data]
-        else:
-            d["tags"] = [frozenset() for _ in range(len(ary._data))]
+        d["tags"] = [getattr(ary_i, "tags", frozenset()) for ary_i in ary]
 
         if len(ary) > 0 and hasattr(ary._data[0], "axes"):
             d["axes_tags"] = [[ax.tags for ax in ary_i.axes] for ary_i in ary._data]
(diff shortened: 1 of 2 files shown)


meshmode/dof_array.py
```diff
@@ -290,7 +292,13 @@ def __setstate__(self, state):
                     "array_context_for_pickling is active.")
 
         self._array_context = actx
-        self._data = tuple([actx.from_numpy(ary_i) for ary_i in state])
+
+        data = state[0]
+        tags = state[1]
```
inducer, line 297:
```suggestion
        data, tags = state
```

meshmode/dof_array.py
```diff
@@ -290,7 +292,13 @@ def __setstate__(self, state):
                     "array_context_for_pickling is active.")
 
         self._array_context = actx
-        self._data = tuple([actx.from_numpy(ary_i) for ary_i in state])
+
+        data = state[0]
```
inducer, line 296: Unless we make this backward-compatible, this will needlessly break existing restart files. (But do include a warning.)

inducer (commented):
Thanks! This makes sense. Some comments below.

inducer:
> A test might be necessary

Agree

meshmode/dof_array.py
```diff
@@ -277,7 +277,9 @@ def __getstate__(self):
         if self.array_context is not actx:
             ary = _thaw(actx, _freeze(self))
 
-        return [actx.to_numpy(ary_i) for ary_i in ary._data]
+        return (
+            [actx.to_numpy(ary_i) for ary_i in ary._data],
+            [ary_i.tags for ary_i in ary._data])
```
kaushikcfd, line 282: 1. We should probably store this as a dict?
2. Also needs `.axes` metadata.
3. What about some array contexts that operate on arrays that do not have `.tags` attribute?
4. Should we care about backward compatibility with the existing data that was pickled before this PR?

kaushikcfd (changes requested):
Thanks, left some questions to discuss. This also needs some tests.

meshmode/dof_array.py
matthiasdiener: Updated by [cfa2262](https://github.com/inducer/meshmode/pull/318/commits/cfa226289c43c89b278e03a988aaad8f8599a2ad)

meshmode/dof_array.py
matthiasdiener: I *think* [cfa2262](https://github.com/inducer/meshmode/pull/318/commits/cfa226289c43c89b278e03a988aaad8f8599a2ad), but haven't tested yet.

meshmode/dof_array.py
matthiasdiener: > 1. We should probably store this as a dict?
Done in https://github.com/inducer/meshmode/pull/318/commits/2283819975aa5b0fdbc38e1abaf987aafc6e84e8
> 2. Also needs `.axes` metadata.
Done in https://github.com/inducer/meshmode/pull/318/commits/de6e398d5d8b4605a3a3dc5b8d870be3e514d418
> 3. What about some array contexts that operate on arrays that do not have `.tags` attribute?
https://github.com/inducer/meshmode/pull/318/commits/17e63d660cb4e0ccb23e9ecb4b3242aa108efcbb should hopefully fix that
> 4. Should we care about backward compatibility with the existing data that was pickled before this PR?

See [cfa2262](https://github.com/inducer/meshmode/pull/318/commits/cfa226289c43c89b278e03a988aaad8f8599a2ad)

matthiasdiener:
This is ready for a second look. There are some strange CI errors that are probably unrelated to this PR.

inducer:
> There are some strange CI errors that are probably unrelated to this PR.

Those were my fault. I screwed something up in ci-support.

meshmode/dof_array.py
```diff
@@ -277,7 +277,16 @@ def __getstate__(self):
...
 
-        return [actx.to_numpy(ary_i) for ary_i in ary._data]
+        d = {}
+        d["data"] = [actx.to_numpy(ary_i) for ary_i in ary._data]
+
+        if hasattr(ary._data[0], "tags"):
+            d["tags"] = [ary_i.tags for ary_i in ary._data]
+        else:
+            d["tags"] = [frozenset() for _ in range(len(ary._data))]
+        d["axes_tags"] = [ax.tags for ary_i in ary._data for ax in ary_i.axes]
```
kaushikcfd, line 287: 1. Another `hasattr(..., "axes")` here?
2. A more natural way to store it is a list of list? The inner list stores the tags corresponding to each axis of the array.

kaushikcfd (approved):
Thanks! Suggested some minor changes, otherwise LGTM.

meshmode/dof_array.py
matthiasdiener: Is 
89d7c19 what you had in mind?

meshmode/dof_array.py
```diff
@@ -277,7 +277,20 @@ def __getstate__(self):
...
+
+        if hasattr(ary._data[0], "tags"):
+            d["tags"] = [ary_i.tags for ary_i in ary._data]
+        else:
+            d["tags"] = [frozenset() for _ in range(len(ary._data))]
+
+        if hasattr(ary._data[0], "axes"):
+            d["axes_tags"] = [[ax.tags for ax in ary_i.axes] for ary_i in ary._data]
+        else:
+            d["axes_tags"] = [[] for _ in range(len(ary._data))]
```
kaushikcfd, line 291: This should be: `d["axes_tags"] = [[frozenset() for _ in range(leaf_ary.ndim)] for leaf_ary in range(len(ary._data))]`

meshmode/dof_array.py
```diff
@@ -290,7 +303,28 @@ def __setstate__(self, state):
...
+
+        if isinstance(state, dict):
+            data = state["data"]
+            tags = state["tags"]
+            axes_tags = state["axes_tags"]
+        else:
+            # For backwards compatibility
+            data = state
+            tags = [frozenset() for _ in range(len(state))]
+            axes_tags = [frozenset() for _ in range(len(state))]
```
kaushikcfd, line 315: This default is incorrect, according to the implementation in `__getstate__`.

meshmode/dof_array.py
```diff
@@ -290,7 +303,28 @@ def __setstate__(self, state):
...
+
+        self._data = []
+
```
kaushikcfd, line 318: How about an assert here that `assert len(data) == len(tags) == len(axes_tags)`?

meshmode/dof_array.py
```diff
@@ -290,7 +303,28 @@ def __setstate__(self, state):
...
+        for idx, ary in enumerate(data):
+            d = actx.tag(tags[idx], actx.from_numpy(ary))
+
```
kaushikcfd, line 321: How about an assert here that `assert len(axes_tags[idx]) == ary.ndim`?

meshmode/dof_array.py
```diff
@@ -290,7 +303,28 @@ def __setstate__(self, state):
...
+            for ida, ax in enumerate(axes_tags[idx]):
+                d = actx.tag_axis(ida, ax, d)
```
kaushikcfd, line 323: How about an assert before this line that `assert isinstance(axes_tags[idx], frozenset)`?

meshmode/dof_array.py
```diff
@@ -277,7 +277,20 @@ def __getstate__(self):
...
             ary = _thaw(actx, _freeze(self))
 
-        return [actx.to_numpy(ary_i) for ary_i in ary._data]
+        d = {}
+        d["data"] = [actx.to_numpy(ary_i) for ary_i in ary._data]
+
+        if hasattr(ary._data[0], "tags"):
+            d["tags"] = [ary_i.tags for ary_i in ary._data]
+        else:
+            d["tags"] = [frozenset() for _ in range(len(ary._data))]
```
kaushikcfd, line 286: Is it reasonable to expect that `ary._data` has at least 1 array? (I'm not sure, someone else can comment on this?)

kaushikcfd (approved):
Thanks. Found some minor bugs, and suggested some asserted that would (I think) catch such errors sooner.

meshmode/dof_array.py
matthiasdiener: Added in bd439cc159d69b0647375fb4c71f2d3734c227ee, thanks!

meshmode/dof_array.py
matthiasdiener: Added in bd439cc159d69b0647375fb4c71f2d3734c227ee, thanks!

meshmode/dof_array.py
matthiasdiener: `axes_tags[idx]` is a list; added this assert in bd439cc159d69b0647375fb4c71f2d3734c227ee

meshmode/dof_array.py
matthiasdiener: I changed it to `d["axes_tags"] = [[frozenset() for _ in range(leaf_ary.ndim)]
                                         for leaf_ary in ary._data]`

meshmode/dof_array.py
matthiasdiener: Thanks, I changed it to `axes_tags = [[frozenset() for _ in range(leaf_ary.ndim)]
                                         for leaf_ary in data]`

meshmode/dof_array.py
matthiasdiener: It seems an empty DOFArray is legal, so I added support for this case in 1ae79bd

meshmode/dof_array.py
```diff
@@ -277,7 +277,21 @@ def __getstate__(self):
...
             ary = _thaw(actx, _freeze(self))
 
-        return [actx.to_numpy(ary_i) for ary_i in ary._data]
+        d = {}
+        d["data"] = [actx.to_numpy(ary_i) for ary_i in ary._data]
+
+        if len(ary) > 0 and hasattr(ary._data[0], "tags"):
+            d["tags"] = [ary_i.tags for ary_i in ary._data]
+        else:
+            d["tags"] = [frozenset() for _ in range(len(ary._data))]
```
alexfikl, line 286:
```suggestion
	d["tags"] = [getattr(ary_i, "tags", frozenset()) for ary_i in ary]
```
This looks more readable to me and it should catch all the cases. What do you think?

alexfikl: There's probably a similar rewrite for the `axes`, but that's not going to look as nice because of the nesting :\

matthiasdiener: Looks good, thanks!

meshmode/dof_array.py
```diff
@@ -290,7 +301,33 @@ def __setstate__(self, state):
...
+        if isinstance(state, dict):
+            data = state["data"]
+            tags = state["tags"]
+            axes_tags = state["axes_tags"]
+        else:
+            # For backwards compatibility
+            data = state
+            tags = [frozenset() for _ in range(len(data))]
+            axes_tags = [[frozenset() for _ in range(leaf_ary.ndim)]
+                                         for leaf_ary in data]
```
inducer, line 314: Please add a warning here that a `DOFArray` is being unpickled without metadata, and that program transformation may fail as a result.

kaushikcfd:
The point at which metadata is being recorded may not have everything gone through metadata inference. Needs a freeze, something like this:

```diff
[line@line meshmode]$ git diff
diff --git a/meshmode/dof_array.py b/meshmode/dof_array.py
index eb3cd3f..4375228 100644
--- a/meshmode/dof_array.py
+++ b/meshmode/dof_array.py
@@ -272,7 +272,7 @@ class DOFArray:
             raise RuntimeError("DOFArray instances can only be pickled while "
                     "array_context_for_pickling is active.")
 
-        ary = self
+        ary = _thaw(freeze(self, self.array_context), self.array_context)
 
         if self.array_context is not actx:
             ary = _thaw(actx, _freeze(self))
@@ -286,7 +286,7 @@ class DOFArray:
             d["axes_tags"] = [[ax.tags for ax in ary_i.axes] for ary_i in ary._data]
         else:
             d["axes_tags"] = [[frozenset() for _ in range(leaf_ary.ndim)]
-                                         for leaf_ary in ary._data]
+                              for leaf_ary in ary._data]
 
         return d
```

(We have something similar in arraycontext-land: https://github.com/inducer/arraycontext/pull/152)

inducer:
> Needs a freeze, something like this:

Also add a comment justifying the thaw/freeze, maybe including a link back to Kaushik's comment.

meshmode/dof_array.py
matthiasdiener: Added in 
137e4dc

matthiasdiener:
> The point at which metadata is being recorded may not have everything gone through metadata inference. Needs a freeze, something like this:
> 

Added in [0438b18](https://github.com/inducer/meshmode/pull/318/commits/0438b18dc4ac7746f08d72ca1ca740a96398472c)

test/test_array.py
```diff
@@ -177,6 +177,43 @@ def test_dof_array_pickling(actx_factory):
     assert flat_norm(mat_of_dofs - mat2_of_dofs, np.inf) == 0
     assert flat_norm(dc_of_dofs - dc2_of_dofs, np.inf) == 0
 
+
+def test_dof_array_pickling_tags(actx_factory):
```
inducer, line 181: I don't think this test is being run against a lazy array context. I think that would be important to do.

matthiasdiener: I added testing with pytato to this file in [c82c6c2](https://github.com/inducer/meshmode/pull/318/commits/c82c6c251c36eb31e50839a7e1428814c1439946)

inducer:
LGTM, thanks!

PR #319, opened 2022-04-06
add run-pylint script

matthiasdiener:


inducer:
I don't particularly care for these (I have https://github.com/inducer/config-and-scripts/blob/main/bin/pylint-dev), but I guess I'll tolerate them.

PR #320, opened 2022-04-07
Optionally return volume-tag-to-element map from `read_gmsh` 

majosm:
Possibly supersedes #292.

### DIFF
--- a/meshmode/mesh/io.py
+++ b/meshmode/mesh/io.py
@@ -311,7 +311,7 @@ def read_gmsh(
 def generate_gmsh(source, dimensions=None, order=None, other_options=None,
         extension="geo", gmsh_executable="gmsh", force_ambient_dim=None,
         output_file_name="output.msh", mesh_construction_kwargs=None,
-        target_unit=None):
+        target_unit=None, return_tag_to_elements_map=False):
     """Run :command:`gmsh` on the input given by *source*, and return a
     :class:`meshmode.mesh.Mesh` based on the result.
 
@@ -346,9 +346,14 @@ def generate_gmsh(source, dimensions=None, order=None, other_options=None,
         parse_gmsh(recv, runner.output_file,
                 force_dimension=force_ambient_dim)
 
-    mesh = recv.get_mesh()
+    result = recv.get_mesh(return_tag_to_elements_map=return_tag_to_elements_map)
 
     if force_ambient_dim is None:
+        if return_tag_to_elements_map:
+            mesh = result[0]
+        else:
+            mesh = result
+
         AXIS_NAMES = "xyz"  # noqa
 
         dim = mesh.vertices.shape[0]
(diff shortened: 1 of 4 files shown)


majosm:
@inducer, @alexfikl Any ideas how I could appease pylint here?

```
************* Module meshmode.mesh.io
[926](https://github.com/inducer/meshmode/runs/5873887120?check_suite_focus=true#step:3:926)
meshmode/mesh/io.py:354:14: E1101: Instance of 'tuple' has no 'vertices' member (no-member)
[927](https://github.com/inducer/meshmode/runs/5873887120?check_suite_focus=true#step:3:927)
meshmode/mesh/io.py:356:16: E1101: Instance of 'tuple' has no 'vertices' member (no-member)
[928](https://github.com/inducer/meshmode/runs/5873887120?check_suite_focus=true#step:3:928)
************* Module test_mesh
[929](https://github.com/inducer/meshmode/runs/5873887120?check_suite_focus=true#step:3:929)
test/test_mesh.py:548:12: E1101: Instance of 'tuple' has no 'groups' member (no-member)
[930](https://github.com/inducer/meshmode/runs/5873887120?check_suite_focus=true#step:3:930)
test/test_mesh.py:549:48: E1101: Instance of 'tuple' has no 'vertices' member (no-member)
[931](https://github.com/inducer/meshmode/runs/5873887120?check_suite_focus=true#step:3:931)
test/test_mesh.py:712:26: E1101: Instance of 'tuple' has no 'groups' member (no-member)
[932](https://github.com/inducer/meshmode/runs/5873887120?check_suite_focus=true#step:3:932)
test/test_mesh.py:879:15: E1101: Instance of 'tuple' has no 'groups' member (no-member)
[933](https://github.com/inducer/meshmode/runs/5873887120?check_suite_focus=true#step:3:933)
test/test_mesh.py:886:42: E1101: Instance of 'tuple' has no 'vertices' member (no-member)
[934](https://github.com/inducer/meshmode/runs/5873887120?check_suite_focus=true#step:3:934)
test/test_mesh.py:893:30: E1101: Instance of 'tuple' has no 'vertices' member (no-member)
```

It doesn't seem to like the fact that the result type of `read_gmsh` depends on its arguments (it thinks it's always a `tuple`, when AFAICT it's not.)

alexfikl:
> Any ideas how I could appease pylint here?

As far as I know, pylint isn't smart enough to figure this out, so probably best to just add some ignores :\

majosm:
Looks like https://github.com/inducer/meshmode/pull/320/commits/5d9dcb98b880b92155998f4f81a71eca0a444e2c fixed it, somehow...

majosm:
Seems like this is the way to go; it's cleaner than anything I've been able to come up with otherwise. In some ways it's unsatisfying, because it sidesteps my sense that if we're tagging surfaces inside the mesh, we ought to be tagging volumes as well; but I haven't been able to find a nice way to integrate that into the current code structure. I guess this approach could still be consistent with that, as long as we think about a given `Mesh` as only ever representing a single volume.

inducer:
LGTM, thanks!

PR #325, opened 2022-05-02
Avoid bad indexing into 0-long axes

### LINKED ISSUE #321: Resampling can lead to expressions that index along a 0-long axis
Expressions like these: https://github.com/inducer/meshmode/blob/cfe8346044d1333ec9ab75e03734cb7addedd2b4/meshmode/discretization/connection/direct.py#L688-L692

could lead to expressions like --
```python
def main(np):
    ary = np.zeros((0, 10))
    idx1 = np.arange(10, dtype="int32").reshape(-1, 1)
    idx2 = np.zeros((10, 5), dtype="int32")

    np.where(True, np.ones((10, 5)), ary[idx1, idx2])


if __name__ == "__main__":
    import numpy as np
    main(np)
```

notice how the expression in `main()` function is only valid for lazy array packages. Another point to note is that `jax` (unlike pytato) makes indexing into 0-long axes illegal to begin with.

kaushikcfd:
Closes #321.

Draft because:
- [x] Pass CI

### DIFF
--- a/meshmode/discretization/connection/direct.py
+++ b/meshmode/discretization/connection/direct.py
@@ -685,20 +685,20 @@ class DirectDiscretizationConnection(DiscretizationConnection):
                     for fgpd in group_pick_info:
                         from_element_indices = actx.thaw(fgpd.from_element_indices)
 
-                        grp_ary_contrib = ary[fgpd.from_group_index][
-                                    from_element_indices.reshape((-1, 1)),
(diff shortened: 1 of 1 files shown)


kaushikcfd:
Ok, the CI here is passing. Dropping the "DROP BEFORE MERGE" commit.

image

inducer:
LGTM, thanks!

PR #326, opened 2022-05-08
Incorporate PyOP2 fix into Firedrake CI

inducer:


inducer:
Barked up the wrong tree. Need to update pymbolic instead.

PR #327, opened 2022-05-09
Support `with_array_context` for DOFArrays

kaushikcfd:
See https://github.com/inducer/arraycontext/pull/158#issuecomment-1120080728 for context.

- :warning: Needs to go in conjunction with https://github.com/inducer/arraycontext/pull/158
- [ ] Drop "Drop before merge" commit.

### DIFF
--- a/meshmode/dof_array.py
+++ b/meshmode/dof_array.py
@@ -40,7 +40,7 @@ from arraycontext import (
         ArrayContext, NotAnArrayContainerError,
         make_loopy_program, with_container_arithmetic,
         serialize_container, deserialize_container,
-        thaw as _thaw, freeze as _freeze,
+        thaw as _thaw, freeze as _freeze, with_array_context,
         rec_map_array_container, rec_multimap_array_container,
         mapped_over_array_containers, multimapped_over_array_containers)
 from arraycontext.container import ArrayOrContainerT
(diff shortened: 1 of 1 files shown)


meshmode/dof_array.py
```diff
@@ -378,25 +378,9 @@ def _raise_index_inconsistency(i, stream_i):
...
-def _thaw_dofarray(ary, actx):
-    if ary.array_context is not None:
-        raise ValueError("cannot thaw DOFArray that already has an array context")
-
-    return type(ary)(
-        actx,
-        tuple(actx.thaw(subary) for subary in ary._data))
+@with_array_context.register(DOFArray)
+def _with_actx_dofarray(ary, actx):
+    return type(ary)(actx, ary._data)
```
alexfikl, line 383: Add a check that the arrays in `_data` are in `actx.array_types`? Maybe under an `if __debug__`

kaushikcfd: Sure, recursively in each component of `_data`?

alexfikl: That shouldn't be necessary, since `DOFArray._data` should just contain the base array type (right?).

kaushikcfd: Done, thanks! (I thought DOFArrays could also contain numpy object arrays of base arrays).

inducer:
Went ahead and force-merged this based on earlier CI passing.

PR #328, opened 2022-05-14
docs: use svg images

alexfikl:


inducer:
Thanks! Just earlier today I was using https://github.com/sphinx-contrib/tikz, and it seemed to work pretty well. What do you think of that?

alexfikl:
Hm, haven't heard of it before! I personally don't mind either way, but latex is pretty slow and big just to generate two images.

inducer:
One downside to SVGs is that they prevent PDFs to be built, AFAIK. Let's try the sphinx+TikZ thing. It's easy enough to back out. Plus it fails somewhat gracefully:

https://github.com/illinois-ceesd/mirgecom/pull/667#issuecomment-1126184169

alexfikl:
> One downside to SVGs is that they prevent PDFs to be built, AFAIK. Let's try the sphinx+TikZ thing. It's easy enough to back out. 

@inducer Gave it a go in the latest version and it seems to work just fine. Also tried to generate a PDF and that worked too! Should be good to go if the Documentation CI passes.

alexfikl:
Hm, actually going by https://github.com/sphinx-contrib/tikz/issues/3, the CI is not going to fail either way. Any suggestions on how to make sure the images actually get generated/uploaded correctly?

inducer:
> Any suggestions on how to make sure the images actually get generated/uploaded correctly?

I don't have a good suggestion, short of hacking the sphinx extension. I agree that this is a risk, but I feel that the potential impact is small enough (a few broken images) that I'd be willing to take it for now. Someone's bound to notice if there's anyone that cares about the code.

doc/images/facial-adjacency-group.tikz
```diff
@@ -1,14 +1,11 @@
-\documentclass{standalone}
-\usepackage{tikz}
-
-\begin{document}
 \begin{tikzpicture}
```
inducer, line 1: I figured we'd linline those. What do you think?

alexfikl: Sure thing, went ahead and inlined them in the latest version.

.github/workflows/ci.yml
```diff
@@ -111,8 +111,10 @@ jobs:
             run: |
                 curl -L -O https://gitlab.tiker.net/inducer/ci-support/raw/main/ci-support.sh
                 . ci-support.sh
+
+                export EXTRA_INSTALL="sphinxcontrib-tikz"
                 build_py_project_in_conda_env
-                conda install graphviz
+                conda install graphviz texlive-core pdf2svg
```
alexfikl, line 117: The `sphinxcontrib-tikz` docs say that this also needs `texlive-pictures` (for `pgfplots` presumably), but that doesn't seem to be packaged in `conda-forge`, so not quite sure if this'll work as expected :worried:

inducer: The docs are built on the beers, where this is all a non-issue. Here, you can simply do `sudo apt install texlive-pictures`.

alexfikl: Installed that in the latest version too! So this should be good to go now!

inducer:
This is nice! Thanks for working on it.

PR #329, opened 2022-05-17
Remove tags before restoring pickled ones

matthiasdiener:
Needed for:
- [ ] https://github.com/inducer/pytato/pull/284

Needs:
- [ ] https://github.com/inducer/pytools/pull/134


meshmode/dof_array.py
```diff
@@ -326,7 +326,14 @@ def __setstate__(self, state):
         for idx, ary in enumerate(data):
             assert len(axes_tags[idx]) == ary.ndim
             assert isinstance(axes_tags[idx], list)
-            d = actx.tag(tags[idx], actx.from_numpy(ary))
+
+            node = actx.from_numpy(ary)
+            from pytools.tag import IgnoredForEqualityTag
+            node.tags = frozenset(
+                tag for tag in node.tags
+                if not isinstance(tag, IgnoredForEqualityTag))
```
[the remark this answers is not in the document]
matthiasdiener: 140b0ae9 attempts to explain this a bit more. Maybe it makes sense to change the filter to `if not (isinstance(tag, IgnoredForEqualityTag) and isinstance(tag, UniqueTag)))`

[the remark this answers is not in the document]
matthiasdiener: I think we have basically these options here:
- Filter out the original CreatedAt tag when pickling, a new CreatedAt tag will be created when unpickling (this would "lose" the original creation location)
- Filter out the "incorrect" CreatedAt tag when unpickling (this is what is currently done in this PR, and keeps the original creation location)
- Make `CreatedAt` an `UniqueButOverwritableTag` or so (this would result in the cleanest code here, but would also lose the original creation location)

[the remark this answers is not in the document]
inducer: I see. I understand the problem better now.

meshmode/dof_array.py
```diff
@@ -326,7 +326,23 @@ def __setstate__(self, state):
...
+            # {{{ Filter IgnoredForEqualityTags
+
+            # actx.from_numpy() may create tags (such as pytato's CreatedAt tag).
+            # Remove the tags that are purely informational, such that
+            # actx.tag() below will not cause NonUniqueTagErrors.
+
+            import sys
+            if "pytato" in sys.modules:
+                from pytato.equality import preprocess_tags_for_equality
+                node.tags = preprocess_tags_for_equality(node.tags)
```
inducer, line 341: Stateful modification of a pytato node is never OK.

inducer, line 341: Why not simply restore the set of tags from the pickled state?

matthiasdiener: That's what happens in the `actx.tag(tags[idx], node)` line below. The problem is that `actx.from_numpy(ary)` above creates a `CreatedAt` tag first, which we need to handle in some way.

inducer: `node = node.without_tags(node.tags).with_tags(old_tags)`?

matthiasdiener: What do you think of 4d43297?

meshmode/dof_array.py
```diff
@@ -326,7 +326,11 @@ def __setstate__(self, state):
         for idx, ary in enumerate(data):
             assert len(axes_tags[idx]) == ary.ndim
             assert isinstance(axes_tags[idx], list)
-            d = actx.tag(tags[idx], actx.from_numpy(ary))
+
+            # actx.from_numpy() may create tags. Remove these tags before adding
+            # the pickled ones such that we do not raise NonUniqueTagErrors.
+            d = actx.from_numpy(ary)
+            d = d.without_tags(d.tags)._with_new_tags(tags[idx])
```
inducer, line 333: If you use `_with_new_tags`, then you don't need to remove the old tags. (This just returns a new thing with the tags set.) But that's an internal API. Perhaps its use is justified here, not sure. I guess I'd be OK with that.

matthiasdiener: https://github.com/inducer/meshmode/pull/329/commits/5b57e45cf827a6c3de719723e9f95bc8cc074a8d

Imho, using `_with_new_tags` makes this much shorter and saves us a rather clumsy comment... ;-)

inducer:
LGTM, thanks!

PR #331, opened 2022-05-17
flake8 on experiments/refinement-playground.py

matthiasdiener:


alexfikl:
Worth moving this to `examples`? So that it stayed `flake8`-ed? (assuming it still runs)

inducer:
Not sure I'd like that in examples. But maybe include experiments in the flake8 check (on Github, Gitlab)? (Could just run black to get things started...)

matthiasdiener:
> Worth moving this to `examples`? So that it stayed `flake8`-ed? (assuming it still runs)

That sounds good to me. What do you think @inducer? Alternatively, we could also add expreiments to CI.

inducer:
Thanks!

PR #334, opened 2022-06-11
Port deprecated uses of freeze and thaw

alexfikl:


alexfikl:
Checked the output in 
https://github.com/inducer/meshmode/runs/6845138643?check_suite_focus=true
and everything seems to be clean now. There is a rather worrisome warning that I don't recall seeing before
```
/home/runner/work/meshmode/meshmode/.miniforge3/envs/testing/lib/python3.10/site-packages
/pytools/persistent_dict.py:567: CollisionWarning: loopy-typed-and-scheduled-cache-v1-2022.1-
islpy2022.1.2-cgen2020.1-a24370e8dfd417e24aa37cfb900b4425af3caab3-v1: key collision in cache 
at '/home/runner/.cache/pytools/pdict-v4-loopy-typed-and-scheduled-cache-v1-2022.1-
islpy2022.1.2-cgen2020.1-a24370e8dfd417e24aa37cfb900b4425af3caab3-v1-py3.10.4.final.0' -- 
these are sufficiently unlikely that they're often indicative of a broken hash key 
implementation (that is not considering some elements relevant for equality comparison)
```

Not sure where it comes from though :worried:

inducer:
Bumped that to an issue, as it's almost certainly unrelated. As is, they're an efficiency issue (as they will trigger redundant work in loopy), but they should not impact correctness, so I think it's OK to proceed here.

Thanks for taking care of this!

PR #337, opened 2022-06-12
Add name hints to `nodes()` and some index arrays in connections

inducer:


inducer:
This should be harmless as long as it doesn't break anything and thus uncontroversial, it just adds some names. Setting to merge.

PR #340, opened 2022-06-28
Enable type checking

majosm:
Splitting this out from #308 to keep the discussion separate.

### DIFF
--- a/meshmode/mesh/processing.py
+++ b/meshmode/mesh/processing.py
@@ -149,7 +149,8 @@ def _filter_mesh_groups(mesh, selected_elements, vertex_id_dtype):
     filtered_vertex_indices = [
             mesh.groups[i_old_grp].vertex_indices[
                     filtered_group_elements[i_new_grp], :]
-            for i_new_grp, i_old_grp in enumerate(new_group_to_old_group)]
+            for i_new_grp, i_old_grp in enumerate(new_group_to_old_group)
+            if mesh.groups[i_old_grp].vertex_indices is not None]
 
     if n_new_groups > 0:
         filtered_vertex_indices_flat = np.concatenate([indices.ravel() for indices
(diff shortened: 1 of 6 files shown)


meshmode/mesh/processing.py
```diff
@@ -1287,14 +1288,14 @@ def map_mesh(mesh, f):  # noqa
...
+        A: Optional[Union[np.generic, np.ndarray]] = None,    # noqa: N803
+        b: Optional[Union[np.generic, np.ndarray]] = None):
     """Apply the affine map :math:`f(x) = A x + b` to the geometry of *mesh*."""
 
-    if isinstance(A, Real):
+    if A is not None and not isinstance(A, np.ndarray):
         A = np.diag([A] * mesh.ambient_dim)             # noqa: N806
 
-    if isinstance(b, Real):
+    if b is not None and not isinstance(b, np.ndarray):
```
majosm, line 1298: @inducer I had a little trouble here: the type checking for the numpy functions doesn't like `Real`. I get this error with the original code:

```
meshmode/mesh/processing.py:1295: error: Argument 1 to "diag" has incompatible type "List[Real]"; expected "Union[_SupportsArray[dtype[Any]], Sequence[_SupportsArray[dtype[Any]]], Sequence[Sequence[_SupportsArray[dtype[Any]]]], Sequence[Sequence[Sequence[_SupportsArray[dtype[Any]]]]], Sequence[Sequence[Sequence[Sequence[_SupportsArray[dtype[Any]]]]]]]"  [arg-type]
```

The change fixes it, but I'm not sure if there's something more specific I can use instead of `np.generic` (or if this can be written in a different way to avoid the issue).

inducer: I think this is OK for now. Mypy is still at war with Python's numerical tower, cf. https://github.com/python/mypy/issues/3186 for example. I expect that this will get better with time.

inducer:
LGTM, thanks! (And I've marked mypy required.)

PR #342, opened 2022-07-01
Add new helper for distributing data across MPI ranks

majosm:
Replaces auto-closed #309.


majosm:
Seems this is no longer needed (see illinois-ceesd/mirgecom#889).

PR #351, opened 2022-10-11
Reorder box generator facial/boundary adjacency loops

### LINKED ISSUE #350: Performance of `generate_box_mesh`
`generate_box_mesh` appears to be taking an inordinate amount of time.  This seems to be due primarily to the part of the generator that determines facial adjacency to domain boundaries.  

For every boundary (including periodic boundaries), the code loops through all elements, and determines adjacency for each face in the element.   For large (10s of millions of elements) with several boundaries (i.e. like those used in CEESD), box grid generation is taking hours.

One thing we can do to help make this a little faster is to [reorder the loops such that every element is visited only once](https://github.com/inducer/meshmode/pull/351), and checked against every boundary (i.e. instead of visiting every element for every boundary).   The following table are the generation times for a 3D periodic mesh with "original" loop ordering, and "modified" loop ordering:

number of elements | original | modified
-------------------- | -------- | --------
48k                            | 6s           |   2s
96k                            | 12s         |   4s
192k                          | 25s         |   9s
384k                         | 48s         |  19s
768k                         | 100s       |  32s
1.5M                          | 205s      |  87s

MTCam:
Reorders box generator's facial/boundary adjacency loop s.t. it checks every boundary for every element instead of checks every element for every boundary.  This limits the number of times we need to visit every element to check adjacency.  

fixes #350

### DIFF
--- a/meshmode/mesh/generation.py
+++ b/meshmode/mesh/generation.py
@@ -1214,53 +1214,57 @@ def generate_box_mesh(axis_coords, order=1, *, coord_dtype=np.float64,
     facial_adjacency_groups = None
     face_vertex_indices_to_tags = {}
     boundary_tags = list(boundary_tag_to_face.keys())
+    nbnd_tags = len(boundary_tags)
 
-    if boundary_tags:
+    if nbnd_tags > 0:
         vert_index_to_tuple = {
                 vertex_indices[itup]: itup
                 for itup in np.ndindex(shape)}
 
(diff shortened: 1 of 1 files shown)


inducer:
LGTM, thanks!

majosm:
> > LGTM, thanks!
> 
> Outstanding! Thanks! @majosm can this be plopped into majosm/meshmode@production?

Yep, added.

PR #353, opened 2022-10-12
Add helper that creates a grid of curves / surfaces

alexfikl:
Mostly to recreate some configurations from Greengard / Martinsson papers.

### DIFF
--- a/meshmode/mesh/generation.py
+++ b/meshmode/mesh/generation.py
@@ -47,13 +47,14 @@ Curve parametrizations
 .. autofunction:: circle
 .. autofunction:: ellipse
 .. autofunction:: cloverleaf
-.. data :: starfish
 .. autofunction:: drop
 .. autofunction:: n_gon
 .. autofunction:: qbx_peanut
 .. autofunction:: apple
 .. autoclass:: WobblyCircle
 .. autoclass:: NArmedStarfish
+.. data:: starfish3
+.. data:: starfish5
 
 Surfaces
 --------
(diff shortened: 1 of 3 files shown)


alexfikl:
The `make_mesh_grid` thing should be able to make things like this
Screenshot_20221012_200430

alexfikl:
The dumbbell curve should make things like the classic (also from https://doi.org/10.1016/j.jcp.2004.10.033). Any idea how to add the ripples?

Screenshot_20221012_200629

And the fidget spinner thing is from https://doi.org/10.1137/120866683

Screenshot_20221012_200654

meshmode/mesh/generation.py
```diff
@@ -240,14 +255,16 @@ class NArmedStarfish(WobblyCircle):
...
+    def __init__(self, n_arms: int, amplitude: float, phase: float = 0.0) -> None:
         coeffs = np.zeros(n_arms)
         coeffs[-1] = amplitude
-        super().__init__(coeffs)
+        super().__init__(coeffs, phase=phase)
 
 
 starfish = NArmedStarfish(5, 0.25)
 
+fidget_spinner = NArmedStarfish(3, 1 / 2, phase=np.pi / 2)
```
alexfikl, line 266: Name? :cry:

inducer: Fidget starfish?

alexfikl: Called this a `starfish3` and the old one a `starfish5`. What do you think?

inducer:
> Any idea how to add the ripples?

Looks like they're just y perturbations modulated by something like $(1-x**2)$?

alexfikl:
> Looks like they're just y perturbations modulated by something like (1−x2)?

Yeah, that was it! Thanks! It looks this (with some careful choice of parameters)
<p align="center">

</p>

inducer:
Thx!

PR #354, opened 2022-10-13
Freeze constant arrays

kaushikcfd:


meshmode/discretization/connection/direct.py
```diff
@@ -232,6 +232,11 @@ class _FromGroupPickData(Generic[ArrayT]):
     from_element_indices: ArrayT
     is_surjective: bool
 
+    @memoize_method
+    def indexed_dof_pick_lists(self, actx):
```
kaushikcfd, line 236: Not completely sure if this is the right object to maintain the memoization dict. (Lmk if you think something's not ideal here)

alexfikl, line 236: Does this work? I would expect you need something like
https://github.com/inducer/meshmode/blob/02a1a34a76ef6f3b32196878198ad87fa3ec3ceb/meshmode/discretization/connection/direct.py#L358=L361
since `actx` isn't exactly hashable?

kaushikcfd: > actx isn't exactly hashable?

Aah yep, thanks. Fixed!

However, I think the code you linked has the wrong cache key. It shouldn't completely ignore `actx` since the frozen array types might differ. What do you think of [31d849e](https://github.com/inducer/meshmode/pull/354/commits/31d849e1691e9d3ce3fb37233b0529277963f740)?

kaushikcfd: > Does this work? 

Also, surprisingly this was working before i.e. `actx`s are hashable for some reason.

alexfikl: > However, I think the code you linked has the wrong cache key. It shouldn't completely ignore `actx` since the frozen array types might differ.

Yeah, you're right about the frozen array types, at the time I think it was all just `cl.Array`. 

> What do you think of [31d849e](https://github.com/inducer/meshmode/pull/354/commits/31d849e1691e9d3ce3fb37233b0529277963f740)?

Using the full `type(actx)` could be safer? In case someone subclasses and gives it the same name, e.g.
https://github.com/inducer/boxtree/blob/529e8d62f365350a80c9f16a0b12b9fb2c1a140e/boxtree/array_context.py#L41

inducer:
Rerunning to check how things go now that https://github.com/inducer/arraycontext/pull/203 is in.

meshmode/discretization/connection/direct.py
kaushikcfd: > Using the full type(actx) could be safer? 

Thanks, done!

meshmode/discretization/connection/direct.py
```diff
@@ -232,6 +232,12 @@ class _FromGroupPickData(Generic[ArrayT]):
     from_element_indices: ArrayT
     is_surjective: bool
 
+    @keyed_memoize_method(key=lambda actx: type(actx))
+    def indexed_dof_pick_lists(self, actx):
```
inducer, line 236: Could you explain why this is needed? I would have figured that `DataWrapper` deduplication would take care of this.

kaushikcfd: This is computing `pt_data[pt_data_0[i, j]]` i.e. a constant that should be computed AOT rather than in every `compiled_rhs`. See [all these terms](https://gist.github.com/kaushikcfd/1193428f2834537e32c31bc27a0bbbb0#file-nspecies_eq_7-c-L1633-L1642).

inducer:
As discussed, I think I would prefer to do this as part of a more general constant folding pass in pytato. (And in this particular case, I think the temporaries being produced are `DOFArray`-sized, which I find fairly big.)

PR #357, opened 2022-10-21
Add types to all of mesh.generation

alexfikl:
And a bold
```
# mypy: disallow-untyped-defs
```
at the top!

### DIFF
--- a/meshmode/mesh/generation.py
+++ b/meshmode/mesh/generation.py
@@ -1,3 +1,5 @@
+# mypy: disallow-untyped-defs
+
 __copyright__ = "Copyright (C) 2013 Andreas Kloeckner"
 
 __license__ = """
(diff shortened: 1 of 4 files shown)


meshmode/mesh/refinement/utils.py
```diff
@@ -33,6 +38,40 @@
 logger = logging.getLogger(__name__)
 
 
+# {{{ base class
+
+class Refiner(ABC):
```
alexfikl, line 43: This may be a bit debatable. Mostly added it to document the interface and use it for more generic typing.

meshmode/mesh/generation.py
```diff
@@ -1340,15 +1360,18 @@ def generate_box_mesh(axis_coords, order=1, *, coord_dtype=np.float64,
...
-                               periodic=None,
-                               order=1,
-                               boundary_tag_to_face=None,
-                               group_cls=None,
-                               mesh_type=None,
-                               n=None,
-                               ) -> Mesh:
+def generate_regular_rect_mesh(
+        a: Tuple[float, ...] = (0, 0),
+        b: Tuple[float, ...] = (1, 1), *,
```
inducer, line 1365:
```suggestion
        a: Sequence[float, ...] = (0, 0),
        b: Sequence[float, ...] = (1, 1), *,
```

meshmode/mesh/refinement/utils.py
inducer: Fine by me.

inducer (commented):
Thanks! LGTM, I'll apply the suggestion.

PR #359, opened 2022-11-01
Weaken mesh check in `make_same_mesh_connection`

### LINKED ISSUE #358: Mesh check in `make_same_mesh_connection` too strict?
@anderson2981 and @tulioricci have both run into errors in `make_same_mesh_connection` when restarting with different orders. Specifically, this check is failing:

```python
    if from_discr.mesh is not to_discr.mesh:
        raise ValueError("from_discr and to_discr must be based on "
                "the same mesh")
```
I suspect it's related to the mesh boundary tag relabeling that grudge does when constructing `DiscretizationCollection`s. Strangely, though, changing the check condition to

```python
   if from_discr.mesh.groups is not to_discr.mesh.groups:
       ...
```
doesn't appear to help, so maybe something else is going on?

This can be worked around for now by just commenting out the check, but I should take a closer look.

majosm:
Weakens the mesh check a bit to fix #358. Also adds an object identity check to the mesh `__eq__` so it doesn't need to compare all the attributes if the two mesh objects are actually the same one.

### DIFF
--- a/meshmode/discretization/connection/same_mesh.py
+++ b/meshmode/discretization/connection/same_mesh.py
@@ -33,7 +33,7 @@ def make_same_mesh_connection(actx, to_discr, from_discr):
             IdentityDiscretizationConnection,
             DirectDiscretizationConnection)
 
-    if from_discr.mesh is not to_discr.mesh:
+    if from_discr.mesh != to_discr.mesh:
         raise ValueError("from_discr and to_discr must be based on "
                 "the same mesh")
 
--- a/meshmode/mesh/__init__.py
+++ b/meshmode/mesh/__init__.py
@@ -1177,6 +1177,8 @@ class Mesh(Record):
 
     def __eq__(self, other):
         return (
+            self is other
+            or (
                 type(self) == type(other)
                 and np.array_equal(self.vertices, other.vertices)
                 and self.groups == other.groups
@@ -1184,7 +1186,7 @@ class Mesh(Record):
                 and self.element_id_dtype == other.element_id_dtype
                 and self._nodal_adjacency == other._nodal_adjacency
                 and self._facial_adjacency_groups == other._facial_adjacency_groups
-                and self.is_conforming == other.is_conforming)
+                and self.is_conforming == other.is_conforming))
 
     def __ne__(self, other):
         return not self.__eq__(other)


inducer:
I don't know that I love this, because pretty much by definition, we'll have redundant meshes flying around, which seems avoidable. First of all, could you remind me where those come from? (Restart or something?) And do you think it'd be plausible to replace those two meshes with one "canonical" copy after determining that they're equal? I wouldn't be opposed to adding tools to help with that.

majosm:
> I don't know that I love this, because pretty much by definition, we'll have redundant meshes flying around, which seems avoidable. First of all, could you remind me where those come from? (Restart or something?) And do you think it'd be plausible to replace those two meshes with one "canonical" copy after determining that they're equal? I wouldn't be opposed to adding tools to help with that.

It comes from [here](https://github.com/inducer/grudge/blob/3af1ca55d62fe1e7b187a63b6543ea35e3273e64/grudge/discretization.py#L96-L141) in inducer/grudge#285, where we promote the "part ID"s (rank, volume, or both depending on the problem) in the incoming mesh's adjacency data to a single data structure `PartID`. So each time a `dcoll` is created it makes a new mesh object. (We have two `dcoll`s temporarily when we restart using a different order than what was stored in the restart file.) We could maybe memoize that function, but a mesh seems like kind of a heavyweight thing to memoize?

inducer:
Could you explain the reason for that normalization step? Why is it needed? (I'm not suggesting that it isn't, I'm just not currently understanding.)

majosm:
That was added to provide a uniform way of retrieving the neighbor parts' rank/volume for all of the possible mesh configurations (single/multiple volume, distributed/non-distributed). It was suggested as an alternative to the way I had been doing it previously with the `part_id_helper`s (see inducer/grudge#239, specifically https://github.com/inducer/grudge/pull/239/commits/09bac09d98f54dce412fac6d0b4f52babce4916f).

inducer:
Still sort of puzzled, sorry. Let's talk about it during our meeting tomorrow.

inducer:
As discussed during meeting: We'll leave this open, but for now the creation of duplicate meshes seems avoidable by checking if the normalization step is a no-op.

majosm:
[Remove mesh copying in grudge dcoll init](https://trello.com/c/OLOylFhK/8-remove-mesh-copying-in-grudge-dcoll-init)

PR #361, opened 2022-11-12
Finish up typing meshmode.mesh.processing

alexfikl:
Following #357.

There are some extra changes in here for some functions that weren't handling the `group.vertex_indices == None` case, which `mypy` started complaining about once the functions were typed.

### DIFF
--- a/meshmode/mesh/processing.py
+++ b/meshmode/mesh/processing.py
@@ -726,11 +726,14 @@ def flip_simplex_element_group(
 
     # Swap the first two vertices on elements to be flipped.
 
-    new_vertex_indices = grp.vertex_indices.copy()
-    new_vertex_indices[grp_flip_flags, 0] \
-            = grp.vertex_indices[grp_flip_flags, 1]
-    new_vertex_indices[grp_flip_flags, 1] \
-            = grp.vertex_indices[grp_flip_flags, 0]
+    if grp.vertex_indices is not None:
+        new_vertex_indices = grp.vertex_indices.copy()
+        new_vertex_indices[grp_flip_flags, 0] \
+                = grp.vertex_indices[grp_flip_flags, 1]
+        new_vertex_indices[grp_flip_flags, 1] \
+                = grp.vertex_indices[grp_flip_flags, 0]
+    else:
+        new_vertex_indices = None
 
     # Apply the flip matrix to the nodes.
     flip_matrix = get_simplex_element_flip_matrix(grp.order, grp.unit_nodes)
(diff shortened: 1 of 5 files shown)


meshmode/mesh/processing.py
```diff
@@ -961,7 +950,7 @@ def split_mesh_groups(
             is_conforming=mesh.is_conforming)
 
     if return_subgroup_mapping:
-        return mesh, subgroup_to_group_map
+        return mesh, subgroup_to_group_map      # type: ignore[return-value]
```
inducer, line 953: We could instead be honest and declare the return type to be a `Union`. Would that work?

meshmode/mesh/processing.py
```diff
@@ -1314,11 +1316,11 @@ def map_mesh(mesh: Mesh, f: Callable[[np.ndarray], np.ndarray]) -> Mesh:
...
     if mesh._facial_adjacency_groups is not None:
         has_adj_maps = any([
-            hasattr(fagrp, "aff_map")
-            and (fagrp.aff_map.matrix is not None
-                or fagrp.aff_map.offset is not None)
+            fagrp.aff_map.matrix is not None or fagrp.aff_map.offset is not None
             for fagrp_list in mesh.facial_adjacency_groups
-            for fagrp in fagrp_list])
+            for fagrp in fagrp_list
+            if isinstance(fagrp, (InteriorAdjacencyGroup, InterPartAdjacencyGroup))
```
inducer, line 1322: This should include a check to ensure that not-known classes aren't sliently not considered.

inducer (commented):
Thanks! Two wrinkles that are, hopefully, easy to address.

meshmode/mesh/processing.py
alexfikl: Fixed in https://github.com/inducer/meshmode/compare/47349894ac4cd5d44460499a3b1e61231b4d86ee..8b5b8ac6c578291aa5cb55a5805ca740ae8d6a65

meshmode/mesh/processing.py
alexfikl: I reverted back to the original version that only checked `hasattr` 
https://github.com/inducer/meshmode/compare/47349894ac4cd5d44460499a3b1e61231b4d86ee..8b5b8ac6c578291aa5cb55a5805ca740ae8d6a65

Would you prefer to change that and use a stricter check?

inducer: I think `hasattr` is OK.

inducer:
Thanks!

PR #365, opened 2022-12-18
Add a simple piecewise C geometry

alexfikl:
This adds a piecewise construction, since that seemed easier. It looks like this:

<p align=center>
</img>
</p>

There's some knobs to control the radius of the outer circle, the inner circle and the size of the gap in front. It should only be C0 though. Do you think it's worth adding a Fourier interpolated smooth version?


inducer:
Looks $C^1$ to me/

inducer:
Thx!

alexfikl:
> Looks C1 to me?

Hm, it does, doesn't it. That should be good enough! :rocket:

PR #369, opened 2023-03-15
Fix for Firedrake removing `_Facets.markers`

inducer:
https://github.com/firedrakeproject/firedrake/commit/9125a65c0cb5bb671c62c33f05a0d42b983e06ed, in which 'deprecate' appears to mean 'remove without prior warning or suggested recourse'.

CI failure: https://github.com/inducer/meshmode/actions/runs/4395544902

### DIFF
--- a/meshmode/interop/firedrake/mesh.py
+++ b/meshmode/interop/firedrake/mesh.py
@@ -199,6 +199,25 @@ def _get_firedrake_boundary_tags(fdrake_mesh, tag_induced_boundary=False):
     return bdy_tags
 
 
+def _get_facet_markers(dm, facets):
+    # based on code removed in
+    # https://github.com/firedrakeproject/firedrake/commit/9125a65c0cb5bb671c62c33f05a0d42b983e06ed
+    import firedrake.cython.dmcommon as dmcommon
+
+    ids = np.empty_like(facets)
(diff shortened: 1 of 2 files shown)


meshmode/interop/firedrake/mesh.py
```diff
@@ -333,7 +333,11 @@ def _get_firedrake_facial_adjacency_groups(fdrake_mesh_topology,
                                   dtype=Mesh.face_id_dtype)
     # If only using some of the cells, throw away unused cells and
     # move to new cell index
-    ext_facet_markers = top.exterior_facets.markers
+
+    import firedrake.cython.dmcommon as dmcommon
+    label = dmcommon.FACE_SETS_LABEL
+    ext_facet_markers = top.topology_dm.getLabelIdIS(label).indices
```
inducer, line 339: @wence-, is this a halfway appropriate fix? (Thanks for looking!)

inducer:
Weird. This works when run on a `master` Firedrake install, but fails in CI?

meshmode/interop/firedrake/mesh.py
wence-: Ah, this picks up all faces (interior and exterior). Can you try with `label = "exterior_facets"` ?

PR #376, opened 2023-05-03
Deprecate `MPIMeshDistributor`

inducer:
cf. https://github.com/inducer/meshmode/pull/342.


majosm (approved):
LGTM apart from Mypy issue. I think it's a bad type annotation in mpi4py and can be ignored; the tutorial for `scatter` uses `None` as well (see https://mpi4py.readthedocs.io/en/stable/tutorial.html -- search for "Scattering Python objects").

PR #378, opened 2023-05-29
Fuse resampling indirections

kaushikcfd:
A meshmode-specific version of https://github.com/inducer/pytato/pull/425.

The implementation in https://github.com/inducer/pytato/pull/425 was not sufficient to fuse+fold the indirections in meshmode expression graphs.

Why draft:
- [ ] Add tests
- [ ] Docs


kaushikcfd:
The testing would be easier if we introduce this in github.com/inducer/grudge instead. (See https://github.com/inducer/grudge/pull/310)

PR #379, opened 2023-05-31
Add axis tags for reshapes, etc., in direction connection code

majosm:
These are part of what was causing the fallback warnings while compiling the Y3 prediction case.


meshmode/discretization/connection/direct.py
```diff
@@ -553,17 +556,22 @@ def _per_target_group_pick_info(
                     _FromGroupPickData(
                         from_group_index=source_group_index,
                         dof_pick_lists=actx.freeze(
```
inducer, line 558: I *think* it would suffice to just tag the element axis in the indirect access where this is used, below.

PR #381, opened 2023-06-21
_reshape_and_preserve_tags: support untaggable arrays

matthiasdiener:
(e.g. np.ndarray)

cc https://github.com/inducer/arraycontext/pull/235

### DIFF
--- a/meshmode/discretization/connection/direct.py
+++ b/meshmode/discretization/connection/direct.py
@@ -48,10 +48,12 @@ from dataclasses import dataclass
 def _reshape_and_preserve_tags(
         actx: ArrayContext, ary: ArrayT, new_shape: Tuple[int, ...]) -> ArrayT:
     try:
-        return actx.tag(ary.tags, ary.reshape(new_shape))
+        tags = ary.tags
     except AttributeError:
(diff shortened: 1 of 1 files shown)


inducer:
I'm not sure why that'd be necessary. For arrays that don't do tagging, wouldn't `actx.tag` be a no-op?

matthiasdiener:
> I'm not sure why that'd be necessary. For arrays that don't do tagging, wouldn't `actx.tag` be a no-op?

`actx.tag` is a no-op in https://github.com/inducer/arraycontext/pull/235, but `ary.tags` won't work (since `ary` is an ndarray in that case).

inducer:
> `actx.tag` is a no-op in [inducer/arraycontext#235](https://github.com/inducer/arraycontext/pull/235), but `ary.tags` won't work (since `ary` is an ndarray in that case).

That makes sense. But in that case, only the access `ary.tags` should be in the `try` block.

matthiasdiener:
> That makes sense. But in that case, only the access `ary.tags` should be in the `try` block.

What do you think of 50f3134?

inducer:
LGTM, thanks!

PR #384, opened 2023-08-04
Tensor product 1d nodes & basis functions

a-alveyblanc:
Without access to the 1D nodes and 1D basis functions used in the tensor product, we are not able to exploit the tensor product structure to reduce the cost required for operator evaluation, e.g. applying a differentiation operator to function data.

This PR:
- Adds functionality to `TensorProductElementGroupBase` to store and retrieve 1D basis functions and 1D nodes used in the tensor product
- Updates subclasses accordingly

Depends on:
- [x] https://github.com/inducer/modepy/pull/72

### DIFF
--- a/meshmode/discretization/poly_element.py
+++ b/meshmode/discretization/poly_element.py
@@ -41,6 +41,7 @@ from meshmode.discretization import (
         InterpolatoryElementGroupBase)
 
 import modepy as mp
+from modepy import Basis
 
 __doc__ = """
 Group types
@@ -518,7 +519,9 @@ class HypercubeElementGroupBase(NodalElementGroupBase):
 
 class TensorProductElementGroupBase(PolynomialElementGroupBase,
         HypercubeElementGroupBase):
-    def __init__(self, mesh_el_group, order, index=None, *, basis, unit_nodes):
+    def __init__(self, mesh_el_group: _MeshTensorProductElementGroup,
+                 order: int, index=None, *, basis: Basis,
+                 unit_nodes: np.ndarray) -> None:
         """
         :arg basis: a :class:`modepy.TensorProductBasis`.
         :arg unit_nodes: unit nodes for the tensor product, obtained by
@@ -530,15 +533,42 @@ class TensorProductElementGroupBase(PolynomialElementGroupBase,
             raise ValueError("basis dimension does not match element group: "
                     f"expected {mesh_el_group.dim}, got {basis._dim}.")
 
+        if isinstance(basis, mp.TensorProductBasis):
+            for b in basis.bases:
+                if b._dim != 1:
+                    raise NotImplementedError(
+                        "All bases used to construct the tensor "
+                        "product must be of dimension 1. Support "
+                        "for higher-dimensional component bases "
+                        "does not yet exist.")
+
         if unit_nodes.shape[0] != mesh_el_group.dim:
             raise ValueError("unit node dimension does not match element group: "
-                    f"expected {mesh_el_group.dim}, got {unit_nodes.shape[0]}.")
+                    f"expected {self.mesh_el_group.dim}, "
+                    f"got {unit_nodes.shape[0]}.")
+
+        # NOTE there are cases where basis is a 1D `_SimplexONB` object. We wrap
+        # in a TensorProductBasis object if this is the case
+        if not isinstance(basis, mp.TensorProductBasis):
+            if basis._dim == 1 and unit_nodes.shape[0] == 1:
+                basis = mp.TensorProductBasis([basis])
+            else:
+                raise ValueError("`basis` is not a TensorProductBasis object, "
+                                 "and `basis` and `unit_nodes` are not both of "
+                                 "dimension 1. Found `basis` dim = {basis._dim}, "
+                                 "`unit_nodes` dim = {unit_nodes.shape[0]}.")
 
         self._basis = basis
+        self._bases_1d = basis.bases[0]
         self._nodes = unit_nodes
 
     def basis_obj(self):
         return self._basis
+
+    def bases_1d(self):
+        """Return 1D component bases used to construct the tensor product basis.
+        """
+        return self._bases_1d
 
     @memoize_method
     def quadrature_rule(self):
@@ -548,6 +578,11 @@ class TensorProductElementGroupBase(PolynomialElementGroupBase,
         weights = np.dot(mass_matrix,
                          np.ones(len(basis_fcts)))
         return mp.Quadrature(nodes, weights, exact_to=self.order)
+
+    @property
+    @memoize_method
+    def unit_nodes_1d(self):
+        return self._nodes[0][:self.order + 1].reshape(1, self.order + 1)
 
     def discretization_key(self):
         # FIXME?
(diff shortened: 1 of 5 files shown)


meshmode/discretization/poly_element.py
```diff
@@ -518,11 +518,16 @@ def from_mesh_interp_matrix(self):
 
 class TensorProductElementGroupBase(PolynomialElementGroupBase,
         HypercubeElementGroupBase):
-    def __init__(self, mesh_el_group, order, index=None, *, basis, unit_nodes):
+    def __init__(self, mesh_el_group, order, index=None, *, basis, basis_1d,
+                 unit_nodes, unit_nodes_1d):
```
alexfikl, line 522: It feels a bit weird to pass in both `(basis, unit_nodes)` and `(basis_1d, unit_nodes_1d)` here since we can construct the first one from the second. Maybe move the construction of the nd ones here?

meshmode/discretization/poly_element.py
```diff
@@ -518,11 +518,16 @@ def from_mesh_interp_matrix(self):
...
         """
         :arg basis: a :class:`modepy.TensorProductBasis`.
+        :arg basis_1d: a representation of the 1D basis from which the tensor
+        product basis is built.
         :arg unit_nodes: unit nodes for the tensor product, obtained by
             using :func:`modepy.tensor_product_nodes`, for example.
+        :arg unit_nodes_1d: the one-dimensional set of nodes used in the tensor
+        product, i.e. the set of nodes from which `unit_nodes` is constructed.
```
alexfikl, line 530:
```suggestion
        :arg basis_1d: a representation of the 1D basis from which the tensor
        	product basis is built.
        :arg unit_nodes: unit nodes for the tensor product, obtained by
            using :func:`modepy.tensor_product_nodes`, for example.
        :arg unit_nodes_1d: the one-dimensional set of nodes used in the tensor
        	product, i.e. the set of nodes from which `unit_nodes` is constructed.
```
These need to be indented for Sphinx to parse it properly (I think?)

meshmode/discretization/poly_element.py
```diff
@@ -549,6 +559,10 @@ def quadrature_rule(self):
                          np.ones(len(basis_fcts)))
         return mp.Quadrature(nodes, weights, exact_to=self.order)
 
+    @property
+    def unit_nodes_1d(self):
+        return np.array([self._nodes_1d])
```
alexfikl, line 564: Why is the `np.array` needed?

alexfikl (commented):
Left a few comments that will hopefully simplify things, but otherwise looks good!

meshmode/discretization/poly_element.py
a-alveyblanc: Fixed this for 1D nodes in my latest push. 1D basis functions doesn't seem to have a function like `mp.tensor_product_nodes`, so still working on that.

meshmode/discretization/poly_element.py
a-alveyblanc: Added a comment to clarify this. In short, `modepy` expects a nested array as an argument when evaluating the basis functions.

.pylintrc-local.yml
inducer, line 1: This wants to stay.

meshmode/discretization/poly_element.py
alexfikl: I don't know if one is needed. You should be able to just construct a `TensorProductBasis` like here (for a more general case)
https://github.com/inducer/modepy/blob/0fa7355ebe979a4706a7da2ae15b7517d229b8b3/modepy/modes.py#L1219-L1232

Would that work?

meshmode/discretization/poly_element.py
```diff
@@ -530,16 +531,37 @@ def __init__(self, mesh_el_group, order, index=None, *, basis, unit_nodes):
...
+                    f"expected {self.mesh_el_group.dim}, "
+                    f"got {unit_nodes.shape[0]}.")
 
         self._basis = basis
+
+        # FIXME?? there are cases where `basis` is a _SimplexONB object, not a
+        # TensorProductBasis object. in those cases, it seems self._bases_1d is
+        # unused. leaving as-is for now.
+        if isinstance(basis, mp.TensorProductBasis):
+            self._bases_1d = basis.bases[0]
```
inducer, line 553: Instead: If it's a 1D affair (check on both basis and nodes before making that decision), and the basis is not TP, wrap the basis in a `TensorProductBasis` for consistency.

meshmode/discretization/poly_element.py
```diff
@@ -518,7 +518,8 @@ def from_mesh_interp_matrix(self):
 
 class TensorProductElementGroupBase(PolynomialElementGroupBase,
         HypercubeElementGroupBase):
-    def __init__(self, mesh_el_group, order, index=None, *, basis, unit_nodes):
+    def __init__(self, mesh_el_group, order, index=None, *, basis,
+                 unit_nodes):
```
inducer, line 522: Try type-annotating?

a-alveyblanc:
This is ready for a final look I think. Only thing I'm unsure about is the type annotations I added to `TensorProductElementGroup`

meshmode/discretization/poly_element.py
```diff
@@ -530,16 +534,44 @@ def __init__(self, mesh_el_group, order, index=None, *, basis, unit_nodes):
             raise ValueError("basis dimension does not match element group: "
                     f"expected {mesh_el_group.dim}, got {basis._dim}.")
 
+        if isinstance(basis, mp.TensorProductBasis):
+            for b in basis.bases:
+                if b._dim != 1:
+                    raise ValueError("All bases used to construct the tensor "
```
inducer, line 540:
```suggestion
                    raise NotImplementedError("All bases used to construct the tensor "
```

meshmode/discretization/poly_element.py
```diff
@@ -530,16 +534,44 @@ def __init__(self, mesh_el_group, order, index=None, *, basis, unit_nodes):
...
+                    f"expected {self.mesh_el_group.dim}, "
+                    f"got {unit_nodes.shape[0]}.")
+
+        # NOTE there are cases where basis is a 1D `_SimplexONB` object. We wrap
+        # in a TensorProductBasis object if this is the case
+        if not isinstance(basis, mp.TensorProductBasis):
+            if basis._dim == 1 and unit_nodes.shape[0] == 1:
+                from modepy.modes import _get_orth_weight
+                basis = mp.TensorProductBasis([basis],
+                                                    _get_orth_weight([basis]))
```
inducer, line 556:
```suggestion
                basis = mp.TensorProductBasis([basis])
```
(see https://github.com/inducer/modepy/pull/72)

inducer (approved):
LGTM with those two fixes.

inducer:
I tried to do a rebuild, but the container pull is failing with some sort of AWS signature message that I don't know how to fix. Maybe https://quay.io is broken ATM?

a-alveyblanc:
Other tests are passing on AMD Epyc, though. Is there a large difference between other tests and Firedrake? Grudge tests passed on Epyc: `test/test_dt_utils.py::test_wave_dt_estimate[<PytatoPyOpenCLArrayContext for <pyopencl.Device 'cpu-AMD EPYC 7763 64-Core Processor' on 'Portable Computing Language'>>-2-1]`

inducer:
They are! It depends on how pocl is installed. By and large, we install pocl from conda, and that works fine. For Firedrake, we install pocl via pip (yes, that's possible... Isuru worked some magic). That's the bit that's broken.

alexfikl:
> They are! It depends on how pocl is installed. By and large, we install pocl from conda, and that works fine. For Firedrake, we install pocl via pip (yes, that's possible... Isuru worked some magic). That's the bit that's broken.

Doesn't it install it from the Ubuntu repos?
https://github.com/inducer/meshmode/blob/9e942e6f602a21d433ffbae76f6c4b3089166e4f/.ci/install-for-firedrake.sh#L10

inducer:
Good catch! Thanks for spotting that. That makes for an obvious thing to try: Use pocl from pypi. :)

PR #386, opened 2023-08-11
Raise an error if no vis_order is provided

alexfikl:
Otherwise this would fail in a confusing manner somewhere deeper in `modepy` when we actually try to use it.


inducer:
Thanks!

PR #387, opened 2023-08-11
DOFArray: support untaggable arrays in setstate

matthiasdiener:
This ~~should fix~~ fixes restart with Numpy actx. cc @MTCam


inducer:
Argh. Should've squashed, not rebased. Oh well.

PR #396, opened 2024-03-05
Remove MeshElementGroup deprecations

### LINKED ISSUE #224: Element groups shouldn't have to change when joining a mesh/discretization

alexfikl:
All of these seem to be at least a year overdue and a quick grep showed that nothing is using them anymore (in `main` at least).

Inspired by #395. 
Fixes #224.


alexfikl:
This enables `flake8-isort` too, but only fixes the two files that were modified here. I can make another PR that does a `python -m isort` over everything?

inducer:
I'm all in favor (also of getting the isort done).

alexfikl:
> I'm all in favor (also of getting the isort done).

Got it! This still needs a bit of work apparently, so I'll ping you when it's done.

meshmode/mesh/__init__.py
```diff
@@ -1000,16 +869,6 @@ def __init__(self, vertices, groups, *, skip_tests=False,
...
 
-        el_nr = 0
-        node_nr = 0
-
-        new_groups = []
-        for g in groups:
-            ng = replace(g, element_nr_base=el_nr, node_nr_base=node_nr)
-            new_groups.append(ng)
-            el_nr += ng.nelements
-            node_nr += ng.nnodes
```
alexfikl, line 1011: This doesn't create a new list of groups with all the groups copied over, since we don't need to renumber them. That shouldn't be an issue?

test/test_firedrake_interop.py
```diff
@@ -377,12 +385,13 @@ def test_bdy_tags(square_or_cube_mesh, bdy_ids, coord_indices, coord_values,
      ("UnitSquare", [10, 20, 30], 2),
      ("UnitCube", [10, 20, 30], 3),
      ("blob2d-order1", ["8e-2", "6e-2", "4e-2"], 2),
-     ("blob2d-order4", ["8e-2", "6e-2", "4e-2"], 2),
+     pytest.param("blob2d-order4", ["8e-2", "6e-2", "4e-2"], 2,
+                  marks=pytest.mark.xfail),
```
alexfikl, line 389: This test seems to be failing for a while now. Should I leave it be an open an issue?

alexfikl:
@inducer This seems to be passing now! 

The last 3 commits are fairly big, but just move things around due to isort and some cleaning up in the firedrake tests (it now uses `actx_factory`!), so probably better to look at just [4e71cfa](https://github.com/inducer/meshmode/pull/396/commits/4e71cfa4c91930d4f33d8d65d0f37cd56e6dbecf).

test/test_firedrake_interop.py
inducer: Yeah. I suppose that's the grown-up approach to this. I had just left it at a failing CI to remind myself to debug it. (I had started at a point, but never got close to figuring out what the issue was.)

inducer (commented):
LGTM, thanks! Just one wrinkle, then this is good to go.

meshmode/mesh/__init__.py
```diff
@@ -251,43 +238,9 @@ class MeshElementGroup(ABC):
     """
 
     order: int
-
-    # NOTE: the mesh supports not having vertices if no facial or nodal
-    # adjacency is required, so we can mark this as optional
```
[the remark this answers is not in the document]
alexfikl: Oops, yes, definitely. I added this to the docs themselves (which was the intention when deleting it).

https://github.com/inducer/meshmode/compare/9916d684cbd29746df7bc6b0c423bdb1cf25c6b5..65ef869eb85252e7293fad39eba04028246e75fa

test/test_firedrake_interop.py
alexfikl: Opened an issue #397. Is there anyone from the Firedrake side that is using this? Maybe they have a better idea of what changed in November-ish last year.. :\

alexfikl:
@inducer Re-added the missing note, so this should be ok now!

inducer:
Thanks!

PR #398, opened 2024-03-06
Allow passing scalars (and containers of scalars) to `DirectDiscretizationConnection`

majosm:
`project(...)` in grudge currently fails when passed an object array of numbers, because it only checks whether the entire input is a number ([code](https://github.com/inducer/grudge/blob/4cab8a846c447bbcd48f2872eb476d77a6bd9840/grudge/projection.py#L81)), not whether it's an array container of them. Checking down inside `DirectDiscretizationConnection` instead fixes that.

### DIFF
--- a/meshmode/discretization/connection/direct.py
+++ b/meshmode/discretization/connection/direct.py
@@ -596,7 +596,10 @@ class DirectDiscretizationConnection(DiscretizationConnection):
 
         # {{{ recurse into array containers
 
-        if not isinstance(ary, DOFArray):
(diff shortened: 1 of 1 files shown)


inducer:
LGTM.

When this is done, could you remove the old (and then-redundant) number handling from `project`?

https://github.com/inducer/grudge/blob/df1c6f28fb8283a6af52edc5df4d7c1aa5f9d85d/grudge/projection.py#L81-L82

PR #400, opened 2024-04-07
Make Mesh into a dataclass

alexfikl:
This starts making `Mesh` into a dataclass. Right now it's mostly just porting away from `pytools.Record`, since it's hard to make it a clean dataclass while being backwards compatible.

The main additions are:
* a `make_mesh` factory function that does all the work of the Mesh constructor.
* a `check_mesh_consistency` function that does all the checking. This seemed useful to make public, since the dataclass version won't do any checking on `replace(mesh, ...)` and it may still be wanted sometimes?
* ~`get_nodal_adjacency` and `get_facial_adjacency_groups` to replace the property versions.~

~The nodal / facial adjacency attributes were very annoying. Currently it overwrites `__getattribute__` to make it behave a bit like before. Let me know if you have some better ideas on how to make that work..~

This is still a WIP, but I had it lying around for a while, so might as well. Things still to do:
* [x] Run the CI and make everything actually works. Currently it *works on my machine*™
* [x] Left a few remarks / questions below that should be looked at.
* [x] Port at least the internal uses of the various functions to avoid deprecation warnings.
* [x] Not quite sure why the documentation build is failing? It seems to dislike the tikz stuff, but that shouldn't have changed..


meshmode/mesh/__init__.py
```diff
@@ -730,52 +733,480 @@ def as_python(self):
...
+        groups: Iterable[MeshElementGroup],
+        *,
+        vertices: Optional[np.ndarray] = None,
+        nodal_adjacency: Optional[NodalAdjacencyLike] = None,
+        facial_adjacency_groups: Optional[FacialAdjacencyLike] = None,
+        is_conforming: Optional[bool] = None,
+        # dtypes
+        vertex_id_dtype: DTypeLike = "uint32",
+        element_id_dtype: DTypeLike = "uint32",
+        face_id_dtype: DTypeLike = "uint8",
```
alexfikl, line 799: Made these `uint` instead of `int`. Does that break anything?

meshmode/mesh/__init__.py
```diff
@@ -730,52 +733,480 @@ def as_python(self):
...
+    # TODO: the default is deprecated
+    is_conforming: Optional[bool] = None
+    """*True* if it is known that all element interfaces are conforming. *False*
+    if it is known that some element interfaces are non-conforming. *None* if
+    neither of the two is known.
+    """
 
-    .. attribute:: vertices
+    # TODO: defaults are deprecated
+    vertex_id_dtype: np.dtype = np.dtype(np.uint32)
```
alexfikl, line 972: Not quite sure how defaults work with `init=False` in dataclasses. Are these needed at all? Should try that out..

meshmode/mesh/__init__.py
```diff
@@ -730,52 +733,480 @@ def as_python(self):
...
+    face_id_dtype: np.dtype = np.dtype(np.uint8)
+    """The :class:`~numpy.dtype` used to index element faces (relative to each
+    element).
+    """
 
-        *None* or an array of vertex coordinates with shape
-        *(ambient_dim, nvertices)*. If *None*, vertices are not
-        known for this mesh.
+    # TODO: these have the wrong types for the moment to silence mypy
+    nodal_adjacency: NodalAdjacency = None
```
alexfikl, line 984: The actual type should be `Union[None, Literal[False], NodalAdjacency]`, but mypy complains on all calls to `mesh.nodal_adjacency` because it doesn't know we did some `__getattribute__` magic below..

meshmode/mesh/__init__.py
```diff
@@ -730,52 +733,480 @@ def as_python(self):
...
+             "standard 'replace' function.",
+             DeprecationWarning, stacklevel=2)
+
+        # TODO: these should be removed with __getattribute__
+        kwargs["nodal_adjacency"] = kwargs.get(
+            "nodal_adjacency", self._nodal_adjacency)
+        kwargs["facial_adjacency_groups"] = kwargs.get(
+            "facial_adjacency_groups", self._facial_adjacency_groups)
+
+        return replace(self, **kwargs)
```
alexfikl, line 1092: The `copy` isn't really deprecated, but I'm not quite sure what to do with it. The problem is that just doing a `replace(mesh, ...)` will force a computation of `nodal_adjacency` because of that `__getattribute__` hackery..

meshmode/mesh/processing.py
```diff
@@ -451,6 +451,9 @@ def _get_mesh_part(
 
     .. versionadded:: 2017.1
     """
+    if mesh.vertices is None:
+        raise ValueError("Mesh must have vertices")
+
```
alexfikl, line 456: These were added for `mypy`. Apparently now that the mesh is a dataclass, it suddenly realized the vertices can be None :grin:

meshmode/mesh/__init__.py
alexfikl: Also, should this set `_factory_constructed=False`? Before, the consistency checks were run on each `Mesh(...)` call, but now they would just be called if the original mesh was not made with `make_mesh(...)`

meshmode/mesh/__init__.py
alexfikl: Well, breaks `firedrake` at least, but that's easily fixable. Might be better to revert..

meshmode/mesh/__init__.py
alexfikl: Seems like it works just fine!

meshmode/mesh/__init__.py
alexfikl: Might be easier to just use a different name for the dataclass fields? Suggestions?

meshmode/mesh/__init__.py
alexfikl: Might be easier to just use a different name for the dataclass fields? Suggestions?

meshmode/mesh/__init__.py
```diff
@@ -730,52 +733,479 @@ def as_python(self):
...
+    """
 
-    .. attribute:: vertices
+    vertex_id_dtype: np.dtype
+    """The :class:`~numpy.dtype` used to index into the vertex array."""
+    element_id_dtype: np.dtype
+    """The :class:`~numpy.dtype` used to index into the element array (relative
+    to each group).
+    """
+    face_id_dtype: np.dtype
```
alexfikl, line 976: Firedrake is currently failing because this isn't a class attribute anymore.

inducer: Seems to be OK now?

meshmode/mesh/__init__.py
inducer: If it isn't meant to be deprecated, why have a warning that it is? I'm not sure I understand.

meshmode/mesh/__init__.py
```diff
@@ -730,52 +733,479 @@ def as_python(self):
...
+                 "removed in 2025. Use 'get_facial_adjacency_groups' instead.",
+                 DeprecationWarning, stacklevel=2)
+
+            return self.get_facial_adjacency_groups()
+        else:
+            return object.__getattribute__(self, name)
+
+    @property
+    def _nodal_adjacency(self) -> NodalAdjacency:
+        return object.__getattribute__(self, "nodal_adjacency")
```
inducer, line 1167: Is this just to hide this attribute access from mypy?

meshmode/mesh/__init__.py
alexfikl: Yeah, I'll remove that. I added it before I realized that the special handling for the adjacency stuff is still required..

Ideally we would deprecate it and rely on `dataclasses.replace` though.

meshmode/mesh/__init__.py
alexfikl: Yeah, fixed the firedrake code.

meshmode/mesh/__init__.py
alexfikl: To explain the issue, hopefully a bit more clearly, e.g. for `nodal_adjacency`:
* On `main`, `nodal_adjacency` is a property that caches the construction of the adjacency.
* With dataclasses, I thought `nodal_adjacency` should a simple attribute so that it gets copied around on `replace`. That's why there's a `get_nodal_adjancency` method now that does the caching.

The property thing really clashes with making it a dataclass field, so I tried to hack around it, but not sure that's very nice..

inducer: Maybe I'm being dense here, but why not have the adjacency-related dataclass attributes be underscored?

alexfikl: Is that ok? I mostly shied away from that because `Mesh(..., _nodal_adjancency=nodal_adjancency)` looks like we're doing something hidden.

meshmode/mesh/__init__.py
alexfikl: No, it's worse.. it's to not go through the `Mesh.__getattribute__` hackery :( This would very nicely go away with your underscoring suggestion though!

meshmode/mesh/__init__.py
inducer: That seems totally fine to me. The constructor is only called internally anyway, at least in the future.

meshmode/mesh/__init__.py
alexfikl: This was resolved by https://github.com/inducer/meshmode/compare/12a8c99be32e13ae8e01b951840f5e69f66edbdf..6abd2d59ee950bbd0eb91f866c26244aa4b89ea6.

meshmode/mesh/__init__.py
alexfikl: Went with your suggestion and made the dataclass fields underscored for those two in
https://github.com/inducer/meshmode/compare/6abd2d59ee950bbd0eb91f866c26244aa4b89ea6..e2f75e8391145ae918392f5a665324c9d5182a9c

That removed a lot of hacks and other things, so it's quite nice!

alexfikl:
@inducer I think all of my initial questions are solved now, so this is ready for a proper review.

>  Port at least the internal uses of the various functions to avoid deprecation warnings.

I'm thinking of leaving this for a followup PR, since this is sufficiently big?

meshmode/mesh/__init__.py
```diff
@@ -730,52 +733,503 @@ def as_python(self):
...
+    """
+    face_id_dtype: np.dtype
+    """The :class:`~numpy.dtype` used to index element faces (relative to each
+    element).
+    """
 
-        *None* or an array of vertex coordinates with shape
-        *(ambient_dim, nvertices)*. If *None*, vertices are not
-        known for this mesh.
+    # TODO: Once the @property(nodal_adjacency) is passed its deprecation period
```
inducer, line 994:
```suggestion
    # TODO: Once the @property(nodal_adjacency) is past its deprecation period
```

meshmode/mesh/__init__.py
```diff
@@ -730,52 +733,504 @@ def as_python(self):
...
+    """A description of the nodal adjacency of the mesh. This can be *False* if
+    no adjacency is known or should be computed, *None* to compute the adjacency
+    on demand or a given :class:`NodalAdjacency` instance.
 
-    .. attribute:: groups
+    To obtain the appropriate nodal adjacency use :meth:`~Mesh.get_nodal_adjacency`.
+    The method will compute the adjacency, if necessary, and cache it in this
+    attribute.
+    """
+    _facial_adjacency_groups: \
```
inducer, line 1006: Add blank lines between documented attributes, for visual separation?

meshmode/mesh/__init__.py
```diff
@@ -730,52 +733,504 @@ def as_python(self):
...
+
+        return self.vertices.dtype
+
+    @property
+    def nodal_adjacency(self) -> NodalAdjacency:
+        warn("The 'nodal_adjacency' property is deprecated and will be removed "
+             "in 2025. Use 'Mesh.get_nodal_adjacency()' instead.",
+             DeprecationWarning, stacklevel=2)
+
+        return self.get_nodal_adjacency()
```
inducer, line 1188: I'm not sure I even agree. Why not leave `@property` `nodal_adjacency` in place indefinitely, and not even introduce `get_nodal_adjacency`?

alexfikl: My thinking was mostly that
* doing `replace(mesh, _nodal_adjacency=new_nodal_adjacency)` is weird with the underscore.
* having a `property` potentially do heavy work is not very clear (was there even something in pep8 about that?)

inducer: > doing `replace(mesh, _nodal_adjacency=new_nodal_adjacency)` is weird with the underscore.

If it were in user code, I would agree. But I can't imagine a use case for that in user code. ("I've made this mesh without adj information, but oh wait, I've just discovered I have that information sitting around after all?")

> having a property potentially do heavy work is not very clear

:shrug: Particularly if it's a cached thing (so that it's asymptotically cheap), I don't feel very strongly there. OTOH, `mesh.thing` reads much more nicely to me than `mesh.get_thing()`.

alexfikl: Yeah, agreed it reads nicer. Fair enough, I'll change those back!

alexfikl: Done! https://github.com/inducer/meshmode/compare/7aa1897d2f0079f4ff6f854f40991652aac95946..2597d8283052430e5d9bd751d6a735009ad6e07f

inducer:
Pushed a commit that ports the internal uses of `Mesh()`. Mostly a `sed` job.

inducer:
I guess I sort of understand why you might have changed the `make_mesh` arguments to `groups, *, vertices`, but I don't think it's worth the headache. 99% of meshes have vertices, and the 1% that don't can just pass `None` for that first argument.

alexfikl:
> I guess I sort of understand why you might have changed the `make_mesh` arguments to `groups, *, vertices`, but I don't think it's worth the headache. 99% of meshes have vertices, and the 1% that don't can just pass `None` for that first argument.

That's fair enough. It definitely makes the porting easier :grin: Thanks for helping with that and fixing all the issues!

inducer:
Thanks for getting this most of the way!

PR #401, opened 2024-04-07
ci: fix tikz generation

alexfikl:
The CI was complaining that there's no `standalone.cls` when building tikz pictures.
https://github.com/inducer/meshmode/actions/runs/8585855922/job/23527696674

This installs `texlive-latex-extra` to fix that. At least that's where `standalone.cls` is on Arch, so fingers crossed :worried:


inducer:
Thx!

PR #404, opened 2024-04-15
Fix, test `generate_urchin`

inducer:
cc @gaohao95


alexfikl:
Hm, `sphinxcontrib-tikz` seems to be causing more problems than it's worth. I can see about replacing it with some static SVGs instead? :\

inducer:
IDK, not yet. All they have to do is not keep breaking stuff. If they do, I feel like I'd just pin the version to something that works and call it a day.

PR #409, opened 2024-04-29
Remove `TensorProductElementGroupBase.bases_1d`

inducer:
To get the tests in https://github.com/inducer/grudge/pull/338 to go, zero-dimensional tensor product discretizations need to work. This `bases_1d` thing (which seems kind of ill-defined anyway) was an obstacle. It appeared entirely unused across meshmode and grudge.

@a-alveyblanc Do you have any objections to this?


a-alveyblanc:
This is necessary to get the 1D basis (bases) used to build the TP basis so that we can retrieve 1D operators.

inducer/grudge#313 depends on bases_1d for fast operator evaluation to work

(sorry if formatting is bad, I'm replying on GH mobile)

inducer:
OK, thanks!

PR #419, opened 2024-07-13
Fix some deprecations

alexfikl:
Mostly around `inverse_mass_matrix` preferring a `Basis` now, since it was causing a bunch of warnings in grudge as well :grin:

### DIFF
--- a/meshmode/discretization/poly_element.py
+++ b/meshmode/discretization/poly_element.py
@@ -296,6 +296,11 @@ class _MassMatrixQuadratureElementGroup(PolynomialSimplexElementGroupBase):
         return mp.Quadrature(nodes, weights, exact_to=self.order)
 
     @property
+    @memoize_method
+    def unit_nodes(self):
+        return self._interp_nodes
+
+    @property
     @abstractmethod
     def _interp_nodes(self):
         """Returns a :class:`numpy.ndarray` of shape ``(dim, nunit_dofs)``
--- a/meshmode/discretization/poly_element.py
+++ b/meshmode/discretization/poly_element.py
@@ -126,7 +126,7 @@ def mass_matrix(grp: InterpolatoryElementGroupBase) -> np.ndarray:
 
     assert grp.is_orthonormal_basis()
     return mp.mass_matrix(
-            grp.basis_obj().functions,
+            grp.basis_obj(),
             grp.unit_nodes)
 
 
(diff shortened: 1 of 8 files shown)


meshmode/discretization/poly_element.py
```diff
@@ -288,13 +288,18 @@ def quadrature_rule(self):
...
+        mass_matrix = mp.mass_matrix(basis, nodes)
         weights = np.dot(mass_matrix,
-                         np.ones(len(basis_fcts)))
+                         np.ones(len(basis.functions)))
         return mp.Quadrature(nodes, weights, exact_to=self.order)
 
+    @property
+    @memoize_method
+    def unit_nodes(self):
+        return self._interp_nodes
```
alexfikl, line 301: Added this due to the failure in here
https://github.com/inducer/meshmode/actions/runs/9920172408/job/27406747663?pr=419

Basically the `test_sanity_qhull_nd` test creates a 4D mesh which uses the monomial basis returned by `mp.basis_for_space`. This then gets passed on to `mass_matrix`, which complains about a non-orthogonal basis and raises an exception.

However, that test just required `grp.unit_nodes`, which only went though the `mass_matrix` call because the base class was doing `quadrature_rule().nodes`. This overwrites that and returns the nodes directly.

test/test_array.py
```diff
@@ -142,33 +142,27 @@ def _get_test_containers(actx, ambient_dim=2):
...
-    # }}}
-
-    # {{{ flat_norm
-
-    # check nested vs actx.np.linalg.norm
-    assert actx.to_numpy(abs(
-            flat_norm(c_test[1], ord=ord)
-            - actx.np.linalg.norm(c_test[1], ord=ord))) < 1e-12
+    # check nested
+    assert actx.to_numpy(abs(flat_norm(c_test[1], ord=ord))) != 0
```
inducer, line 158: I'm not sure I understand the reasoning here. Why is asserting non-zero-ness useful here?

alexfikl: Oops, put that in as a placeholder and forgot to update it!

alexfikl: Compared it against numpy in https://github.com/inducer/meshmode/compare/e78d829329bb7a22cca9159e6f4f316516b987da..3ea05d6e8a0e19f70242e6e3153aa48bd3c249ba

meshmode/discretization/poly_element.py
inducer: That seems fine. Thanks for flagging it.

PR #421, opened 2024-08-01
Cleanups 2024

inducer:


meshmode/mesh/__init__.py
```diff
@@ -929,22 +943,22 @@ def is_mesh_consistent(
...
+        nodal_adjacency: NodalAdjacencyLike | None = None,
+        facial_adjacency_groups: FacialAdjacencyLike | None = None,
+        is_conforming: bool | None = None,
         # dtypes
-        vertex_id_dtype: DTypeLike = "int32",
-        element_id_dtype: DTypeLike = "int32",
-        face_id_dtype: DTypeLike = "int8",
+        vertex_id_dtype: DTypeLike = np.dtype("int32"),  # noqa: B008
+        element_id_dtype: DTypeLike = np.dtype("int32"),  # noqa: B008
+        face_id_dtype: DTypeLike = np.dtype("int8"),  # noqa: B008
```
alexfikl, line 955: I think this breaks Sphinx. 

It doesn't like it when defaults are functions and gives up on trying to create links for the types and other things. Haven't checked in a while, so it might work now :confused:

alexfikl (commented):
Quite surprised this doesn't break anything. Woohoo :tada:

meshmode/mesh/__init__.py
alexfikl: Hm, seems to work. Nevermind!

PR #423, opened 2024-08-06
Port to actx.np.zeros

alexfikl:
Plus two unrelated commits that popped out while grepping for `actx.zeros` :grin:

### DIFF
--- a/meshmode/discretization/__init__.py
+++ b/meshmode/discretization/__init__.py
@@ -343,9 +343,7 @@ class Discretization:
 
     .. automethod:: __init__
     .. automethod:: copy
-    .. automethod:: empty
     .. automethod:: zeros
-    .. automethod:: empty_like
     .. automethod:: zeros_like
     .. automethod:: nodes
     .. automethod:: num_reference_derivative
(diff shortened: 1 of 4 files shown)


meshmode/discretization/__init__.py
```diff
@@ -472,10 +472,14 @@ def empty(self, actx: ArrayContext,
...
         if not isinstance(actx, ArrayContext):
-            raise TypeError("'actx' must be an ArrayContext, not '%s'"
-                    % type(actx).__name__)
+            raise TypeError(
+                f"'actx' must be an ArrayContext, not '{type(actx).__name__}'")
 
-        return self._new_array(actx, actx.empty, dtype=dtype)
+        warn(f"'{type(self).__name__}.empty' is deprecated and will be removed "
+             f"in 2025. Use '{type(self).__name__}.zeros' instead.",
+             DeprecationWarning, stacklevel=2)
```
alexfikl, line 480: This also adds a warning that `Discretization.empty` is deprecated following arraycontext.

meshmode/mesh/__init__.py
```diff
@@ -361,13 +361,12 @@ def vertex_unit_coordinates(self) -> np.ndarray:
         return mp.unit_vertices_for_shape(self._modepy_shape).T
 
     @classmethod
-    # Type ignore because I do not know how to annotate this factory
-    # function usage. Possible reference:
-    # https://github.com/python/typing/discussions/1191#discussioncomment-2796173
-    def make_group(cls, order: int,  # type: ignore[override]
+    def make_group(cls,
+                   order: int,
```
alexfikl, line 365: Unless I misunderstood, `mypy` was upset here because the base class only had `**kwargs`. Also adding `*args` like used here seems to fix it. Is that not ok?

inducer: That's fine IMO.

inducer:
Thx!

PR #424, opened 2024-08-09
Fix issues with ruff 0.5.7

alexfikl:
This adds an ignore for https://docs.astral.sh/ruff/rules/unnecessary-literal-within-tuple-call/, since I don't think we want to use generators in tuple calls.

Setting `preview = true` might have been a bit optimistic. They seem to still make a lot of changes in that area even for established rules :\


inducer:
Thx!

PR #425, opened 2024-08-09
Restore `bcast_numpy_array=True` in `DOFArray` for now

majosm:
This change broke compatibility with mirgecom, since mirgecom still broadcasts non-object numpy arrays in some places. Should the behavior be deprecated instead? (Not sure how to do that formally, other than just letting the arraycontext warning take care of it.)

Depends on #424.


inducer:
Done via ffa34b8a7c7b2612c1d3b76659895dcb5e3a468b.

PR #427, opened 2024-08-12
Try out NumpyArrayContext

alexfikl:
This adds a subclass of `NumpyArrayContext` and uses it in `test_array.py` and `test_meshmode.py` (these also used the `pytato` array context).

I'm not sure if this is supposed to just work™, but I was curious what fails :grin:


alexfikl:
From a quick skim:
* There seem to be some cool segmentation faults in there.
* The loopy kernels are not all super friendly with the C target (most seem to be missing a barrier implementation?)

inducer:
Thanks for doing this. All this is definitely intended to "just work", at least eventually.

inducer:
What made you decide to close this? I figure we'll want something like this in the medium term?

alexfikl:
> What made you decide to close this? I figure we'll want something like this in the medium term?

Just didn't think it was needed. I'm happy to reopen!

meshmode/array_context.py
```diff
@@ -198,6 +200,24 @@ def _transform_with_element_and_dof_inames(t_unit, el_inames, dof_inames):
...
+        default_ep = t_unit.default_entrypoint
+        options = default_ep.options
+        if not (options.return_dict and options.no_numpy):
+            raise ValueError("Loopy kernel passed to call_loopy must "
+                    "have return_dict and no_numpy options set. "
+                    "Did you use arraycontext.make_loopy_program "
+                    "to create this kernel?")
+
+        import loopy as lp
+        return lp.add_inames_for_unused_hw_axes(t_unit)
```
alexfikl, line 216: This was an attempt to silence some loopy errors. Not sure if it's a good idea..

PR #428, opened 2024-08-12
Add pytential deps for testing

alexfikl:
The pytential test wasn't running because it wasn't installing `sympy` (likely some change from the `pyproject` migration). This adds the extra dependencies to make sure it runs again.


alexfikl:
Seems like it was `sympy` that was missing. Tests are running again :grin:

PR #430, opened 2024-08-19
Skip element orientation test for face meshes

majosm:
The test only works if `mesh.dim == mesh.ambient_dim`, so it always ends up emitting a warning in the `make_mesh` call here.


majosm:
Mypy errors appear to be unrelated.

inducer:
Thx!

PR #433, opened 2024-09-01
Fix editable installs

alexfikl:
This was required to locally see the errors from
https://github.com/inducer/pytential/actions/runs/10652244738/job/29525807445?pr=164
when everything is installed with `--editable`.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -70,6 +70,10 @@ include = [
     "meshmode*",
 ]
 
+[tool.setuptools.package-dir]
+# https://github.com/Infleqtion/client-superstaq/pull/715
+"" = "."
+
 [tool.setuptools.package-data]
 meshmode = [
     "py.typed",
--- a/doc/interop.rst
+++ b/doc/interop.rst
@@ -52,7 +52,7 @@ straightforward. Some language is different:
   created by calling the function :func:`~firedrake.functionspace.FunctionSpace`
   and referred to as a "function space"
 * In a mesh, any vertices, faces, cells, etc. are :mod:`firedrake`
-  "entities" (see `the PETSc documentation on dmplex <https://www.mcs.anl.gov/petsc/petsc-current/docs/manualpages/DMPLEX/index.html>`__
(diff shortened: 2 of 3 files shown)


pyproject.toml
```diff
@@ -70,6 +70,10 @@ include = [
     "meshmode*",
 ]
 
+[tool.setuptools.package-dir]
+# https://github.com/Infleqtion/client-superstaq/pull/715
+"" = "."
```
inducer, line 75: So, do we now need this on every package that we're hoping to install editable and use static analysis tools with?

inducer: Leaving this open to track...

alexfikl: Yeah, it seems like this is needed if e.g. I install both `meshmode` and `pytential` in editable mode and then run `mypy` in `pytential`. The CI installs them properly, so there it just works..

That PR links to the relevant `mypy` and `setuptools` issue, but it doesn't seem like there's any progress there.

alexfikl: Also worth noting.. I think this also works if they're installed with
```
python -m pip install --config-settings editable_mode=strict --editable .
```
but it seemed like a better idea to add this snippet to `pyproject.toml`, since the extra flags are easy to forget.

alexfikl: Just remembered I already rambled about this :cry: 
https://github.com/inducer/modepy/pull/88

inducer: Oh dear, I dislike this quite a bit. But at the same time, well, it might be the least bad solution.

alexfikl: Agreed.. their flaky implementation of editable installs annoyed me enough to mostly move away from setuptools :(

inducer: What's your preferred alternative?

alexfikl: Went for [hatchling](https://hatch.pypa.io/latest/config/build/#build-system). It seemed quite small and had nice defaults for these things.

It's just two-ish years old though, but it's in the `pypa` org, so hopefully it'll stay around for a while :\ Not sure there's a nice solid choice to be made at the moment..

PR #436, opened 2024-10-11
Use `TensorProductQuadrature` instead of `Quadrature` in `TensorProductElementGroupBase`

a-alveyblanc:
Attempting to access the 1D quadratures of the default quadrature rule supplied in `TensorProductElementGroupBase` is not possible. 

This PR makes it so that a `TensorProductQuadrature` is constructed from 1D `Quadrature`s using the 1D interpolation nodes and 1D basis functions associated with `TensorProductElementGroupBase`.

### DIFF
--- a/meshmode/discretization/poly_element.py
+++ b/meshmode/discretization/poly_element.py
@@ -525,12 +525,14 @@ class TensorProductElementGroupBase(PolynomialElementGroupBase,
 
     @memoize_method
     def quadrature_rule(self):
-        basis = self._basis
-        nodes = self._nodes
+        basis = self._basis.bases[0]
+        nodes = self.unit_nodes_1d
         mass_matrix = mp.mass_matrix(basis, nodes)
         weights = np.dot(mass_matrix,
                          np.ones(len(basis.functions)))
-        return mp.Quadrature(nodes, weights, exact_to=self.order)
+        quads = (mp.Quadrature(nodes, weights, exact_to=self.order),)*self.dim
+
+        return mp.TensorProductQuadrature(quads)
 
     @property
     @memoize_method
--- a/meshmode/discretization/poly_element.py
+++ b/meshmode/discretization/poly_element.py
@@ -534,9 +534,7 @@ class TensorProductElementGroupBase(PolynomialElementGroupBase,
         else:
             nodes_tp = self._nodes
 
-        for idim, zipped in enumerate(zip(nodes_tp, self._basis.bases)):
-            nodes, basis = zipped
-
+        for idim, (nodes, basis) in enumerate(zip(nodes_tp, self._basis.bases)):
             # get current dimension's nodes from fastest varying axis
             if self.dim != 1:
                 nodes = np.swapaxes(nodes, 0, idim)[:, *(0,)*(self.dim-1)]
--- a/meshmode/discretization/poly_element.py
+++ b/meshmode/discretization/poly_element.py
@@ -539,7 +539,7 @@ class TensorProductElementGroupBase(PolynomialElementGroupBase,
 
             # get current dimension's nodes from fastest varying axis
             if self.dim != 1:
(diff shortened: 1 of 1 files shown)


meshmode/discretization/poly_element.py
```diff
@@ -525,12 +525,14 @@ def basis_obj(self):
...
-        nodes = self._nodes
+        basis = self._basis.bases[0]
+        nodes = self.unit_nodes_1d
         mass_matrix = mp.mass_matrix(basis, nodes)
         weights = np.dot(mass_matrix,
                          np.ones(len(basis.functions)))
-        return mp.Quadrature(nodes, weights, exact_to=self.order)
+        quads = (mp.Quadrature(nodes, weights, exact_to=self.order),)*self.dim
+
+        return mp.TensorProductQuadrature(quads)
```
alexfikl, line 535: This should maybe do a `for nodes, basis in zip(self._nodes, self._basis.bases)` instead? In case the tensor product is not the same in all directions.

meshmode/discretization/poly_element.py
```diff
@@ -525,12 +525,29 @@ def basis_obj(self):
...
+        from modepy.tools import reshape_array_for_tensor_product_space as fold
+
+        quads = []
+
+        if self.dim != 1:
+            nodes_tp = fold(self.space, self._nodes)
+        else:
+            nodes_tp = self._nodes
+
+        for idim, zipped in enumerate(zip(nodes_tp, self._basis.bases)):
```
alexfikl, line 537:
```suggestion
        for idim, (nodes, basis) in enumerate(zip(nodes_tp, self._basis.bases)):
```
should unpack it correctly, so there's no need for the next line.

meshmode/discretization/poly_element.py
alexfikl: That looks a bit nastier than I expected :( Thanks for changing it!

meshmode/discretization/poly_element.py
```diff
@@ -525,12 +525,27 @@ def basis_obj(self):
 
     @memoize_method
     def quadrature_rule(self):
-        basis = self._basis
-        nodes = self._nodes
-        mass_matrix = mp.mass_matrix(basis, nodes)
-        weights = np.dot(mass_matrix,
-                         np.ones(len(basis.functions)))
-        return mp.Quadrature(nodes, weights, exact_to=self.order)
+        from modepy.tools import reshape_array_for_tensor_product_space as fold
```
inducer, line 528: Why the rename?

meshmode/discretization/poly_element.py
```diff
@@ -525,12 +525,27 @@ def basis_obj(self):
...
+        if self.dim != 1:
+            nodes_tp = fold(self.space, self._nodes)
+        else:
+            nodes_tp = self._nodes
+
+        for idim, (nodes, basis) in enumerate(zip(nodes_tp, self._basis.bases)):
+            # get current dimension's nodes from fastest varying axis
+            if self.dim != 1:
+                nodes = np.swapaxes(nodes, 0, idim)[:, *(0,)*(self.dim-1)]
+
```
inducer, line 541: This seems unnecessarily complicated?

Why not `nodes_tp[*((0,)*idim), :, *((0,)*(n-1-idim)]`?

meshmode/discretization/poly_element.py
```diff
@@ -525,12 +525,27 @@ def basis_obj(self):
...
+            nodes_1d = nodes.reshape(1, -1)
+            mass_matrix = mp.mass_matrix(basis, nodes_1d)
+            weights = np.dot(mass_matrix, np.ones(len(basis.functions)))
+
+            quads.append(mp.Quadrature(nodes_1d, weights, exact_to=self.order))
+
+        return mp.TensorProductQuadrature(quads)
```
inducer, line 548: This quadrature rule is expected to use the same nodes as the element group. Since we're making fresh nodes, there should be an `assert` here ensuring that.

PR #438, opened 2024-11-04
Bump Python to 3.10

alexfikl:


meshmode/mesh/generation.py
```diff
@@ -1478,7 +1479,7 @@ def generate_regular_rect_mesh(
             "lower topological dimension and map it.)")
 
     axis_coords = [np.linspace(a_i, b_i, npoints_i)
-            for a_i, b_i, npoints_i in zip(a, b, npoints_per_axis)]
+            for a_i, b_i, npoints_i in zip(a, b, npoints_per_axis, strict=False)]
```
alexfikl, line 1482: This is set to false because many tests set `a = (-0.5, -0.5, -0.5)` and pass the whole thing in. Not sure it's worth being very strict here..

inducer: For now I think it's fair to capture current usage. We could add a warning later if we feel it's an issue.

PR #439, opened 2024-11-14
mypy: fix error from MultiVector typing

alexfikl:


meshmode/mesh/processing.py
```diff
@@ -624,7 +624,8 @@ def unpack_single(ary: np.ndarray | None) -> np.ndarray:
 
     from pymbolic.geometric_algebra import MultiVector
 
-    mvs = [MultiVector(vec) for vec in spanning_object_array]
+    mvs: list[MultiVector[np.floating]] = (
+        [MultiVector(vec) for vec in spanning_object_array])
```
alexfikl, line 628: I'm not very familiar with this code, but this seems incorrect to me. 

From what I can tell, `vec` is an `np.ndarray`, but that's not an allowed type for `MultiVector`. Should this be wrapped again like `MultiVector({0: vec})`?

alexfikl, line 628: I'm not very familiar with this code, but this doesn't seem to be correct.

`spanning_object_array` is a 2D object array, so vec will be a 1D object array of the spanning vectors. Then the `CoeffT` for `MultiVector` should be.. `np.ndarray`?

inducer: I think the runtime behavior is fine. There's a `np.ndarray` case in the `MultiVector` constructor:

https://github.com/inducer/pymbolic/blob/main/pymbolic/geometric_algebra/__init__.py#L537-L544

What *is* weird is that that case does not provide a constraint on `CoeffT`, and that mypy is OK with that. (IMO it should not.) At the same time, changing that type signature to be `| np.ndarray[Any, CoeffT]` does not work because
```
pymbolic/geometric_algebra/__init__.py:539: error: Type argument "CoeffT" of "ndarray" must be a subtype of "dtype[Any]"  [type-var]
```
Bridging this gap of mutual misunderstanding might be interesting if I was a mypy or numpy developer. In actual reality though, I vote we wait until somebody figures out a way of teaching the type system which way is up.

inducer:
Thanks!

meshmode/mesh/processing.py
alexfikl: Yeah, it is weird that mypy just went along with this.. maybe it yells more with `--strict`? Anyway, agreed that this is good enough.

PR #440, opened 2024-11-14
Pass `strict=True` in zip() where applicable

inducer:


inducer:
Imagine my excitement when I realized that this had long since been done. :facepalm:

PR #445, opened 2024-12-06
Permutation connection

jonathan-macart:
Added:
1. `get_reordering_by_pymetis`: Compute mesh element reorderings (permutations) via [Metis's nested dissection algorithm](https://dlasalle.github.io/publications/mtmetis2015nd.pdf).
2. `make_element_permutation_connection`: First attempt at generating a connection that permutes the elements of an existing mesh, say `from_mesh`, to those of a new mesh, `to_mesh`, using the ND-generated permutation. The goal is to be able to pass `to_mesh` to a CFD solver. I do not need the mesh nodes to be reordered. My understanding is that `all_elements_perm` should contain the permuted element tags of `from_mesh`, though this could be completely incorrect.

Currently, (2) returns a discretization connection containing two identical discretizations/meshes. A few thoughts/questions:

- Is constructing `to_discr = from_discr.copy()` a problem? My understanding is that this creates an exact replica of `from_discr.mesh`, which is not the goal. `to_discr.mesh` should contain the same nodes and elements of `from_discr.mesh`, but with the mesh ordering permuted using `perm`.
- Is a discretization connection a mapping between two "immutable" meshes? Or is `to_discr.mesh` actually changed upon creation of the connection?
- If the answer to the last question is "no," how do I construct a copy of `from_discr.mesh` with the elements permuted?

Any advice would be appreciated! Thank you!


meshmode/discretization/connection/permute.py
```diff
@@ -0,0 +1,90 @@
+__copyright__ = "Copyright (C) 2014 Andreas Kloeckner"
```
inducer, line 1: Please fix the copyright notice.

inducer:
Side note: please disregard the RUF052 (https://github.com/astral-sh/ruff/issues/14796) and pylint linter failures (https://github.com/pylint-dev/pylint/issues/10000#issuecomment-2519899671).

meshmode/discretization/connection/permute.py
```diff
@@ -0,0 +1,90 @@
...
+def make_element_permutation_connection(actx, from_discr):
+    """Build discretization connection containing a permuted mesh.
+
+    :arg actx: an :class:`~arraycontext.ArrayContext`.
+    :arg from_discr: a :class:`Discretization`.
+    """
+
+    # Reorder the local mesh using Metis
+    from meshmode.distributed import get_reordering_by_pymetis
+    perm, iperm = get_reordering_by_pymetis(from_discr.mesh)
```
inducer, line 49: Could you make the permutation a parameter to this function, rather than unconditionally obtaining it from pymetis?

inducer, line 49: Will `perm, iperm` match the return value of `get_reordering_by_pymetis`? It seems that returns only a single array?

meshmode/discretization/connection/permute.py
```diff
@@ -0,0 +1,90 @@
...
+    InterpolationBatch,
+)
+from meshmode.transform_metadata import (
+    DiscretizationElementAxisTag,
+)
+
+
+# {{{ permute mesh elements
+
+def make_element_permutation_connection(actx, from_discr):
```
inducer, line 40:
```suggestion
def make_element_permutation_connection(actx: ArrayContext, from_discr: Discretization) -> DirectDiscretizationConnection:
```

meshmode/distributed.py
```diff
@@ -445,6 +445,49 @@ def get_partition_by_pymetis(mesh, num_parts, *, connectivity="facial", **kwargs
     return np.array(p)
 
 
+# FIXME: Move somewhere else, since it's not strictly limited to distributed?
+def get_reordering_by_pymetis(mesh, *, connectivity="facial", **kwargs):
```
inducer, line 449:
```suggestion
def get_reordering_by_pymetis(mesh: meshmode.mesh.Mesh, *, connectivity: Literal["facial"] | Literal["nodal"], **kwargs) -> np.ndarray:
```

meshmode/discretization/connection/permute.py
inducer, line 40: From @jonathan-macart's email, my understanding was that the objective of this is a mesh with the elements reordered. This connection instead shuffles DOF data between elements, while leaving the discretization in place.

To achieve the goal I mention above, you would need to make a permuted mesh, then make a discretization based on that.

meshmode/discretization/connection/permute.py
```diff
@@ -0,0 +1,90 @@
...
+        # Permuted elements
+        all_elements_perm = actx.tag(NameHint(f"all_el_ind_perm_grp{igrp}"),
+                                     actx.tag_axis(0,
+                                                   DiscretizationElementAxisTag(),
+                                                   actx.from_numpy(perm)))
+        all_elements_perm = actx.freeze(all_elements_perm)
+        ibatch = InterpolationBatch(
+                from_group_index=igrp,
+                from_element_indices=all_elements,
+                to_element_indices=all_elements_perm,  ## CHK??
```
inducer, line 75: This mishandles meshes/discretizations with multiple element groups (e.g. a mesh that's part-hex-part-tet), in that it is silently broken (by using out-of-bounds indices) in that case. It's fine to assume single-group only, but then the code should error if presented with multiple groups.

meshmode/distributed.py
inducer, line 449: I think an argument could be made that the connectivity-getter could be factored into a separate function, since it seems it's duplicated from `get_partition_by_pymetis`?

meshmode/discretization/connection/permute.py
inducer, line 1: Please add tests.

PR #450, opened 2025-02-08
Add a cruller (twisted torus) geometry

alexfikl:
This just gently modifies `generate_torus_and_cycle_vertices` to allow creating the cruller geometry from https://doi.org/10.1016/j.jcp.2019.109047. Looks like this:

<p align="center">

</p>

### DIFF
--- a/meshmode/mesh/generation.py
+++ b/meshmode/mesh/generation.py
@@ -71,6 +71,7 @@ Surfaces
 .. autofunction:: generate_cube_surface
 .. autofunction:: generate_sphere
 .. autofunction:: generate_torus
+.. autofunction:: generate_cruller
 .. autofunction:: refine_mesh_and_get_urchin_warper
 .. autofunction:: generate_urchin
 .. autofunction:: generate_surface_of_revolution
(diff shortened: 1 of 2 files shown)


meshmode/mesh/generation.py
```diff
@@ -933,7 +949,51 @@ def generate_torus(
...
+
+    where :math:`H(u, v) = 1.0 + 0.25 * \cos(5 u + 3 v)`. The standard torus is
+    obtained using :math:`H(u, v) = 1`.
+    """
+    mesh, _, _ = generate_torus_and_cycle_vertices(
+            r_major, r_minor,
+            n_major=n_major,
+            n_minor=n_minor,
+            order=order,
+            twist=lambda u, v: 1.0 + 0.25 * np.cos(5.0 * u + 3.0 * v),
```
alexfikl, line 996: This doesn't have knobs to actually tweak this, since I thought it's easy enough to use `generate_torus_and_cycle_vertices` directly in that case. Maybe that `0.25` could be tweakable to control how strong the waviness is..

meshmode/mesh/generation.py
```diff
@@ -761,20 +761,31 @@ def ensure_radius(arr: np.ndarray) -> np.ndarray:
...
 def generate_torus_and_cycle_vertices(
         r_major: float, r_minor: float,
-        n_major: int = 20, n_minor: int = 10, order: int = 1,
+        *,
+        n_major: int = 20,
+        n_minor: int = 10,
+        order: int = 1,
         node_vertex_consistency_tolerance: float | bool | None = None,
         unit_nodes: np.ndarray | None = None,
+        twist: Callable[[np.ndarray, np.ndarray], np.ndarray] | None = None,
```
alexfikl, line 770: Namin? :\

meshmode/mesh/generation.py
```diff
@@ -933,7 +949,51 @@ def generate_torus(
...
+
+        \begin{aligned}
+        x &= \cos(u) (r_\text{major} + r_\text{minor} H(u, v) \cos(v)) \\
+        y &= \sin(u) (r_\text{major} + r_\text{minor} H(u, v) \sin(v)) \\
+        z &= r_\text{minor} H(u, V) \sin(v)
+        \end{aligned}
+
+    where :math:`H(u, v) = 1.0 + 0.25 * \cos(5 u + 3 v)`. The standard torus is
+    obtained using :math:`H(u, v) = 1`.
+    """
```
inducer, line 990: Maybe acknowledge the source/inspiration?

meshmode/mesh/generation.py
inducer: Still trying to figure out what this even does... :slightly_smiling_face: "Twist" makes sense, but then what puts the "divets" into the surface?

alexfikl: Hm, as it is now, it does some funky amplitude modulation on `r_minor`, so maybe something more in that direction.. I don't know..

meshmode/mesh/generation.py
alexfikl: Done in latest version:
https://github.com/inducer/meshmode/compare/0407b6f981f89d5f6b275c0c13f4820ba64d9565..da325a29d74a7d21a84be944c3ea57b5f9f47bb4

meshmode/mesh/generation.py
alexfikl: Renamed it to `r_minor_func` and added some docs to explain what it does. Any suggestions welcome! :grin:

inducer: That's much better!

inducer: (I previously thought "twist" would "twist" the u-v coordinate system (without shrinking the radius).)

PR #461, opened 2025-05-07
All tags

inducer:
cc @anderson2981

https://github.com/inducer/gmsh_interop/pull/36 should pass with this


.github/workflows/ci.yml
```diff
@@ -55,6 +55,19 @@ jobs:
                 python -m pip install mypy
                 ./run-mypy.sh
 
+    basedpyright:
+        name: Mypy
```
matthiasdiener, line 59:
```suggestion
```

inducer: Thx!

PR #470, opened 2025-07-28
Add types to poly_element

alexfikl:
Not that many missing, but this should make it fairly complete!

### DIFF
--- a/.basedpyright/baseline.json
+++ b/.basedpyright/baseline.json
@@ -488,102 +488,6 @@
         ],
         "./meshmode/discretization/__init__.py": [
             {
-                "code": "reportUnannotatedClassAttribute",
-                "range": {
-                    "startColumn": 13,
-                    "endColumn": 27,
-                    "lineCount": 1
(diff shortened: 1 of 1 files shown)


meshmode/discretization/__init__.py
```diff
@@ -51,13 +51,14 @@
...
-    from collections.abc import Hashable, Iterable, Sequence
+    from collections.abc import Callable, Hashable, Iterable, Sequence
 
     from numpy.typing import DTypeLike
 
+    from modepy.typing import ArrayF
     from pytools.obj_array import ObjectArray
 
-    from meshmode.mesh import Mesh as _Mesh, MeshElementGroup as _MeshElementGroup
+    from meshmode.mesh import Mesh, MeshElementGroup
```
alexfikl, line 61: From what I recall, these had an underscore so that they're not mistakenly imported from here. Since they're in a `TYPE_CHECKING` block now, that shouldn't be an issue.

PR #471, opened 2025-08-02
Remove deprecated keyword in mesh.generation

alexfikl:
This didn't have a deadline, but they were deprecated in 2020 (from `git blame`) and there are no uses of the old behavior in `pytential`, `grudge`, etc.


inducer:
Thx!

PR #472, opened 2025-08-03
Improve types in mesh.generation

alexfikl:
Mostly `np.ndarray` to `NDArray[np.floating]`, but it also has some other small typing updates.

### DIFF
--- a/meshmode/mesh/generation.py
+++ b/meshmode/mesh/generation.py
@@ -28,6 +28,7 @@ from typing import TYPE_CHECKING, Any
 
 import numpy as np
 import numpy.linalg as la
+from typing_extensions import override
 
 import modepy as mp
 from pytools import log_process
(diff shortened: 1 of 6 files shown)


meshmode/mesh/generation.py
```diff
@@ -1594,17 +1624,21 @@ def generate_box_mesh(
 def generate_regular_rect_mesh(
         a: Sequence[float] = (0, 0),
         b: Sequence[float] = (1, 1), *,
-        nelements_per_axis: int | tuple[int, ...] | None = None,
-        npoints_per_axis: int | tuple[int, ...] | None = None,
-        periodic: bool | None = None,
+        nelements_per_axis: tuple[int, ...] | None = None,
+        npoints_per_axis: tuple[int, ...] | None = None,
+        periodic: tuple[bool, ...] | None = None,
```
alexfikl, line 1629: The code doesn't seem to handle `npoints_per_axis` to be just an `int` (?), so this just removes it.

inducer:
Thx!

PR #475, opened 2025-08-05
Add annotations to mesh.processing

alexfikl:
Similar to #472.


meshmode/mesh/processing.py
```diff
@@ -1145,7 +1147,7 @@ def _match_boundary_faces(
 
     m_idx_to_n_idx = find_point_to_point_mapping(
         bdry_pair_mapping.aff_map(bdry_m_vertices),
-        bdry_n_vertices)
+        bdry_n_vertices, tol=tol)
```
alexfikl, line 1150: This function took a `tol` that wasn't used anywhere and this seemed like the only place it could go. I'm not familiar with the code here, so not very sure..

PR #495, opened 2026-03-24
Do not create duplicates in untag_loopy_call_results.

kaushikcfd:
See https://github.com/inducer/pytato/pull/653/.


inducer:
Thx!

PR #505, opened 2026-07-12
Remove unused ruff noqa

alexfikl:
Also, it seems like this got optimized in Python 3.14, so it may no longer be an issue either way.

xref: https://github.com/astral-sh/ruff/pull/25707


inducer:
> Also, it seems like this got optimized in Python 3.14, so it may no longer be an issue either way.

Oh, good to know!

